PR #37, opened 2023-04-18
DMP-44 Data Modelling with JPA mappings for MVP (darts entities)

richardbull:
https://tools.hmcts.net/jira/browse/DMP-44

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/darts/entities/Annotation.java
+++ b/src/main/java/uk/gov/hmcts/reform/darts/entities/Annotation.java
@@ -1,0 +1,71 @@
+package uk.gov.hmcts.reform.darts.entities;
+
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.persistence.JoinColumn;
+import jakarta.persistence.OneToOne;
+import jakarta.persistence.Table;
+import jakarta.persistence.Temporal;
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/hmcts/reform/darts/entities/Annotation.java
```diff
@@ -0,0 +1,71 @@
...
+    @Temporal(TemporalType.DATE)
+    @Column(name = "c_start")
+    private Date start;
+
+    @Temporal(TemporalType.DATE)
+    @Column(name = "c_end")
+    private Date end;
+
+    @Column(name = "c_courthouse", length = 64)
+    private String courthouse;
```
jackmaloney, line 51: Do wen really need to put length on these fields in the Java model?

src/main/java/uk/gov/hmcts/darts/entities/Annotation.java
```diff
@@ -0,0 +1,71 @@
+package uk.gov.hmcts.darts.entities;
```
jackmaloney, line 1: Think these need to be moved to the common package e.g. `uk.gov.hmcts.darts.common.entities`

src/main/java/uk/gov/hmcts/reform/darts/entities/Annotation.java
jackmaloney: ok, guess we create another ticket on backlog to track this and decide how to proceed

jackmaloney (approved):

PR #38, opened 2023-04-19
Dmp 58 setup app insights

alancoleman:

https://tools.hmcts.net/jira/browse/DMP-58

### Change description ###
Attempt to service logs to App Insights

```
[ ] Yes
[x ] No
```


infrastructure/.terraform-version
```diff
@@ -1 +0,0 @@
-1.4.4
```
jackmaloney, line 1: Did we want to be removing this?

lib/applicationinsights.json
```diff
@@ -1,4 +1,5 @@
 {
+  "connectionString": "${file:/mnt/secrets/draft-store/app-insights-connection-string}",
   "role": {
     "name": "rpe-demo"
```
jackmaloney, line 4: This role name needs to be updated to something DARTS related I think

infrastructure/variables.tf
```diff
@@ -1,15 +1,18 @@
-variable "product" {}
-
-variable "component" {}
-
-variable "location" {
-  default = "UK South"
+variable "resource_group_name" {
```
jackmaloney, line 1: Did you mean to remove all of the other fields that existed here?

jackmaloney (commented):
Did we not want to follow the pattern defined in this file? Or is there a reason that's not compatible? https://github.com/hmcts/draft-store/blob/master/infrastructure/app-insights.tf

infrastructure/.terraform-version
alancoleman: Thanks @jackmaloney, the Terraform version is described in providers.tf and the plan executes on Demo without this file, however it's probably a little naive of me to remove it so I'll add it back in.

alancoleman: I did notice that removing this file create a conflict anyway. So adding it back in makes sense.

lib/applicationinsights.json
alancoleman: I think I'll need to add the required Azure resources before attempting to connect or pass anything to them. That's why I left the application part for another commit. Thanks

infrastructure/variables.tf
alancoleman: Thanks @jackmaloney,  yes, location, name and tags are the only arguments available when creating an Azure Resource Group, didn't want anything else confusing matters. https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/resource_gro
up

alancoleman:
> Did we not want to follow the pattern defined in this file? Or is there a reason that's not compatible? https://github.com/hmcts/draft-store/blob/master/infrastructure/app-insights.tf

Thanks @jackmaloney, it looks as though most of the draft-store infrastructure directory is applicable to the draft-store app, the database set up etc, not just the app insights. I think what we're looking for is a lot simpler: Resource Group, App Insights, Azure Key and then connect the app with the resulting strings. Kicking things off with a Resource Group will at least give me an intro into how this process works and what basic config is required. Happy to discuss. Thanks

PR #45, opened 2023-04-21
Dmp 45 postgre sql terraform

scott-robertson1:

### Change description ###

```
[ ] Yes
[ ] No
```


config/owasp/suppressions.xml
```diff
@@ -7,6 +7,13 @@
     </notes>
     <cve>CVE-2016-1000027</cve>
   </suppress>
+  <suppress>
+    <notes>False Positives
```
jackmaloney, line 11: Could we add a date to this one like we do here? https://github.com/hmcts/darts-gateway/blob/dd4a252245e408ebd2880f00941e54628c0e7bd6/config/owasp/suppressions.xml#L13 
That way we won't suppress it forever, if there is a fix that gets released.

src/main/resources/application.yaml
```diff
@@ -20,27 +20,39 @@ spring:
     import: "optional:configtree:/mnt/secrets/darts/"
   application:
     name: Spring Boot Template
```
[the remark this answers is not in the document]
scott-robertson1: yeah will get that updated, probably should be darts-api

config/owasp/suppressions.xml
```diff
@@ -7,6 +7,13 @@
...
+      CVE-2023-20863 - https://github.com/spring-projects/spring-framework/issues/24434#issuecomment-579669626
```
jackmaloney, line 12: This can probably be removed now as covered in the suppressions below

PR #53, opened 2023-04-26
Dmp 45 postgre sql terraform jack test

jackmaloney:


charts/darts-api/values.stg.template.yaml
```diff
@@ -1,17 +1,22 @@
...
-    environment:
-    ENABLE_FLYWAY: true
-    RUN_DB_MIGRATION_ON_STARTUP: true
-    DB_HOST: "{{ .Release.Name }}-postgresql"
-    DB_NAME: "{{ .Values.postgresql.auth.database}}"
-    DB_USER: "{{ .Values.postgresql.auth.username}}"
-    DB_PASS: "{{ .Values.postgresql.auth.password}}"
-    DB_PORT: "5432"
+  keyVaults:
+    cpo:
```
davet1985, line 6: this should be `darts`

src/main/resources/db/migration/V0001__init.sql
```diff
@@ -0,0 +1,771 @@
+-- List of Table Aliases
```
davet1985, line 1: Where did all this come from? I'm not sure we want this in here yet?

PR #55, opened 2023-04-28
DMP-45 Add h2 database

cmagowan-hmcts:
Changed tests to use in memory h2 database

```
[ ] Yes
[x] No
```


build.gradle
```diff
@@ -228,6 +228,7 @@ dependencies {
   implementation 'org.postgresql:postgresql:42.6.0'
   runtimeOnly 'org.hsqldb:hsqldb'
 
+  testImplementation 'com.h2database:h2:2.1.214'
```
richardbull, line 231: Do we need to specify the version here? Or is this because of a CVE - I can't remember what I seen before with the dependency added above on line 229. I originally tried h2 there but it failed a build / security check. So I just switched it to use hsqldb to start the app at the time.

cmagowan-hmcts: Its always best to specify a version in case an update to the package happens which breaks your build.

PR #67, opened 2023-05-03
DMP-122 Using Govuk notify API key from vault

davet1985:
- in jenkins
- in the application

Running functional tests using test API Key

### Change description ###

```
[ ] Yes
[x] No
```

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -12,7 +12,7 @@ def branchesToSync = ['demo']
 
 def secrets = [
   'darts-${env}': [
-    secret('GovukNotifyTestApiKey', 'GovukNotifyApiKey')
+    secret('GovukNotifyTestApiKey', 'GOVUK_NOTIFY_API_KEY')
   ],
 ]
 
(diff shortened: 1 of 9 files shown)


Jenkinsfile_CNP
```diff
@@ -10,11 +10,27 @@ def component = "api"
 
 def branchesToSync = ['demo']
 
+def secrets = [
+  'darts-${env}': [
+    secret('GovukNotifyTestApiKey', 'GovukNotifyApiKey')
```
cmagowan-hmcts, line 15: i think the envVar needs to be whatever is in application.yaml.
Not sure if it will work with the dot in it, but it might be:-
`secret('GovukNotifyTestApiKey', 'darts.GovNotifyApiKey')`

Looking at SSCS code, they seem to reference the spring property as normal:-
image

Here it is defined in the application.yaml:-
image

Not sure why they have 2 different properties for the prod and test keys, i think we could handle that inside the profiles.

Then they set the secret here:-
image

and then the secrets are stored in the vault
image

davet1985: Yeah I was following the same thing, think it should be like that now? I don't believe it will work with a dot so I've taken that out.

cmagowan-hmcts: yup, looks good.

tonymort (approved):

scott-robertson1 (approved):

PR #74, opened 2023-05-05
DMP-121 Schedule Notification requests to Gov Notify using Spring Sch…

cmagowan-hmcts:
…eduler.

https://tools.hmcts.net/jira/browse/DMP-121 

Schedule Notification requests to Gov Notify using Spring Scheduler.

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/darts-api/Chart.yaml
+++ b/charts/darts-api/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "1.0"
 description: A Helm chart for darts-api App
 name: darts-api
 home: https://github.com/hmcts/darts-api
-version: 0.0.8
+version: 0.0.9
 maintainers:
   - name: HMCTS darts team
 dependencies:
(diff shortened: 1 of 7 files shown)


davet1985:
This PR doesn't cover this part of the task
```
The cron expression should be able to be configured differently for each environment. This will involve adding some config backed by an env var which can be configured and passed into the process via flux and helm.
```

davet1985:
I think it would also be better to use a cron expression as this will give us more options, such as not running notifications at the weekend etc.

cmagowan-hmcts:
> This PR doesn't cover this part of the task
> 
> > The cron expression should be able to be configured differently for each environment. This will involve adding some config backed by an env var which can be configured and passed into the process via flux and helm.

done

cmagowan-hmcts:
> I think it would also be better to use a cron expression as this will give us more options, such as not running notifications at the weekend etc.

done

tonymort (approved):
LGTM

PR #84, opened 2023-05-11
Dmp 45 postgres ql2

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-45

Setup Postgres database

```
[ ] Yes
[x] No
```


bin/main/application-dev.yaml
```diff
@@ -0,0 +1,61 @@
+spring:
```
chrisbellingham-hmcts, line 1: What's the purpose of this `.yaml`, and how does its purpose differ from `src/main/resources/application-dev.yaml`?

src/main/resources/application-dev.yaml
```diff
@@ -5,11 +5,28 @@ spring:
...
+    type: com.zaxxer.hikari.HikariDataSource
+  jpa:
+    database: postgresql
+    show-sql: false
+    hibernate:
+      ddl-auto: create
     properties:
-      charSet: UTF-8
+      hibernate:
+        #dialect: org.hibernate.dialect.PostgreSQLDialect
```
chrisbellingham-hmcts, line 19: Remove this unused config item.

src/main/resources/application.yaml
```diff
@@ -41,27 +41,33 @@ spring:
...
+    driver-class-name: org.postgresql.Driver
+    url: jdbc:postgresql://${DARTS_API_DB_HOST:localhost}:5432/${DARTS_API_DB_NAME:darts}${DARTS_API_DB_OPTIONS:}
+    username: ${DARTS_API_DB_USERNAME:darts}
+    password: ${DARTS_API_DB_PASSWORD:darts}
+    type: com.zaxxer.hikari.HikariDataSource
+  jpa:
+    database: postgresql
+    show-sql: false
+    hibernate:
+      ddl-auto: create
```
chrisbellingham-hmcts, line 55: Is this what we want, or should we instead be relying on flyway to create the necessary DB schemas?

src/main/resources/application.yaml
```diff
@@ -41,27 +41,33 @@ spring:
             token-uri: https://hmctsdartsb2csbox.b2clogin.com/hmctsdartsb2csbox.onmicrosoft.com/B2C_1_darts_externaluser_signin/oauth2/v2.0/token
             jwk-set-uri: https://hmctsdartsb2csbox.b2clogin.com/hmctsdartsb2csbox.onmicrosoft.com/B2C_1_darts_externaluser_signin/discovery/v2.0/keys
 
-#  datasource:
```
chrisbellingham-hmcts, line 44: `application-local.yaml` has the following comment:
```
# Once the postgres config is uncommented in the main application.yml the below config may not be needed
```
So it may be worth reviewing if `application-local.yaml` needs to be updated in line with these changes.

.gitignore
```diff
@@ -23,3 +23,6 @@ bin/main/application.yaml
 
 applicationinsights-agent-*.jar
 *.log
+
+### Visual Studio Code ###
+bin
```
davet1985, line 28: Can you remove the files in `bin/` that have been added as part of this PR

src/main/resources/application.yaml
davet1985: Yes, I believe it should be flyway?

cmagowan-hmcts: yes, but flyway isn't working yet, Flyway changes are waiting for this to be merged first, so we need to temporarily have this until its working. So this can be changed back as part of flyway

.gitignore
cmagowan-hmcts: done

davet1985: Oh it looks like `bin/run-in-docker.sh` shouldn't have been removed?

Might need this

```
bin
!bin/run-in-docker.sh
```

bin/main/application-dev.yaml
cmagowan-hmcts: deleted

src/main/resources/application-dev.yaml
cmagowan-hmcts: removed

.gitignore
cmagowan-hmcts: added back in now. Couldn't exclude it like you suggested, as according to git:-

> It is not possible to re-include a file if a parent directory of that file is excluded. Git doesn’t list excluded directories for performance reasons, so any patterns on contained files have no effect, no matter where they are defined.

https://git-scm.com/docs/gitignore

So have specified the folders within bin to exclude individually.

cmagowan-hmcts: although it comes up as an empty file in the diff view, but displays fine if you look at the branch, which is odd.

davet1985 (approved):
Happy if @chrisbellingham-hmcts has checked over the Java parts more closely than me 👍

PR #86, opened 2023-05-11
DMP-145: Secure Backend Endpoints

chrisbellingham-hmcts:
### [DMP-145](https://tools.hmcts.net/jira/browse/DMP-145)

Key changes:
* Enable OAUTH2 authentication for all non-whitelisted DARTS API endpoints.
* Disable security layer for all integration tests
* Add new `FunctionalTest` class, which is intended to be extended by all functional tests. This provides a straightforward way for functional tests to obtain a JWT from Azure AD B2C, and to call DARTS API with that JWT in the request header.
* Update `client_id` and `client_secret` config for `local` and `dev` to align with new Azure AD client registration `darts-dev`.
* Update helm to obtain env vars for functional test.
* Update README to document new env vars and profile usage guidelines.

```
[ ] Yes
[x] No
```


charts/darts-api/values.dev.template.yaml
scott-robertson1, line 1: are these keys only required for the dev environment ?

chrisbellingham-hmcts: They're only required when the functional tests are run. So would that include both `dev` and `stg`?

hemantasharma1129 (approved):

PR #81, opened 2023-05-09
DMP-142 - Create DB table to store the Audio Request, along with the …

garethmorriscgi:
https://tools.hmcts.net/jira/browse/DMP-142

Create audtio requst database table and entity class

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #87, opened 2023-05-11
DMP-157 create audio metadata table

marcmollins:
https://tools.hmcts.net/jira/browse/DMP-157

### Change description ###

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/entity/Media.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/entity/Media.java
@@ -23,6 +23,9 @@ public class Media {
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Integer id;
 
+    @Column(name = "moj_crt_id")
+    private Integer courthouseId;
+
     @Column(name = "r_media_object_id", length = 16)
     private String legacyObjectId;
 
@@ -43,14 +46,17 @@ public class Media {
     @Column(name = "c_end")
     private Date end;
 
-    @Column(name = "c_courthouse", length = 64)
-    private String courthouse;
-
     @Column(name = "c_courtroom", length = 64)
     private String courtroom;
 
     @Column(name = "c_reporting_restrictions")
     private Integer reportingRestrictions;
+
+    @Column(name = "c_case_id", length = 32)
+    private String caseId;
+
+    @Column(name = "r_case_object_id", length = 16)
+    private String caseObjectId;
 
     @Column(name = "r_version_label", length = 32)
     private String legacyVersionLabel;
--- a/src/main/resources/db/migration/media_table.sql
+++ b/src/main/resources/db/migration/media_table.sql
@@ -1,0 +1,23 @@
+-- Table: darts_schema.moj_media
+
+-- DROP TABLE IF EXISTS darts_schema.moj_media;
+
+CREATE TABLE IF NOT EXISTS darts_schema.moj_media
+(
+    moj_med_id INTEGER NOT NULL,
+    moj_crt_id INTEGER,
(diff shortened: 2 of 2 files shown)


src/main/resources/db/migration/media_table.sql
```diff
@@ -0,0 +1,22 @@
+-- Table: darts_schema.notification
+
+-- DROP TABLE IF EXISTS darts_schema.media;
+
+CREATE TABLE IF NOT EXISTS darts_schema.media
```
richardbull, line 5: One for @hemantasharma1129 / @jackmaloney to confirm the ask...
The entity we have under `common` was an initial mapping based on what Barry had in Sharepoint (v5).

It looks like there is a newer version (v8) where the DDL is

The common entities would need a refresh. e.g. add sequences,remove character/numeric size limits, change DATE to TIMESTAMP
```
CREATE TABLE DARTS.moj_media
(moj_med_id                 INTEGER					NOT NULL
,moj_crt_id                 INTEGER
,r_media_object_id          CHARACTER VARYING(16)
,c_channel                  NUMERIC
,c_total_channels           NUMERIC
,c_reference_id             CHARACTER VARYING
,c_start                    TIMESTAMP
,c_end                      TIMESTAMP
,c_courtroom                CHARACTER VARYING
,c_reporting_restrictions   NUMERIC
,c_case_id                  CHARACTER VARYING(32)[]                    --this is a placeholder for moj_case_document_r.c_case_id, known to be repeated for moj_media object types
,r_case_object_id           CHARACTER VARYING(16)[]                    --this is a placeholder for moj_case_s.r_object_id, and must be derived from moj_case_document_r.c_case_id
,r_version_label            CHARACTER VARYING(32)
,i_superseded               BOOLEAN
,i_version_label            SMALLINT);
```

src/main/java/uk/gov/hmcts/darts/common/entity/Media.java
```diff
@@ -62,4 +62,6 @@ public class Media {
     @Column(name = "i_version_label")
     private Short version;
 
+    @Column(name = "i_storage_location", length = 64)
```
richardbull, line 65: This column doesn't match the migration sql below where it's named `storage_location`.
We were also going to drop the length attribute.
But suspect @hemantasharma1129 / the leads were wanting a new entity close to what @garethmorriscgi did in `audio` https://github.com/hmcts/darts-api/pull/81

src/main/resources/db/migration/media_table.sql
```diff
@@ -0,0 +1,22 @@
+-- Table: darts_schema.notification
```
davet1985, line 1: incorrect name here?

src/main/java/uk/gov/hmcts/darts/common/entity/Media.java
davet1985, line 65: I think just call this

```suggestion
    @Column(name = "storage_location", length = 64)
```

Also where did the length come from?

davet1985: We have some inconsistencies in the format of the tables we are creating. Some of them are derived from Barry's work (like here) and some we have simply made up (possibly because it doesn't exist in legacy darts). I think we need to decide which approach we are going with CC @jackmaloney @hemantasharma1129

cmagowan-hmcts:
> Apologies, it wasn't clear on the ticket, but could the name of the table please be `moj_media` and the column names match those in the entity class, ie. using the prefixes.

are we keeping the moj_ prefix??

davet1985 (approved):
LGTM, please bring up to date with master

PR #111, opened 2023-05-31
DMP-234 Use Spring Envers for notification table

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-234

Added Spring envers functionality to Notification table.

```
[ ] Yes
[x] No
```


src/main/resources/application-generateDDL.yaml
```diff
@@ -0,0 +1,4 @@
+spring.jpa.properties.javax.persistence.schema-generation.create-source: metadata
+spring.jpa.properties.javax.persistence.schema-generation.scripts.action: create
+spring.jpa.properties.javax.persistence.schema-generation.scripts.create-target: src/main/resources/db/migration/V1_1__create_from_hibernate.sql
```
richardbull, line 3: Is this yaml file just a temporary thing to get the DDL from our entities using hibernate?

richardbull: I think you will need to add the flyway script to add the corresponding audit table for the Notification entity that is `@Audited`

richardbull (changes requested):
Please can you add the next V1_4__ flyway script to add the new envers rev / audit tables (notification_audit, revinfo) and FK constraint similar to Barry's convention.

`CREATE SEQUENCE IF NOT EXISTS revinfo_seq;`

PR #138, opened 2023-06-09
dmp-438 : Add Token Filter in Security Config

hemantasharma1129:

- [x] tests have been updated / new tests has been added (if needed)

https://tools.hmcts.net/jira/browse/DMP-438

### Change description ###
Added Token Filter in Security Config to check if Authorization Headers exist

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/darts/authentication/AuthenticationFunctionalTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/darts/authentication/AuthenticationFunctionalTest.java
@@ -6,7 +6,7 @@ import org.junit.jupiter.api.Test;
 import uk.gov.hmcts.darts.FunctionalTest;
 
 import static io.restassured.RestAssured.given;
-import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.*;
 
 class AuthenticationFunctionalTest extends FunctionalTest {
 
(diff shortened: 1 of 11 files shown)


src/main/resources/application-local.yaml
```diff
@@ -20,11 +20,26 @@ spring:
             prompt: login
             issuer-uri: https://hmctsdartsb2csbox.b2clogin.com/${AAD_B2C_TENANT_ID_KEY:00000000-0000-0000-0000-000000000000}/v2.0/
             provider: external-azure-ad-provider
+          internal-azure-ad:
```
richardbull, line 23: I don't think we should be adding the same properties to local, dev profiles if they are already on the default `application.yaml`. We should only be adding the properties here that we want to override

hemantasharma1129: This is a placeholder for internal AD and we will have new config/properties once justice team provides it.

cmagowan-hmcts (approved):

src/functionalTest/java/uk/gov/hmcts/darts/authentication/AuthenticationFunctionalTest.java
```diff
@@ -34,7 +35,7 @@ void shouldReturnUnauthorizedWhenSecuredEndpointIsCalledWithoutAuth() {
             .then()
             .extract().response();
 
-        assertEquals(401, response.statusCode());
+        assertFalse(response.getHeader("Location").isEmpty());
```
jackmaloney, line 38: Double negative makes this line abit hard to read - is there a better way? Do we know what the location should be or does it change each time? 
If not, maybe `assertTrue(response.getHeader("Location").isNotEmpty());` might be clearer?

hemantasharma1129: As discussed changed to assertNotnull.

PR #146, opened 2023-06-13
DMP-184 Add event type records to database

richardbull:

https://tools.hmcts.net/jira/browse/DMP-184

### Change description ###

Add 376 rows in postgres (darts-db) moj_event_type

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/common/entity/Courthouse.java
```diff
@@ -20,8 +23,8 @@ public class Courthouse {
 
     @Id
     @Column(name = "moj_crt_id")
-    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "courthouse_gen")
-    @SequenceGenerator(name = "courthouse_gen", sequenceName = "moj_crt_seq", allocationSize = 1)
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "moj_cth_gen")
```
cmagowan-hmcts, line 26: should the sequence generator not match the name of the primary key? i.e. moj_crt_gen ?

src/main/java/uk/gov/hmcts/darts/common/entity/Case.java
```diff
@@ -5,76 +5,64 @@
...
 @SuppressWarnings({"PMD.ShortClassName"})
 public class Case {
 
     private static final String MAPPED_BY_THE_CASE = "theCase";
 
     @Id
     @Column(name = "moj_cas_id")
-    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "moj_cas_gen")
+    @SequenceGenerator(name = "moj_cas_gen", sequenceName = "moj_cas_seq", allocationSize = 1)
```
cmagowan-hmcts, line 31: i don't think we should have an allocationSize of 1

src/main/java/uk/gov/hmcts/darts/common/entity/Event.java
```diff
@@ -18,9 +19,16 @@ public class Event {
 
     @Id
     @Column(name = "moj_eve_id")
-    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "moj_eve_gen")
+    @SequenceGenerator(name = "moj_eve_gen", sequenceName = "moj_eve_seq", allocationSize = 1)
```
cmagowan-hmcts, line 23: i don't think allocationSize should be this low.

src/main/java/uk/gov/hmcts/darts/common/entity/Case.java
richardbull: Need to leave at 1 for now to match our sql. Barry is going to review the proposed schema and will update his scripts with some sequences having a modified `increment` and `cache` options and we can change our entities to match after that.

cmagowan-hmcts (approved):

PR #153, opened 2023-06-19
Update NotificationServiceImpl to log to application insights

marcmollins:
added a logging.info for when a notification has failed but sending will be retried. added a logging.error for when the max tries have been exhausted and will not be resent.

https://tools.hmcts.net/jira/browse/DMP-84

### Change description ###

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/notification/service/impl/NotificationServiceImpl.java
```diff
@@ -106,6 +106,7 @@ public void sendNotificationToGovNotify() {
                     govNotifyRequest,
                     e.getMessage()
                 );
+                log.error("notification failed to send");
```
cmagowan-hmcts, line 109: this is probably not needed, as there is an error being logged on the line above.

src/main/java/uk/gov/hmcts/darts/notification/service/impl/NotificationServiceImpl.java
```diff
@@ -122,8 +123,10 @@ private void incrementNotificationFailureCount(Notification notification) {
         if (attempts <= maxRetry) {
             notification.setAttempts(attempts);
             notification.setStatus(String.valueOf(NotificationStatus.PROCESSING));
+            log.info("Notification has failed to send, retrying");
```
cmagowan-hmcts, line 126: might be good to specify the id of the notification that has failed.

cmagowan-hmcts (approved):

PR #154, opened 2023-06-19
DMP-181 receive and store Daily List

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-181

Receive and store a daily list in the database.


src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListServiceImpl.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Override
+    /*
+    Retrieve the new Daily List, and store it in the database.
+     */
+    public void processIncomingDailyList(DailyListPostRequest postRequest) {
+        DailyList dailyList = postRequest.getDailyList();
+
+        dailyList.getCrownCourt().getCourtHouseName();
+        Courthouse courthouse = retrieveCourtHouse(dailyList);
```
davet1985, line 39: I thought we had agreed to use an access pattern where by we expose a single point of entry into a feature for use by other features, an interface in an "api" package. As per @chrisbellingham-hmcts presentation a little while ago.

src/integrationTest/java/uk/gov/hmcts/darts/authentication/controller/impl/LoginOrRefreshIntTest.java
```diff
@@ -35,6 +36,8 @@ class LoginOrRefreshIntTest {
 
     @MockBean
     private CourthouseRepository courthouseRepository;
```
HMCTSDeveloper, line 38: is this repository need for this test? I can see there is a notification repo and audio repo as well.

src/integrationTest/java/uk/gov/hmcts/darts/common/controller/CourthouseApiTest.java
```diff
@@ -183,15 +178,15 @@ void courthousesPut() throws Exception {
             .contentType(MediaType.APPLICATION_JSON_VALUE);
 
         mockMvc.perform(requestBuilder).andExpect(status().is2xxSuccessful())
-            .andExpect(jsonPath("$.courthouse_name", is("SWANSEA")))
-            .andExpect(jsonPath("$.code", is(457)))
+            .andExpect(jsonPath("$.courthouse_name", is("test")))
```
HMCTSDeveloper, line 181: what was the reason for changing the test values? I think its a good idea for tests to use real data where we can.

src/main/java/uk/gov/hmcts/darts/courthouse/service/CourthouseServiceImpl.java
```diff
@@ -48,4 +52,43 @@ public Courthouse addCourtHouse(uk.gov.hmcts.darts.courthouse.model.Courthouse c
...
+        if (courthouseOptional.isEmpty()) {
+            //code not found, lookup name instead
+            courthouseOptional = repository.findByCourthouseName(courthouseNameUC);
+            if (courthouseOptional.isEmpty()) {
+                throw new CourthouseNameNotFoundException(courthouseNameUC);
+            }
+            Courthouse courthouse = courthouseOptional.get();
+            if (courthouse.getCode() == null && courthouseCode != null) {
+                //update courthouse with new code
+                courthouse.setCode(courthouseCode.shortValue());
```
HMCTSDeveloper, line 83: I guess the assumption here is that when this method is called the courthouse code is valid and is actually for the courthouseName.

src/main/java/uk/gov/hmcts/darts/courthouse/service/CourthouseServiceImpl.java
```diff
@@ -48,4 +52,43 @@ public Courthouse addCourtHouse(uk.gov.hmcts.darts.courthouse.model.Courthouse c
...
+     * retrieves the courtroom from the database.
+     *
+     * @param courthouseCode Optional parameter. If it is not provided, then name will be used by itself.
+     * @param courthouseName Name of the courthouse to search for.
+     * @return the found courtroom
+     * @throws CourthouseNameNotFoundException when the courthouse isn't found
+     * @throws CourthouseCodeNotMatchException when the courtroom is found, but it has a different code that expected.
+     */
+    @Override
+    public Courthouse retrieveCourtHouse(Integer courthouseCode, String courthouseName)
```
HMCTSDeveloper, line 67: would it be possible to put the update in a different location , feels like retrieve method doesn't imply there is an update.

src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListServiceImpl.java
cmagowan-hmcts: i guess in the future, this should be broken up into multiple tickets?
1. Create the new endpoint on Courthouse
2. create a feignClient to call it. (although that quite a small job)
3. create the dailyList service

src/main/java/uk/gov/hmcts/darts/common/entity/DailyListEntity.java
```diff
@@ -52,42 +50,37 @@ public class DailyList {
...
-    private OffsetDateTime timeStamp;
+    private OffsetDateTime timestamp;
 
     @Column(name = DAILY_LIST_ID)
-    private Integer dailylistId;
+    private Integer dailyListId;
 
     @Column(name = START_DATE)
-    private OffsetDateTime startDate;
+    private Date startDate;
```
richardbull, line 71: Are we saying these 2 columns aren't `TIMESTAMP WITH TIMEZONE`? If not, seems wrong to change the Java type here

If they are to become simply `DATE` types in postgres then I think they should be `LocalDate` Java types here?

src/main/resources/db/migration/V1_8__Create_daily_list_table.sql
```diff
@@ -10,6 +10,8 @@ CREATE TABLE IF NOT EXISTS moj_daily_list (
 , c_end_date                     TIMESTAMP WITH TIME ZONE
 , c_daily_list_id_s              CHARACTER VARYING
 , c_daily_list_source            CHARACTER VARYING
+, daily_list_content             CHARACTER VARYING
```
richardbull, line 13: Should we be going back and changing old flyway scripts - does that not confuse the flyway schema history?

src/main/resources/openapi/dailyList.yaml
```diff
@@ -36,11 +36,11 @@ paths:
         '500':
           description: Internal Server Error
 
-  /dailylist/getCases:
+  /dailylists/getCases:
```
richardbull, line 39: I guess we should change our paths here to be `/dailylists/get-cases`?
[Must: Use lowercase separate words with hyphens for Path Segments [129]](https://hmcts.github.io/restful-api-standards/#129)

src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListServiceImpl.java
davet1985: @cmagowan-hmcts not sure I understand. My comment is that we shouldn't be accessing the service within a different "feature" directly. Instead we should access through a single entry point as defined by an interface.

I'm not saying we should use the API endpoint.

Does that make sense?

src/main/resources/db/migration/V1_8__Create_daily_list_table.sql
davet1985: At the moment flyway is set to clean and re-create everything it is run, we left it like that because of the data model changing so frequently and we have no data to be concerned about. So this is fine right now.

src/integrationTest/java/uk/gov/hmcts/darts/common/controller/CourthouseApiTest.java
cmagowan-hmcts: the test is testing the put values. the SWANSEA value was already in the database, and so couldn't be used as it causes a unique constraint violation. It accepts any string, and will not exist after this test, so doesn't matter what string is used...as long as it isn't swansea. i need to make sure that its a me that isn't going to be added to flyway in the future, as then the test would break.

src/main/java/uk/gov/hmcts/darts/courthouse/service/CourthouseServiceImpl.java
cmagowan-hmcts: i've decided to rewrite this section

src/main/java/uk/gov/hmcts/darts/courthouse/service/CourthouseServiceImpl.java
cmagowan-hmcts: have renamed

src/main/resources/openapi/dailyList.yaml
davet1985: Agreed

src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListServiceImpl.java
cmagowan-hmcts: yeah, now implemented

src/integrationTest/resources/tests/dailylist/DailyListServiceTest/insert1_duplicate_ok/DailyListRequest.json
```diff
@@ -0,0 +1,89 @@
+{
+  "document_id": {
+    "document_name": "DailyList_457_20210219174938.xml",
+    "unique_id": "CSDDL1613756980160",
+    "document_type": "DL",
+    "time_stamp": "2021-02-19T17:49:38.391+01:00"
```
richardbull, line 6: I thought we were expecting to receive only UTC date-time strings?
`OffsetDateTime` should handle this fine however, it is just a bit confusing seeing different timezones in the same request :)

src/main/java/uk/gov/hmcts/darts/dailylist/mapper/DailyListMapper.java
```diff
@@ -0,0 +1,45 @@
...
+        DocumentID documentId = dailyList.getDocumentId();
+        dailyListEntity.setCourthouse(courthouse);
+        dailyListEntity.setUniqueId(documentId.getUniqueId());
+        dailyListEntity.setStatus(String.valueOf(JobStatusType.NEW));
+        dailyListEntity.setTimestamp(documentId.getTimeStamp());
+        dailyListEntity.setStartDate(dailyList.getListHeader().getStartDate());
+        dailyListEntity.setEndDate(dailyList.getListHeader().getEndDate());
+        dailyListEntity.setSource(postRequest.getSourceSystem());
+
+        ObjectMapper objectMapper = new ObjectMapper();
```
richardbull, line 36: Can we use the `ObjectMapper` that Spring Boot uses it, or do we need to customise it?

richardbull (approved):

PR #157, opened 2023-06-20
DMP-197 added service functionality for audit api.

HMCTSDeveloper:
https://tools.hmcts.net/jira/browse/DMP-197

### Change description ###
Adding the audit functionality and linking to the controller.

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/audit/service/AudiServiceImpl.java
```diff
@@ -0,0 +1,62 @@
...
+
+    }
+
+    private List<Predicate> getPredicates(AuditSearchQuery auditSearchQuery, CriteriaBuilder criteriaBuilder, Root<Audit> root) {
+        List<Predicate> predicates = new ArrayList<>();
+        if (auditSearchQuery.getCaseId() != null) {
+            predicates.add(criteriaBuilder.equal(root.get(CASE_ID), auditSearchQuery.getCaseId()));
+        }
+
+        if (auditSearchQuery.getFromDate() != null && auditSearchQuery.getToDate() != null) {
```
cmagowan-hmcts, line 52: what if only a from date is specified?

HMCTSDeveloper: The validation is checking to make sure both dates are supplied.

src/integrationTest/resources/db/migration/V1_9999__integrationTestData.sql
```diff
@@ -0,0 +1 @@
+INSERT INTO audit VALUES (1, 2, 3, 4, '2023-06-13T08:13:09.688537759Z', 'application_server', 'additional_data');
```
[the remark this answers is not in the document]
HMCTSDeveloper: What's wrong with inserting at 1? is there an issue where tests might interfere with each other?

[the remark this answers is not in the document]
cmagowan-hmcts: you haven't reset the sequence, so the nextval of the sequence is still sitting at 1, so the next time the code tries to insert into it, it will try to insert with ID 1 also. And will therefore cause a unique constraint violation.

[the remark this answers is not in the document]
HMCTSDeveloper: ACCEPT. its not issue now because we only need 1 insert to test but changed the id to be a high value.

src/main/java/uk/gov/hmcts/darts/Application.java
```diff
@@ -26,4 +28,9 @@ public static void main(final String[] args) {
         SpringApplication.run(Application.class, args);
     }
 
+    @Bean
```
richardbull, line 31: I think the convention currently is to add a `Configuration` class under a module e.g. `audit.config` to hold our custom beans. Alternatively we could just specify `@Component` on the `AuditSearchQueryValidator` class

src/main/java/uk/gov/hmcts/darts/audit/service/AudiServiceImpl.java
```diff
@@ -0,0 +1,62 @@
...
+import uk.gov.hmcts.darts.audit.model.AuditSearchQuery;
+import uk.gov.hmcts.darts.common.entity.Audit;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@AllArgsConstructor
+@Service
+@Slf4j
+public class AudiServiceImpl implements AuditService {
```
richardbull, line 22: Should be renamed to be `AuditServiceImpl`

src/main/resources/db/migration/V1_9__create_audit_table.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS audit (
+  id                    INTEGER                    NOT NULL
+, case_id               INTEGER                    NOT NULL
+, event_id              INTEGER                    NOT NULL
+, moj_usr_id            INTEGER                    NOT NULL
```
richardbull, line 5: Should this column be `user_id` as it looks different from the others?

Table appears to be slightly different here than what's specified in the Data Model section of https://tools.hmcts.net/confluence/display/DMP/Case+audit+service+LLD
?

src/main/java/uk/gov/hmcts/darts/audit/service/AudiServiceImpl.java
HMCTSDeveloper: ACCEPT

src/main/resources/db/migration/V1_9__create_audit_table.sql
HMCTSDeveloper: ACCEPT.

src/main/java/uk/gov/hmcts/darts/Application.java
HMCTSDeveloper: ACCEPT

src/main/java/uk/gov/hmcts/darts/audit/AuditSearchQueryValidator.java
```diff
@@ -0,0 +1,28 @@
...
+        AuditSearchQuery auditSearchQuery = (AuditSearchQuery) target;
+        if (auditSearchQuery.getCaseId() == null
+            && auditSearchQuery.getEventId() == null
+            && auditSearchQuery.getFromDate() == null
+            && auditSearchQuery.getToDate() == null) {
+            errors.rejectValue(null, "filters.empty", "All filters were empty.");
+            return;
+        }
+
+        if (auditSearchQuery.getFromDate() == null ^ auditSearchQuery.getToDate() == null) {
```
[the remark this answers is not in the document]
HMCTSDeveloper: That wont work in the case where both are null. We only want the validation to fail if one of them is null but if both of them are null that is fine.

src/main/java/uk/gov/hmcts/darts/audit/service/AuditServiceImpl.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+@AllArgsConstructor
+@Service
+@Slf4j
+public class AuditServiceImpl implements AuditService {
+
+    public static final String CASE_ID = "caseId";
```
cmagowan-hmcts, line 24: if we make every column name in src/main/java/uk/gov/hmcts/darts/common/entity/Audit.java a constant, then you don't have to define them separately here.
see DailyListEntity.java for reference.

src/main/java/uk/gov/hmcts/darts/common/entity/Audit.java
```diff
@@ -0,0 +1,43 @@
...
+import org.hibernate.annotations.CreationTimestamp;
+
+import java.time.OffsetDateTime;
+
+@Entity
+@Table(name = "audit")
+@Data
+public class Audit {
+    @Id
+    @Column(name = "id")
```
cmagowan-hmcts, line 20: see previous comment about making the name of each column a  public constant.

src/main/java/uk/gov/hmcts/darts/audit/service/AuditServiceImpl.java
HMCTSDeveloper: these are not column names but entity member variable names. I doesn't make sense to make constants out of that.

cmagowan-hmcts (approved):

PR #169, opened 2023-06-23
DMP-437: Implement RFC-7807 compliant error response components

chrisbellingham-hmcts:
[DMP-437](https://tools.hmcts.net/jira/browse/DMP-437)

This change introduces [RFC-7807](https://datatracker.ietf.org/doc/html/rfc7807) "Problem Details" compliant error handling.

The intention is for the following error handling pattern to be followed for all features:

1. Define feature-specific error scenarios in an enum that implements `DartsApiError`. See `AuthenticationError` for an example. The expectation is that each error has a unique error code, a semantically correct HTTP error code, and a human-readable title.
2. Application code can then be written that throws/re-throws a `DartsApiException` as needed. This throwable accepts a  `DartsApiError` value as a constructor param. There are additional constructors to give flexibilitiy to the developer who may wish to include dynamic details in the exception/response.
3. The error-handling mechanism provided by `ExceptionHandler` and `DartsApiTrait` will intercept this exception type, map it to a `Problem` object and render this an RFC-7807-compliant HTTP response.

Some existing classes (`RestExceptionHandler`, `BasicRestException`) are marked `@Deprecated` and all usages are intended to be migrated to the approach above. This work will be handled under DMP-516.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/common/exception/DartsApiException.java
```diff
@@ -0,0 +1,39 @@
...
+
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+@Getter
+public class DartsApiException extends RuntimeException {
+
+    private final DartsApiError error;
+    private String detail;
```
richardbull, line 11: Should this be `final`?

src/main/resources/openapi/problem.yaml
```diff
@@ -0,0 +1,33 @@
+type: object
```
richardbull, line 1: Is this OAS viewable - is this just a template `ProblemError` schema for copying elsewhere?
Wondering what fields are `required`?

chrisbellingham-hmcts: Intention is to `$ref` this schema in a given package's openapi.yaml. So if I use `audioRequest.yaml` as an example, something like this (these lines would sit under `responses`):

```
        '409':
          description: audio request item already exists
          content:
            application/json-problem:
              schema:
                $ref: "./problem.yaml"
              example:
                type: "AUDIO_REQUEST_100"
                title: "Audio request item already exists"
                status: 409
```

The schema would render as such:
image

And the example response as:
image

The domain classes would also be generated, although we don't actually need to use them (the intention here is only to document the error responses).

Regarding what is required, good idea I will add `required: true` to the `type`, `title` and `status` fields.

chrisbellingham-hmcts: Turns out adding `required: true` breaks the build/generator. I think because this constraint is indended for requests rather than responses.

src/main/java/uk/gov/hmcts/darts/common/exception/DartsApiException.java
chrisbellingham-hmcts: Yes this would be better, change implemented.

richardbull (approved):

HMCTSDeveloper (approved):

PR #173, opened 2023-06-26
DMP-405 initial data management service to save and get binary blob data

garethmorriscgi:
https://tools.hmcts.net/jira/browse/DMP-405 

### Change description ###
Introduced a data management service that initially has two methods. One to save binary data to blob storage for a passed in container name which returns the stored unique id and a get method to fetch binary data for a particular container name and blob id.

The connection string is fetched from the key vault and passed in the services constructor.

Local this will need the Azurite emulator. The read me file has been updated with instructions.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/datamanagement/dao/impl/DataManagementDaoImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/datamanagement/dao/impl/DataManagementDaoImpl.java
@@ -1,0 +1,45 @@
+package uk.gov.hmcts.darts.datamanagement.dao.impl;
+
+import com.azure.storage.blob.BlobClient;
+import com.azure.storage.blob.BlobContainerClient;
+import com.azure.storage.blob.BlobServiceClient;
+import com.azure.storage.blob.BlobServiceClientBuilder;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.darts.datamanagement.config.DataManagementConfiguration;
+import uk.gov.hmcts.darts.datamanagement.dao.DataManagementDao;
+
+import java.util.UUID;
+
+@Component
+@Slf4j
+@RequiredArgsConstructor
+public class DataManagementDaoImpl implements DataManagementDao {
+
+    private final DataManagementConfiguration dataManagementConfiguration;
+    private BlobServiceClient blobServiceClient;
+
+    @Override
+    public BlobContainerClient getBlobContainerClient(String containerName) {
+        if (blobServiceClient == null) {
+            blobServiceClient = getBlobServiceClient();
+        }
+        return blobServiceClient.getBlobContainerClient(containerName);
+    }
+
+    @Override
+    public BlobClient getBlobClient(BlobContainerClient containerClient, UUID blobId) {
+        return containerClient.getBlobClient(String.valueOf(blobId));
+    }
+
+    public BlobServiceClient getBlobServiceClient() {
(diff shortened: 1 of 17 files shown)


cmagowan-hmcts (approved):

src/main/java/uk/gov/hmcts/darts/datamanagement/dao/impl/DataManagementDaoImpl.java
```diff
@@ -0,0 +1,41 @@
...
+@Component
+@Slf4j
+@RequiredArgsConstructor
+public class DataManagementDaoImpl implements DataManagementDao {
+
+    private final DataManagementConfiguration dataManagementConfiguration;
+    private BlobServiceClient blobServiceClient;
+
+    @Override
+    public BlobContainerClient getBlobContainerClient(String containerName) {
```
chrisbellingham-hmcts, line 24: Is there benefit to exposing `getBlobContainerClient()` as public? It looks like all usages are only calling this to next call `getBlobClient()`.

Wondering if it would make usage easier/cleaner if we simply had `getBlobClient()` as public.

(If motivation is testability, consider making `getBlobContainerClient()` package private instead).

src/main/java/uk/gov/hmcts/darts/datamanagement/service/impl/DataManagementServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+public class DataManagementServiceImpl implements DataManagementService {
+
+    @Autowired
+    private DataManagementDao dataManagementDao;
+
+    @Override
+    public BinaryData getBlobData(String containerName, UUID blobId) {
+
+        BlobContainerClient containerClient = dataManagementDao.getBlobContainerClient(containerName);
+        BlobClient blobClient = dataManagementDao.getBlobClient(containerClient, blobId);
```
chrisbellingham-hmcts, line 25: The pattern of calling `getBlobContainerClient()` followed by `getBlobClient()` is repeated in `saveBlobData()`, and it could probably be simplified by making it the responsibility of `getBlobClient()` to create the `BlobContainerClient` (see comment in `DataManagementDaoImpl.java`

src/main/java/uk/gov/hmcts/darts/datamanagement/service/impl/DataManagementServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+        return blobClient.downloadContent();
+    }
+
+    @Override
+    public UUID saveBlobData(String containerName, BinaryData binaryData) {
+
+        UUID uniqueBlobId = UUID.randomUUID();
```
chrisbellingham-hmcts, line 32: A debug log statement may be useful here, such that we can see the random `UUID` in dev/test environments.

src/main/resources/application.yaml
```diff
@@ -121,6 +121,15 @@ darts:
         transcription_available: a928742c-5eb7-46f0-9f61-d89d5394785c
         transcription_request_approved: 29c26992-df77-4ba6-8ba7-03324eb5ae37
         transcription_request_rejected: 739a31cf-13a1-49bc-bcf9-0794f4670dbb
+  storage:
+    blob:
+      connection-string: ${AZURE_STORAGE_CONNECTION_STRING:}
```
chrisbellingham-hmcts, line 126: I think this will default to an empty string if the env var is not set. Is that really what we want though? 

Instead, consider removing the `:`, such that the application will fail to start (fail fast) if the env var is not set, i.e:
```yaml
connection-string: ${AZURE_STORAGE_CONNECTION_STRING}
```

PR #176, opened 2023-06-27
DMP-399 Audio Transformation Service - MediaRequest: processAudioRequest

richardbull:

https://tools.hmcts.net/jira/browse/DMP-399

### Change description ###

* Updated `MediaRequest` entity
* Add `AudioTransformationService#processAudioRequest`
* Added enums and updated integrationTests

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+
+import static uk.gov.hmcts.darts.audio.enums.AudioRequestStatus.PROCESSING;
+
+@Service
+@RequiredArgsConstructor
+public class AudioTransformationServiceImpl implements AudioTransformationService {
+
+    private final MediaRequestRepository mediaRequestRepository;
+
+    @Transactional
```
cmagowan-hmcts, line 18: Does this need to be Transactional?
My understanding is that as Keda is spinning up a pod per request, there shouldn't be any concurrent access.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/MediaRequestServiceImpl.java
```diff
@@ -3,18 +3,25 @@
...
 @Service
-public class AudioRequestServiceImpl implements AudioRequestService {
+public class MediaRequestServiceImpl implements MediaRequestService {
+
+    private final MediaRequestRepository mediaRequestRepository;
 
-    private final AudioRequestRepository audioRequestRepository;
+    @Override
+    public MediaRequest getMediaRequestById(Integer id) {
+        return mediaRequestRepository.getReferenceById(id);
```
cmagowan-hmcts, line 23: I am assuming that the use-case for this endpoint will be to get the data of the whole MediaRequest object, instead of just its ID, so assume we will want to do a `find` here.

src/main/java/uk/gov/hmcts/darts/audio/repository/MediaRequestRepository.java
```diff
@@ -4,10 +4,7 @@
 import org.springframework.stereotype.Repository;
 import uk.gov.hmcts.darts.audio.entity.MediaRequest;
 
-import java.util.List;
-
 @Repository
-public interface AudioRequestRepository extends JpaRepository<MediaRequest, Integer> {
+public interface MediaRequestRepository extends JpaRepository<MediaRequest, Integer> {
```
cmagowan-hmcts, line 8: Can we use the opportunity to rename the `MediaRequest` entity to `MediaRequestEntity`?

src/test/java/uk/gov/hmcts/darts/audio/service/impl/MediaRequestServiceImplTest.java
```diff
@@ -17,22 +17,21 @@
...
     private static final Integer TEST_REQUESTER = 1234;
     private static final String OFFSET_T_09_00_00_Z = "2023-05-31T09:00:00Z";
     private static final String OFFSET_T_12_00_00_Z = "2023-05-31T12:00:00Z";
-    private static final String DOWNLOAD_REQ_TYPE = "Download";
-
-    private static final String REQUEST_STATUS_OPEN = "OPEN";
 
     @InjectMocks
-    private AudioRequestServiceImpl audioRequestService;
+    private MediaRequestServiceImpl audioRequestService;
```
cmagowan-hmcts, line 31: should we rename this to `mediaRequestService` also?

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
richardbull: I thought it should be similar to `createCourtoom`. I'm not sure what else will be added to the `processAudioRequest`, it seemed obvious candidate for marking as `Transactional`. It appears that removing the annotation here still works currently with `AudioTransformationServiceTest`, but I did need it added in other places for the Tests to run `- no Session`

src/main/java/uk/gov/hmcts/darts/audio/service/impl/MediaRequestServiceImpl.java
richardbull: I don't really understand this comment - this is getting the whole MediaRequest object?
I only added this service here because of the ticket description having
`getAudioRequest (requestId) - Fetch the Audio Request from Audio Request Table.`

Where requestId, we understand to be the PK :)
FYI @hemantasharma1129

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
cmagowan-hmcts: createCourtroom could be accessed by multiple means in the future, e.g. getCases, addEvent, addAudio etc. And so, especially with event, we could get a flurry of similar ones at once, and so if courtroom doesn't exist, we only want it to create 1 new one, and not multiple identical ones.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/MediaRequestServiceImpl.java
cmagowan-hmcts: there is a functional difference between `getReferenceById` and `findById`.
`getReferenceById` returns a proxy version of the entity, and would be used when you only want to link it to another entity, and you don't want to get or set any of the columns on it, but just refer to it.
This might explain it better
https://www.baeldung.com/jpa-entity-manager-get-reference

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
richardbull: - I've kept the `@Transactional` on this top level service, since it's updating the database.
- Added `@EnableTransactionManagement` to the Application class.
- After changing the `getMediaRequestById` to use `mediaRequestRepository.findById`, I was able to re-enable the failing test I disabled yesterday in `MediaRequestServiceTest`. 
- Also removed the `@Transactional` added yesterday for the Tests to run `- no Session`.

hemantasharma1129 (approved):

cmagowan-hmcts (approved):

PR #210, opened 2023-07-10
Dmp 539 table renaming

tonymort:
The test changes made in this PR are temporary.  The next ticket will address the inconsistencies and disabled tests.  The TABLESPACE statements have also been commented out temporarily as it needs directory permissions setting up in the environments.


chrisbellingham-hmcts (approved):
Approved on the basis tests will be cleaned up under a separate task.

PR #240, opened 2023-07-17
Dmp 461.advanced search

cmagowan-hmcts:

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/darts/cases/mapper/AdvancedSearchRequestMapper.java
```diff
@@ -0,0 +1,151 @@
...
+    private boolean joinHearing = false;
+    private boolean joinCourtroom = false;
+    private boolean joinCourthouse = false;
+    private boolean joinJudges = false;
+    private boolean joinDefendants = false;
+    private boolean joinEvents = false;
+
+    SQLQueryModel sqlModel = new SQLQueryModel();
+
+    public SQLQueryModel mapToSQL(GetCasesSearchRequest request) {
```
jackmaloney, line 23: This method could benefit from applying some of the functional programming principles outlined in this page https://www.baeldung.com/java-functional-programming 

Specifically, refer to the section about side affects: "Side effects can be anything apart from the intended behaviour of the method. For instance, side effects can be as simple as updating a local or global state". 
At present, it looks like this method is at risk of side effects, if for example another method is introduced in the future which updates the value of those same class variables. You could end up with unintended side effects, with this method potentially producing different results, depending on the order in which the method is called against an object.

PR #247, opened 2023-07-21
DMP-647 Refactor default handler for Event API

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-647

### Change description ###

Refactored the common elements of the default handler so it can be used by other event handlers. Also renamed it to StandardEventHandler.

```
[ ] Yes
[ X] No
```


tonymort (approved):
LGTM

src/main/java/uk/gov/hmcts/darts/event/handler/CourtroomCourthouseCourtcase.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.Builder;
+import lombok.Data;
+import uk.gov.hmcts.darts.common.entity.CaseEntity;
+import uk.gov.hmcts.darts.common.entity.CourthouseEntity;
+import uk.gov.hmcts.darts.common.entity.CourtroomEntity;
+
+
+@Data
+@Builder
+public class CourtroomCourthouseCourtcase {
```
cmagowan-hmcts, line 12: perhaps move this into a Model package?

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventHandlerBase.java
```diff
@@ -66,4 +95,107 @@ protected String buildKey(String type, String subType) {
...
+        return CourtroomCourthouseCourtcase.builder()
+            .courthouseEntity(actualCourtHouse)
+            .courtroomEntity(actualCourtRoomEntity)
+            .caseEntity(actualCourtCaseEntity)
+            .isHearingNew(hearingIsNew)
+            .isCourtroomDifferentFromHearing(isCourtroomDifferentFromHearing)
+            .build();
+    }
+
+    protected CourtroomEntity getCourtRoom(DartsEvent dartsEvent) {
```
cmagowan-hmcts, line 143: i think we should be creating a courtroom if its not been found instead of raising an exception.
There is already existing code in CommonApiImpl.retrieveOrCreateCourtroom that could be used, and raises more specific errors, although they are in the old format, so maybe could be updated.

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventHandlerBase.java
```diff
@@ -66,4 +95,107 @@ protected String buildKey(String type, String subType) {
...
+        }
+        final var actualCourtHouse = actualCourtRoomEntity.getCourthouse();
+        final var actualEventCaseNumber = caseNumbers.get(0);
+        final var actualEventDate = dartsEvent.getDateTime();
+
+        var actualCourtCaseEntity = getCourtCaseEntityOrCreate(actualCourtHouse, actualEventCaseNumber);
+
+        boolean hearingIsNew = false;
+        HearingEntity actualHearingEntity;
+        var optionalHearingEntity = getHearingEntity(actualEventDate, actualCourtCaseEntity);
```
cmagowan-hmcts, line 114: i think this logic is wrong. i think we need to be checking what courtroom the hearing is in also, as we could have had 2 hearings for the same case in 1 day.
1 in courtroom 1, another in courtroom2.
Therefore we want to make sure that it is the correct hearing.

Here is another scenario:-
Lets say we have 1 hearing in the hearing table, that is for courtroom 1 and is only scheduled (hearingIsActual is false).
lets say, they chnged their mind, and it actually happened in courtroom 2, when we receive teh event, we will look for a hearing in courtrroom 2 and not find one, but before creating it, lets check that there is any hearings in this courthouse that have hearingIsActual==false, if we do find one, then we want to grab it, update it to be the correct courtroom and set the hearingIsActual=true

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventHandlerBase.java
tonymort: This behaviour was agreed upon as part of DMP-153.  Jack wasn't sure we should be following the same logic as CommonApiImpl.retrieveOrCreateCourtroom.  Please catch up with Jack before changing this.  It isn't behaviour introduced as part of this ticket either.

cmagowan-hmcts (approved):
approving for now, but i think changes will need to be made, but they can be part of a new ticket.

PR #249, opened 2023-07-21
Dmp 463-implement_service_code

marcmollins:
https://tools.hmcts.net/jira/browse/DMP-463

```
[ ] Yes
[x ] No
```


src/main/resources/openapi/events.yaml
```diff
@@ -83,6 +83,48 @@ paths:
...
   /courtlogs:
+    get:
+      tags:
+        - Event
+      summary: An Endpoint which allows users to request courtlogs
+      description: |-
+        An Endpoint which allows users to request courtlogs using courthouse, casenumber, startdate and enddate
+      parameters:
+        - in: query
+          name: Courthouse
```
cmagowan-hmcts, line 94: can you make this camelCase?

src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerCourtLogsTest.java
```diff
@@ -120,4 +146,54 @@ private List<EventEntity> getAllLogEventsMatchingText() {
...
+
+        mockMvc.perform(requestBuilder).andExpect(MockMvcResultMatchers.status().isBadRequest()).andExpect(
+            MockMvcResultMatchers.header().string(
+                "Content-Type",
+                "application/problem+json"
+            ));
+    }
+
+    @Test
+    void courtLogsGetResultMatch() throws Exception {
```
cmagowan-hmcts, line 175: Could we have a test whereby there are 2 events for the case we are interested in, and 2 events for a different case on the same day in the same courtroom, to make sure we only get back the 2 events linked to the case we are interested in?

src/main/java/uk/gov/hmcts/darts/event/mapper/EventEntityToCourtLogMapper.java
```diff
@@ -0,0 +1,37 @@
...
+        List<CourtLog> logs = new ArrayList<>();
+
+        for (EventEntity entity : entities) {
+            logs.add(mapSingleEntityToCourtLog(entity));
+        }
+
+        return logs;
+    }
+
+    private CourtLog mapSingleEntityToCourtLog(EventEntity entity) {
```
cmagowan-hmcts, line 24: probably don't need the `SingleEntity` part of the name, as its implied by the fact that it accepts a single EventEntity parameter, so you could name this method `mapToCourtLog`

src/main/java/uk/gov/hmcts/darts/event/mapper/EventEntityToCourtLogMapper.java
```diff
@@ -0,0 +1,37 @@
...
+import uk.gov.hmcts.darts.common.entity.EventEntity;
+import uk.gov.hmcts.darts.event.model.CourtLog;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@UtilityClass
+public class EventEntityToCourtLogMapper {
+
+    public List<CourtLog> mapFromEntityToCourtLogs(List<EventEntity> entities, String courthouse, String caseNumber) {
```
cmagowan-hmcts, line 13: probably don't need the `FromEntity` part of the name, as its implied by the fact that it accepts an EventEntity List parameter, so you could name this method `mapToCourtLogList`

cmagowan-hmcts, line 13: are`courthouse` and `caseNumber` being used?

cmagowan-hmcts, line 13: might be better to call `entities` something more specific like `events` or `eventEntities`

src/main/java/uk/gov/hmcts/darts/event/mapper/EventEntityToCourtLogMapper.java
cmagowan-hmcts, line 24: might be better to call `entity` something more specific like `eventEntity`

src/main/resources/openapi/events.yaml
```diff
@@ -83,6 +83,48 @@ paths:
                 $ref: './problem.yaml'
 
   /courtlogs:
+    get:
+      tags:
+        - Event
+      summary: An Endpoint which allows users to request courtlogs
```
cmagowan-hmcts, line 89: wonder should we put `(events)` in here also, to make it obvious what we mean by courtlogs

src/main/resources/openapi/events.yaml
```diff
@@ -121,6 +163,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/EventsResponse'
```
cmagowan-hmcts, line 164: as there are more than 1 endpoint now, i think this should be renamed to make it clear that its for the Post request, so maybe something like `EventsPostResponse`

cmagowan-hmcts (approved):

PR #256, opened 2023-07-24
extract varchar arrays into their own tables

cmagowan-hmcts:
extract varchar arrays into their own tables


src/main/java/uk/gov/hmcts/darts/common/entity/CaseEntity.java
```diff
@@ -14,76 +13,98 @@
...
 import java.util.ArrayList;
 import java.util.List;
 
 @Entity
-@Table(name = "court_case")
+@Table(name = CaseEntity.TABLE_NAME)
 @SuppressWarnings({"PMD.ShortClassName"})
 @Getter
 @Setter
 public class CaseEntity {
```
tonymort, line 28: Been wondering whether we should change this to CourtCaseEntity to align with the table name?

tonymort (approved):

src/main/java/uk/gov/hmcts/darts/common/entity/CaseEntity.java
cmagowan-hmcts: hmm, probably should yeah, i'll make the change

PR #257, opened 2023-07-24
DMP-458 Add case

HMCTSDeveloper:

https://tools.hmcts.net/jira/browse/DMP-458

### Change description ###
Creating a rest endpoint to allow add/update case and related hearing.

```
[ ] Yes
[X ] No
```


src/main/java/uk/gov/hmcts/darts/cases/exception/CaseApiError.java
```diff
@@ -0,0 +1,40 @@
...
+@RequiredArgsConstructor
+public enum CaseApiError implements DartsApiError {
+
+    COURTHOUSE_PROVIDED_DOES_NOT_EXIST(
+        "100",
+        HttpStatus.BAD_REQUEST,
+        "Provided courthouse does not exist"
+    ),
+
+    COURTROOM_PROVIDED_DOES_NOT_EXIST(
```
davet1985, line 18: I'm not sure this error makes sense? I think if we don't have the courtroom already it should be created.

src/main/java/uk/gov/hmcts/darts/cases/exception/CaseApiError.java
```diff
@@ -0,0 +1,40 @@
...
+        "101",
+        HttpStatus.BAD_REQUEST,
+        "Provided courtroom does not exist"
+    ),
+    HEARING_MISMATCH(
```
davet1985, line 23: When is this error used?

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
```diff
@@ -31,10 +49,110 @@ public List<ScheduledCase> getCases(GetCasesRequest request) {
...
+            .filter(hearingEntity -> caseEntity.getId().equals(hearingEntity.getCourtCase().getId())
+                && hearingEntity.getHearingDate().equals(LocalDate.now())).findFirst();
+
+
+        return hearing.map(hearingEntity -> hearingRepository.saveAndFlush(mapToHearingEntity(
+            addCaseRequest,
+            caseEntity,
+            hearingEntity
+        ))).orElseGet(() -> saveNewHearingEntity(addCaseRequest, caseEntity));
+    }
```
davet1985, line 103: Shouldn't this check if the courtroom matches as well?

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
```diff
@@ -31,10 +49,110 @@ public List<ScheduledCase> getCases(GetCasesRequest request) {
...
+    private CaseEntity saveNewCaseEntity(AddCaseRequest addCaseRequest) {
+        CaseEntity caseEntity = mapAddCaseRequestToCaseEntity(addCaseRequest, new CaseEntity());
+        return caseRepository.saveAndFlush(caseEntity);
+    }
+
+    private HearingEntity mapToHearingEntity(AddCaseRequest addCaseRequest, CaseEntity caseEntity, HearingEntity hearingEntity) {
+        if (!StringUtils.isBlank(addCaseRequest.getCourtroom())) {
+            CourtroomEntity courtroomEntity = getExistingCourtroom(addCaseRequest);
+            if (courtroomEntity == null) {
+                throw new DartsApiException(CaseApiError.COURTROOM_PROVIDED_DOES_NOT_EXIST);
```
davet1985, line 122: I think we are meant to create the courtroom if it doesn't exist as we won't ever have a definitive list of them.

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
```diff
@@ -31,10 +49,110 @@ public List<ScheduledCase> getCases(GetCasesRequest request) {
...
+    private CourtroomEntity getExistingCourtroom(AddCaseRequest addCaseRequest) {
+        CourtroomEntity courtroomEntity;
+        courtroomEntity = courtroomRepository.findByNames(
+            addCaseRequest.getCourthouse(),
+            addCaseRequest.getCourtroom()
+        );
+        return courtroomEntity;
+    }
+
+    private CaseEntity mapAddCaseRequestToCaseEntity(AddCaseRequest addCaseRequest, CaseEntity caseEntity) {
```
davet1985, line 142: Should this go in `CaseMapper`?

src/main/resources/db/migration/common/V1_1__base_script.sql
```diff
@@ -415,7 +415,7 @@ IS 'used to record acceptable external locations, found in external_object_direc
 CREATE TABLE darts.hearing
 (hea_id                     INTEGER					   NOT NULL
 ,cas_id                     INTEGER                    NOT NULL
-,ctr_id                     INTEGER                    NOT NULL
+,ctr_id                     INTEGER
```
davet1985, line 418: Not sure about this change, to create a hearing we should always have a courtroom.

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerTest.java
```diff
@@ -108,9 +80,176 @@ private static HearingEntity setupCase3(CourthouseEntity swanseaCourthouse, Cour
...
 
         var hearingForCase1 = aHearingForCaseInRoom(case1, swanseaCourtroom1);
-        hearingForCase1.setJudges(List.of("{Judge1}"));
+        hearingForCase1.setJudges(List.of("{Judge3}"));
         hearingForCase1.setHearingDate(LocalDate.parse(HEARING_DATE));
         hearingForCase1.setScheduledStartTime(LocalTime.parse("11:00"));
         return hearingForCase1;
     }
+
+    private static HearingEntity setupCase4(CourthouseEntity swanseaCourthouse, CourtroomEntity swanseaCourtroom1) {
```
davet1985, line 89: Might be helpful to add a comment to say what type of case situation this is setting up

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerTest.java
```diff
@@ -108,9 +80,176 @@ private static HearingEntity setupCase3(CourthouseEntity swanseaCourthouse, Cour
...
+
+        String actualResponse = response.getResponse().getContentAsString();
+
+        String expectedResponse = getContentsFromFile(
+            "tests/cases/CaseControllerTest/casesPostEndpoint/expectedResponseCourthouseMissing_400.json");
+        assertEquals(expectedResponse, actualResponse, JSONCompareMode.NON_EXTENSIBLE);
+    }
+
+    @Test
+    void casesPostWithNonExistingCourtroom() throws Exception {
```
davet1985, line 196: As per other comments, I don't think this one is correct.

src/test/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImplTest.java
```diff
@@ -0,0 +1,418 @@
...
+        DartsApiException thrownException = assertThrows(
+            DartsApiException.class,
+            () -> service.addCaseOrUpdate(request)
+        );
+
+        assertEquals("Provided courthouse does not exist", thrownException.getMessage());
+    }
+
+    @Test
+    void testAddCaseNonExistingCourtroom() {
```
davet1985, line 240: Another courtroom one which I don't think is correct.

src/test/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImplTest.java
```diff
@@ -0,0 +1,418 @@
...
+
+        assertEquals("1", updatedHearingEntity.getCourtroom().getName());
+        assertEquals(SWANSEA, updatedHearingEntity.getCourtroom().getCourthouse().getCourthouseName());
+        assertEquals(1, updatedHearingEntity.getJudges().size());
+        assertEquals(LocalDate.now(), updatedHearingEntity.getHearingDate());
+        assertNotNull(updatedHearingEntity.getCourtCase());
+    }
+
+    @Test
+    void testUpdateCaseWithMultipleHearingsWithoutCourtroomInRequest() {
```
davet1985, line 372: Is there a test which does update case with multiple (or just one) hearings with a matching date, but not a matching courtroom?

It should create a new hearing for the same date with a different courtroom.

src/main/java/uk/gov/hmcts/darts/common/entity/CaseEntity.java
```diff
@@ -45,6 +46,7 @@ public class CaseEntity {
 
     @ManyToOne(cascade = {CascadeType.PERSIST, CascadeType.MERGE})
     @JoinColumn(name = "cth_id")
+    @JsonIgnore
```
davet1985, line 49: What's this change for?

davet1985 (changes requested):
A few comments/questions, happy to discuss.

davet1985:
@HMCTSDeveloper I think there's overlap between the code you have written for matching/creating/updating cases/hearings and what we have for events which needs to do the same thing.

I'm not saying change this, but could you have a quick look and maybe we can create a ticket to refactor it into a common place.

src/main/java/uk/gov/hmcts/darts/cases/exception/CaseApiError.java
HMCTSDeveloper: I can see that it was implemented in the case of getCases, I can implement it for addCase.

davet1985: There's something in the common/api package for get or create courtroom, can we use that?

src/main/java/uk/gov/hmcts/darts/cases/exception/CaseApiError.java
HMCTSDeveloper: Not required anymore, it was added when we our design could potentially not distinguish between multiple hearings.

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
HMCTSDeveloper: ACCEPT

src/main/java/uk/gov/hmcts/darts/cases/exception/CaseApiError.java
HMCTSDeveloper: Yes there is a method called retrieveOrCreateCourtroom.

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerTest.java
HMCTSDeveloper: ACCEPT

src/test/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImplTest.java
HMCTSDeveloper: The result is wrong as it should automatically create the courtroom as per your previous comment but I think we should keep the test to cover that scenario.

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerTest.java
HMCTSDeveloper: The result is wrong as it should automatically create the courtroom as per your previous comment but I think we should keep the test to cover that scenario.

src/test/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImplTest.java
HMCTSDeveloper: I don't think we have test that covers that exact scenario, we have the test testUpdateCaseWithMultipleHearingsWithCourtroomInRequest which has two courtrooms(1,2) and a hearing for each. We then do an update request for courtroom 2 and check if the case is updated and related hearing is updated.

src/main/java/uk/gov/hmcts/darts/common/entity/CaseEntity.java
HMCTSDeveloper: ACCEPT. Not required.

src/test/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImplTest.java
HMCTSDeveloper: Added a new test testUpdateCaseWithMultipleHearingsWithOldHearingDateWithCourtroomInRequest for the requested scenario.

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
HMCTSDeveloper: ACCEPT

src/main/java/uk/gov/hmcts/darts/cases/controller/CaseController.java
```diff
@@ -63,16 +65,22 @@ public ResponseEntity<List<ScheduledCase>> casesGet(@NotNull @Parameter(name = "
...
+        @Size(max = 30) @Parameter(name = "judge_name", description = "Full or partial Judge name", in = ParameterIn.QUERY) @Valid @RequestParam(value = "judge_name", required = false) String judgeName,
+        @Size(max = 30) @Parameter(name = "defendant_name", description = "Full or partial Defendant name", in = ParameterIn.QUERY) @Valid @RequestParam(value = "defendant_name", required = false) String defendantName,
+        @Parameter(name = "date_from", description = "DateFrom to search for the hearings in.", in = ParameterIn.QUERY) @Valid @RequestParam(value = "date_from", required = false) @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) LocalDate dateFrom,
+        @Parameter(name = "date_to", description = "DateTo to search for the hearings in.", in = ParameterIn.QUERY) @Valid @RequestParam(value = "date_to", required = false) @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) LocalDate dateTo,
+        @Size(max = 100) @Parameter(name = "event_text_contains", description = "Full or partial event_text", in = ParameterIn.QUERY) @Valid @RequestParam(value = "event_text_contains", required = false) String eventTextContains
```
davet1985, line 83: Not sure these changes should be here?

davet1985: @cmagowan-hmcts would this cause any issues with your PR?

src/main/resources/openapi/cases.yaml
```diff
@@ -52,6 +52,32 @@ paths:
...
+        '201':
+          description: Created
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/scheduled_case'
+        '400':
+          description: A required parameter is missing or an invalid datatype or value was provided for property.
+        '409':
+          description: Resource already exists.
```
davet1985, line 77: I don't think we would ever return this?

src/test/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImplTest.java
davet1985: I'm not sure that test covers the scenario above? It looks as if hearings exist for courtroom 1 and 2, then the request is for courtroom 1 again?

We need a hearing match on date only and check that it creates a new hearing with the specified courtroom.

src/main/java/uk/gov/hmcts/darts/cases/controller/CaseController.java
HMCTSDeveloper: These changes are already in master , so it should be fine.

src/main/resources/openapi/cases.yaml
HMCTSDeveloper: Yes, ACCEPT.

src/main/java/uk/gov/hmcts/darts/cases/controller/CaseController.java
davet1985: If that's the case they shouldn't be showing as a diff

HMCTSDeveloper: The diff is that  @jakarta.validation.Valid has been changed to @Valid , no point in having the qualifier when we are importing jakarta.validation.Valid;

davet1985: Ok, no problem.

src/test/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImplTest.java
HMCTSDeveloper: I covered the inverse of what you suggested previously, so I have created a new Test testUpdateCaseWithNonExistingCourtroomAndMatchingHearingDate. This will create a courtroom with an existing case and hearing for todays date. Then an update request will be made with a different , non existing, courtroom. This should match on hearing_date but not on courtroom thus creating a new hearing.

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerTest.java
```diff
@@ -33,14 +36,74 @@
...
+        case2.setProsecutorList(createListOfProsecutor(2, case2));
+
+        var hearingForCase2 = createHearingWith(case2, swanseaCourtroom1);
+        hearingForCase2.setJudgeList(createListOfJudgesForHearing(1, hearingForCase2));
+        hearingForCase2.setHearingDate(LocalDate.parse("2023-06-20"));
+        hearingForCase2.setScheduledStartTime(LocalTime.parse("10:00"));
+        return hearingForCase2;
+    }
+
+    // To be replaced by something more reusable as the last PR in this ticket
```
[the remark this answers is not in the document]
HMCTSDeveloper: Wasn't me 🤷

davet1985 (approved):

PR #258, opened 2023-07-24
DMP-402: Implement playback operation

chrisbellingham-hmcts:
[DMP-402](https://tools.hmcts.net/jira/browse/DMP-402)

Implement audio playback functionality. This largely consists of calling existing lower-level functionality such that the requested audio is:
* concatenated
* merged
* trimmed
* re-encoded
* and stored into outbound storage

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
@@ -10,6 +10,7 @@ import uk.gov.hmcts.darts.audio.service.AudioOperationService;
 import uk.gov.hmcts.darts.common.entity.MediaEntity;
 
 import java.nio.file.Path;
+import java.time.Duration;
 import java.time.OffsetDateTime;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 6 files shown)


karen-hedges (approved):

hemantasharma1129 (approved):

src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
```diff
@@ -136,20 +151,35 @@ private List<AudioFileInfo> concatenateByChannel(List<AudioFileInfo> audioFileIn
...
+
+        return audioOperationService.merge(audioFileInfos, StringUtils.EMPTY);
+    }
+
+    private AudioFileInfo trimToPeriod(AudioFileInfo audioFileInfo, OffsetDateTime start,
+                                       OffsetDateTime end) throws ExecutionException, InterruptedException {
+        return audioOperationService.trim(
+            StringUtils.EMPTY,
+            audioFileInfo,
+            start.toLocalTime().toString(),
```
richardbull, line 165: I'm not sure this conversion to LocalTime will work here, as the `Application` sets the Default Timezone to be UTC.  The darts-api.jar also sets the system property `-Duser.timezone=UTC` in the Dockerfile.

Also the `trim` startTime, endTime strings are simply the start, stop position of the media using the ffmpeg Time duration in `HH:MM:SS` format (http://ffmpeg.org/ffmpeg-utils.html#Time-duration)

chrisbellingham-hmcts: Good spot, have updated. Please can you re-review this area? If the approach looks reasonable, would it be better to move the `OutboundFileProcessorImpl.toTimeString(Duration duration)` method into `AudioOperationService` do you think?

chrisbellingham-hmcts: Discussed with @richardbull and we can leave `toTimeString` here. If other usages in the future will benefit from it being in `AudioOperationService`, we can easily move it.

richardbull: I think it looks ok what you've done in `OutboundFileProcessorImpl#trimToPeriod`
were you thinking you would change the service in the `AudioOperationService` to overload the trim method to use `Duration `params instead?

Up to you, whether you feel it's worth the hassle - not sure about moving the toTimeString thing - maybe useful as a utility method perhaps

richardbull (approved):

PR #274, opened 2023-07-26
DMP-614 Authorisation - Create Models - Roles, Permission & Userstate

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-614

### Change description ###

Created authorisation model objects for Role, Permission and UserState

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/authorisation/model/Permission.java
+++ b/src/main/java/uk/gov/hmcts/darts/authorisation/model/Permission.java
@@ -1,0 +1,19 @@
+package uk.gov.hmcts.darts.authorisation.model;
+
+import lombok.AllArgsConstructor;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/darts/authentication/model/Permission.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.hmcts.darts.authentication.model;
```
richardbull, line 1: I think the model package (3 classes) here should be `authorisation` instead of authentication?

karen-hedges: Good spot

richardbull (approved):

PR #297, opened 2023-07-31
DMP-700_Implement_REST_endpoint_for_get_case_by_ID

marcmollins:
https://tools.hmcts.net/jira/browse/DMP-700

```
[ ] Yes
[x ] No
```


src/main/resources/openapi/cases.yaml
```diff
@@ -76,6 +76,31 @@ paths:
         '500':
           description: Internal Server Error
 
+  /cases/{caseId}:
+    get:
+      tags:
+        - Cases
+      summary: Allows the retrieval of a single case using a given caseId
+      description: |-
+        Retrieves a single hearing for a single caseId
```
cmagowan-hmcts, line 85: Can you fix this description?

src/main/resources/openapi/cases.yaml
```diff
@@ -224,6 +249,9 @@ components:
       required:
         - courthouse
       properties:
+        case_id:
```
cmagowan-hmcts, line 252: we should maybe remove this again, if we are not going to use it.

src/main/resources/openapi/cases.yaml
```diff
@@ -260,6 +288,45 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/defence'
+        retain_until:
```
cmagowan-hmcts, line 291: we should maybe remove this again, if we are not going to use it.

src/main/java/uk/gov/hmcts/darts/cases/controller/CaseController.java
```diff
@@ -109,4 +110,13 @@ public ResponseEntity<List<Hearing>> casesCaseIdHearingsGet(
         return new ResponseEntity<>(caseService.getCaseHearings(caseId), HttpStatus.OK);
     }
 
+    @Override
+    public ResponseEntity<SingleCase> casesCaseIdGet(
+        @Parameter(name = "caseId", description = "caseId is the internal cas_id of the case.", required = true, in = ParameterIn.PATH) @PathVariable("caseId") Integer caseId
```
chrisbellingham-hmcts, line 115: Do we need to repeat these annotations/constraints from the interface, will they not be inherited?

src/main/resources/openapi/cases.yaml
```diff
@@ -260,6 +288,45 @@ components:
...
+        courthouse:
+          type: string
+          example: SWANSEA
+        case_number:
+          type: string
+          example: CASE1001
+        upload_priority:
+          type: string
+          example: 10
+        defendantList:
```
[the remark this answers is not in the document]
cmagowan-hmcts: we have a slight problem with that, as defence cannot easily be pluralised, so it was decided to stick 'List' on the end of everything to make it consistent.

src/main/java/uk/gov/hmcts/darts/cases/controller/CaseController.java
marcmollins: yes, I believe so, but all other methods have these annotations - so should I update each method and remove the annotations?

cmagowan-hmcts (approved):

PR #302, opened 2023-08-02
DMP-713 - update entities to v40 database

cmagowan-hmcts:
update database to v40


src/main/java/uk/gov/hmcts/darts/common/entity/HearingEntity.java
```diff
@@ -69,7 +72,7 @@ public class HearingEntity {
 
     @ManyToMany(fetch = FetchType.LAZY)
```
richardbull, line 73: We can remove this LAZY fetch attribute - default for ManyToMany

src/main/java/uk/gov/hmcts/darts/common/entity/HearingEntity.java
```diff
@@ -58,9 +55,15 @@ public class HearingEntity {
...
+        joinColumns = {@JoinColumn(name = HEA_ID)},
+        inverseJoinColumns = {@JoinColumn(name = "jud_id")})
+    private List<JudgeEntity> judges = new ArrayList<>();
+
     @ManyToMany
     @JoinTable(name = "hearing_media_ae",
-        joinColumns = {@JoinColumn(name = "hea_id")},
+        joinColumns = {@JoinColumn(name = HEA_ID)},
         inverseJoinColumns = {@JoinColumn(name = "med_id")})
     private List<MediaEntity> mediaList;
```
richardbull, line 68: Probably need to make this similar conventions to the others? `= new ArrayList<>();`

src/main/java/uk/gov/hmcts/darts/common/entity/JudgeEntity.java
```diff
@@ -20,21 +18,16 @@
 @EqualsAndHashCode(callSuper = false)
```
richardbull, line 18: Do we need to specify the attribute `(callSuper = false)` - this might be a bit of a hangover when I was getting rid of a warning where we were using lombok data annotation and I was extending of `VersionedEntity`.

We recently changed to using `@Getter @Setter` instead because of some issue Tony knows about. Equals contracts in entities are notoriously tricky. :)

src/main/java/uk/gov/hmcts/darts/common/entity/HearingEntity.java
```diff
@@ -58,9 +55,15 @@ public class HearingEntity {
     @Column(name = "judge_hearing_date")
     private String judgeHearingDate;
 
+    @ManyToMany(fetch = FetchType.EAGER, cascade = {CascadeType.PERSIST, CascadeType.MERGE})
```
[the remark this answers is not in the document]
cmagowan-hmcts: As discussed, this was intentional as when we retrieve a hearing or a case, we always want that Judges retrieved also.

PR #306, opened 2023-08-03
DMP-639 Integrate Authorisation with Authentication

richardbull:

https://tools.hmcts.net/jira/browse/DMP-639

### Change description ###

Validate Azure Active Directory B2C access token - JWT required claims: [aud, iss, exp, iat, sub, emails]
B2C_1_darts_externaluser_signin | Application claims
* Display Name String Display Name of the User.
* Email Addresses StringCollection Email addresses of the user.
* Given Name String The user's given name (also known as first name).
* Surname String The user's surname (also known as family name or last name).

```
[ ] Yes
[x] No
```


hemantasharma1129 (approved):

src/main/java/uk/gov/hmcts/darts/authentication/controller/impl/AuthenticationExternalUserController.java
```diff
@@ -1,22 +1,32 @@
...
 @Slf4j
 @RestController
 @RequestMapping("/external-user")
 @RequiredArgsConstructor
 public class AuthenticationExternalUserController implements AuthenticationController {
 
+    private static final String EMAILS_CLAIM_NAME = "emails";
+
     private final AuthenticationService authenticationService;
+    private final AuthorisationService authorisationService;
```
chrisbellingham-hmcts, line 29: As `AuthorisationService` lives in a seaparte feature, we shouldn't access it directly. Instead, we should expose/wrap the needed functionality via an "api" interface within the `authorisation` feature. Any code in `authentication` would then use that api interface to call the needed functionality.

This pattern is intended to make inter-feature coupling explict and manageable. I will add a section to the Best Practices document on this.

richardbull: @chrisbellingham-hmcts Done in commit f466057

chrisbellingham-hmcts (approved):

PR #314, opened 2023-08-07
DMP-182 Create spring scheduler task to process daily lists

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-182

### Change description ###

Added code to enable automated cron tasks to be configured. Added stub for processing daily lists

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/repository/AutomatedTaskRepository.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/repository/AutomatedTaskRepository.java
@@ -4,7 +4,10 @@ import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.stereotype.Repository;
 import uk.gov.hmcts.darts.common.entity.AutomatedTaskEntity;
 
+import java.util.Optional;
+
 @Repository
 public interface AutomatedTaskRepository extends JpaRepository<AutomatedTaskEntity, Integer> {
 
+    Optional<AutomatedTaskEntity> findByTaskName(String taskName);
 }
--- a/src/main/java/uk/gov/hmcts/darts/task/service/impl/AutomatedTaskServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/task/service/impl/AutomatedTaskServiceImpl.java
@@ -1,0 +1,232 @@
+package uk.gov.hmcts.darts.task.service.impl;
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import net.javacrumbs.shedlock.core.LockProvider;
+import org.springframework.scheduling.TaskScheduler;
+import org.springframework.scheduling.Trigger;
+import org.springframework.scheduling.config.ScheduledTask;
+import org.springframework.scheduling.config.ScheduledTaskHolder;
+import org.springframework.scheduling.config.ScheduledTaskRegistrar;
+import org.springframework.scheduling.config.Task;
+import org.springframework.scheduling.config.TriggerTask;
+import org.springframework.scheduling.support.CronExpression;
+import org.springframework.scheduling.support.CronTrigger;
+import org.springframework.stereotype.Service;
+import uk.gov.hmcts.darts.common.entity.AutomatedTaskEntity;
+import uk.gov.hmcts.darts.common.exception.DartsApiException;
+import uk.gov.hmcts.darts.common.repository.AutomatedTaskRepository;
+import uk.gov.hmcts.darts.task.model.TriggerAndAutomatedTask;
+import uk.gov.hmcts.darts.task.runner.AutomatedTask;
+import uk.gov.hmcts.darts.task.runner.AutomatedTaskName;
+import uk.gov.hmcts.darts.task.runner.impl.AbstractLockableAutomatedTask;
+import uk.gov.hmcts.darts.task.runner.impl.ProcessDailyListAutomatedTask;
+import uk.gov.hmcts.darts.task.service.AutomatedTaskService;
+import uk.gov.hmcts.darts.task.status.AutomatedTaskStatus;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import static uk.gov.hmcts.darts.task.exception.AutomatedTaskSetupError.FAILED_TO_FIND_AUTOMATED_TASK;
+import static uk.gov.hmcts.darts.task.exception.AutomatedTaskSetupError.INVALID_CRON_EXPRESSION;
+import static uk.gov.hmcts.darts.task.runner.AutomatedTaskName.PROCESS_DAILY_LIST_TASK_NAME;
+
+
+/**
+ * Refer to https://docs.spring.io/spring-framework/reference/integration/scheduling.html#scheduling-cron-expression
+ * for details of spring cron expressions
+ */
+@Service
+@Slf4j
+@RequiredArgsConstructor
+@SuppressWarnings({"PMD.TooManyMethods", "PMD.TooFewBranchesForASwitchStatement", "PMD.SingularField"})
+public class AutomatedTaskServiceImpl implements AutomatedTaskService {
+
+    private final AutomatedTaskRepository automatedTaskRepository;
+
+    private final ScheduledTaskHolder taskHolder;
+
+    private final LockProvider lockProvider;
+
+    private final TaskScheduler taskScheduler;
+
+    private final Map<String, Trigger> taskTriggers = new ConcurrentHashMap<>();
+
+    @Override
+    public void configureAndLoadAutomatedTasks(ScheduledTaskRegistrar taskRegistrar) {
+        addProcessDailyListToTaskRegistrar(taskRegistrar);
+    }
+
+    @Override
+    public Optional<AutomatedTaskEntity> getAutomatedTaskEntityByTaskName(String taskName) {
+        return automatedTaskRepository.findByTaskName(taskName);
+    }
+
+    public String getAutomatedTaskCronExpression(AutomatedTask automatedTask) {
+        String cronExpression = getAutomatedTaskCronExpression(automatedTask.getTaskName());
+        if (cronExpression == null) {
+            log.info("Unable to find cron expression for task: {}", automatedTask.getTaskName());
+            throw new DartsApiException(INVALID_CRON_EXPRESSION);
+        }
+        return cronExpression;
+    }
+
+    public String getAutomatedTaskCronExpression(String taskName) {
+        String cronExpression;
+        Optional<AutomatedTaskEntity> automatedTaskEntity = getAutomatedTaskEntityByTaskName(taskName);
+        if (automatedTaskEntity.isPresent()) {
+            cronExpression = automatedTaskEntity.get().getCronExpression();
+            log.debug("Task: {} cron expression: {}", taskName, cronExpression);
+        } else {
+            throw new DartsApiException(FAILED_TO_FIND_AUTOMATED_TASK);
+        }
+        return cronExpression;
+    }
+
+    /**
+     * Updates the automated task cron expression or throws DartsApiException.
+     * Calling this without the automated task being cancelled first, and then reloaded means the cron expression will only get picked up
+     * after the next execution once the automated task has run and then calculates the next execution time. Cancelling
+     * the automated task first means it will use the given cronExpression if the cron expression is valid.
+     * @param taskName name of automated task
+     * @param cronExpression cron expression
+     * @return true if the automated task is successfully updated
+     */
+    @Override
+    public boolean updateAutomatedTaskCronExpression(String taskName, String cronExpression) {
+        Optional<AutomatedTaskEntity> automatedTaskEntity = getAutomatedTaskEntityByTaskName(taskName);
+        if (automatedTaskEntity.isPresent()) {
+            AutomatedTaskEntity automatedTask = automatedTaskEntity.get();
+            if (CronExpression.isValidExpression(cronExpression)) {
+                automatedTask.setCronExpression(cronExpression);
+                automatedTaskRepository.saveAndFlush(automatedTask);
+                log.debug("Updated the task {} with cron expression {}", taskName, cronExpression);
+            } else {
+                log.error("Unable to update the task {} with cron expression {}", taskName, cronExpression);
+                throw new DartsApiException(INVALID_CRON_EXPRESSION);
+            }
+        } else {
+            log.error("Failed to update the task {} with cron expression {}", taskName, cronExpression);
+            throw new DartsApiException(FAILED_TO_FIND_AUTOMATED_TASK);
+        }
+        return true;
+    }
+
+    @Override
+    public void reloadTaskByName(String taskName) {
+        switch (AutomatedTaskName.valueOfTaskName(taskName)) {
+            case PROCESS_DAILY_LIST_TASK_NAME -> rescheduleProcessDailyListAutomatedTask();
+            default -> throw new DartsApiException(FAILED_TO_FIND_AUTOMATED_TASK);
+        }
+    }
+
+    @Override
+    public boolean cancelAutomatedTaskAndUpdateCronExpression(String taskName, boolean mayInterruptIfRunning, String cronExpression) {
+        boolean cancelled = cancelAutomatedTask(taskName, mayInterruptIfRunning);
+        boolean updatedCronExpresssion = false;
+        if (cancelled) {
+            updatedCronExpresssion = updateAutomatedTaskCronExpression(taskName, cronExpression);
+            if (updatedCronExpresssion) {
+                reloadTaskByName(taskName);
+            } else {
+                throw new DartsApiException(FAILED_TO_FIND_AUTOMATED_TASK);
+            }
+        }
+        return cancelled && updatedCronExpresssion;
+    }
+
+    @Override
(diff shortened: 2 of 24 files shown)


src/main/resources/db/migration/common/V1_110__insert_automated_task_data.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO darts.automated_task (aut_id,task_name,task_description,cron_expression,cron_editable)
+VALUES (1,'ProcessDailyList','Process the latest daily list for each courthouse','0 2 * * * *',true);
```
cmagowan-hmcts, line 2: we might not want this running over the weekend.

chrisbellingham-hmcts:
Reviewing

src/main/java/uk/gov/hmcts/darts/common/service/impl/AutomatedTaskServiceImpl.java
```diff
@@ -0,0 +1,162 @@
+package uk.gov.hmcts.darts.common.service.impl;
```
cmagowan-hmcts, line 1: I'm not sure if this should be in common package.
Common package is for common code that other modules use. I don't think anything in the common module uses code outwith the common module.
I think introducing this code here that runs code in Daily List etc, might introduce a "tangle":-
https://structure101.com/help/java/studio/Content/restructure101/tangles.html

Perhaps we should move it to its own module.

src/main/resources/db/migration/common/V1_110__insert_automated_task_data.sql
richardbull, line 2: Do we need to restart the `aut_seq` sequence at the end of this script?

src/main/java/uk/gov/hmcts/darts/common/task/AutomatedTask.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.hmcts.darts.common.task;
```
richardbull, line 1: Wondering if the `task` package should be up one level and the various subpackages for service, exception also be moved there?

src/main/resources/db/migration/common/V1_110__insert_automated_task_data.sql
karen-hedges: Do you know if any courts are open on the weekend? Is the process of generating the daily list going to only work out the daily list for the open courts? Is there the possibility a case will be heard on a weekend so it would make sense to run even though most the time there won't be any hearings?

src/main/java/uk/gov/hmcts/darts/common/repository/AutomatedTaskRepository.java
```diff
@@ -4,7 +4,10 @@
 import org.springframework.stereotype.Repository;
 import uk.gov.hmcts.darts.common.entity.AutomatedTaskEntity;
 
+import java.util.Optional;
+
 @Repository
 public interface AutomatedTaskRepository extends JpaRepository<AutomatedTaskEntity, Integer> {
 
+    Optional<AutomatedTaskEntity> findByTaskName(String taskName);
```
richardbull, line 12: Please can you check you are happy with definition we have in the modern schema / base script?
`TABLE automated_task` it feels like task_name should be unique
If so, please can you involve Judith W. to help make that change to the v41+ schema

src/main/java/uk/gov/hmcts/darts/common/service/impl/AutomatedTaskServiceImpl.java
karen-hedges: So my initial thinking was the shared code would be in common and the actual tasks would be in their respective packages but it probably makes more sense to put them in own package

src/main/java/uk/gov/hmcts/darts/common/service/impl/AutomatedTaskServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+
+    public String getAutomatedTaskCronExpression(AutomatedTask automatedTask) {
+        String cronExpression = getAutomatedTaskCronExpression(automatedTask.getTaskName());
+        if (cronExpression == null) {
+            log.info("Unable to find cron expression for task: {}", automatedTask.getTaskName());
+        }
+        return cronExpression;
+    }
+
+    public String getAutomatedTaskCronExpression(String taskName) {
```
chrisbellingham-hmcts, line 66: Can this be made `private`?

src/main/java/uk/gov/hmcts/darts/common/service/impl/AutomatedTaskServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+
+    @Override
+    public void loadTaskByName(String taskName) {
+        switch (taskName) {
+            case ProcessDailyListAutomatedTask.PROCESS_DAILY_LIST_TASK_NAME -> loadProcessDailyListAutomatedTask();
+            default -> throw new DartsApiException(FAILED_TO_FIND_AUTOMATED_TASK);
+        }
+    }
+
+    public void loadProcessDailyListAutomatedTask() {
```
chrisbellingham-hmcts, line 103: Can this be made `private`?

src/main/java/uk/gov/hmcts/darts/common/service/impl/AutomatedTaskServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+        ProcessDailyListAutomatedTask processDailyListAutomatedTask = new ProcessDailyListAutomatedTask();
+        Trigger trigger = createAutomatedTaskTrigger(processDailyListAutomatedTask);
+        taskScheduler.schedule(processDailyListAutomatedTask, trigger);
+    }
+
+    private void loadProcessDailyListAutomatedTask(ScheduledTaskRegistrar taskRegistrar) {
+        ProcessDailyListAutomatedTask processDailyListAutomatedTask = new ProcessDailyListAutomatedTask();
+        Trigger trigger = createAutomatedTaskTrigger(processDailyListAutomatedTask);
+        taskRegistrar.addTriggerTask(processDailyListAutomatedTask, trigger);
+    }
```
chrisbellingham-hmcts, line 113: It appears there are two different ways to load a scheduled task - via the `taskScheduler` or via the `taskRegister`. Are two ways necessary, or can we inify on a single approach?

src/main/java/uk/gov/hmcts/darts/common/service/impl/AutomatedTaskServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+    private Trigger createCronTrigger(AutomatedTask automatedTask) {
+        return new Trigger() {
+            @Override
+            public Instant nextExecution(TriggerContext triggerContext) {
+                String cronExpression = getAutomatedTaskCronExpression(automatedTask);
+                CronTrigger crontrigger = new CronTrigger(cronExpression);
+                return crontrigger.nextExecution(triggerContext);
+            }
+        };
+    }
```
chrisbellingham-hmcts, line 161: Can be refactored to the following slightly simpler version using lambas:

```java
    private Trigger createCronTrigger(AutomatedTask automatedTask) {
        return triggerContext -> {
            String cronExpression = getAutomatedTaskCronExpression(automatedTask);
            var crontrigger = new CronTrigger(cronExpression);
            return crontrigger.nextExecution(triggerContext);
        };
    }
```

src/main/java/uk/gov/hmcts/darts/common/service/impl/AutomatedTaskServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+        Set<ScheduledTask> scheduledTasks = taskHolder.getScheduledTasks();
+        for (ScheduledTask scheduledTask : scheduledTasks) {
+            Task task = scheduledTask.getTask();
+            if (task instanceof TriggerTask) {
+                TriggerTask triggerTask = (TriggerTask) task;
+                if (cancelTriggerTask(taskName, scheduledTask, triggerTask, mayInterruptIfRunning)) {
+                    result = true;
+                    break;
+                }
+            }
```
chrisbellingham-hmcts, line 127: To improve readability, suggest eliminating the boolean flags by simply returning early, e.g.:
```java
    @Override
    public boolean cancelAutomatedTask(String taskName, boolean mayInterruptIfRunning) {
        Set<ScheduledTask> scheduledTasks = taskHolder.getScheduledTasks();
        for (ScheduledTask scheduledTask : scheduledTasks) {
            Task task = scheduledTask.getTask();
            if (task instanceof TriggerTask triggerTask) {
                if (cancelTriggerTask(taskName, scheduledTask, triggerTask, mayInterruptIfRunning)) {
                    return true;
                }
            }
        }
        return false;
    }
```

src/main/java/uk/gov/hmcts/darts/common/service/impl/AutomatedTaskServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+            AutomatedTask automatedTask = (AutomatedTask) triggerTask.getRunnable();
+            if (automatedTask.getTaskName().equals(taskName)) {
+                log.info("About to cancel task: " + taskName);
+                scheduledTask.cancel(mayInterruptIfRunning);
+                result = true;
+                taskTriggers.remove(taskName);
+            }
+        }
+        return result;
+    }
```
chrisbellingham-hmcts, line 150: To improve readability, suggest eliminating the boolean flags by simply returning early, e.g.:

```java
    private boolean cancelTriggerTask(String taskName, ScheduledTask scheduledTask, TriggerTask triggerTask, boolean mayInterruptIfRunning) {
        if (triggerTask.getRunnable() instanceof AutomatedTask) {
            AutomatedTask automatedTask = (AutomatedTask) triggerTask.getRunnable();
            if (automatedTask.getTaskName().equals(taskName)) {
                log.info("About to cancel task: " + taskName);
                scheduledTask.cancel(mayInterruptIfRunning);
                taskTriggers.remove(taskName);
                return true;
            }
        }
        return false;
    }
```

src/main/java/uk/gov/hmcts/darts/common/task/ProcessDailyListAutomatedTask.java
```diff
@@ -0,0 +1,22 @@
...
+    public static final String PROCESS_DAILY_LIST_TASK_NAME = "ProcessDailyList";
+
+    @Override
+    public String getTaskName() {
+        return PROCESS_DAILY_LIST_TASK_NAME;
+    }
+
+    @Override
+    public void run() {
+        log.info("Task: {} running at: {} {}", getTaskName(),
```
chrisbellingham-hmcts, line 19: Leave a TODO here to indicate this is just a temporary implementation. Ideally reference the ticket number that will implement this.

src/main/resources/application.yaml
```diff
@@ -139,7 +139,11 @@ darts:
         outbound: darts-outbound
   portal:
     url: https://darts-portal.staging.platform.hmcts.net
-
+  automated:
+    task:
+      thread:
+        pool:
+          size: 7
```
chrisbellingham-hmcts, line 146: The default value was `5` yet this is `7`, any reason for the difference?

src/main/resources/db/migration/common/V1_110__insert_automated_task_data.sql
chrisbellingham-hmcts, line 2: If we have this in our main migration SQL then I suppose all test environments will have this scheduled task enabled. Maybe check with test team whether this is desirable, as it may be better for them to only add these tasks to the DB when needed.

src/integrationTest/java/uk/gov/hmcts/darts/common/service/AutomatedTaskServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+@Slf4j
+class AutomatedTaskServiceTest extends IntegrationBase {
+    @Autowired
+    private AutomatedTaskService automatedTaskService;
+
+    @Autowired
+    private ScheduledTaskHolder scheduledTaskHolder;
+
+    @Test
+    void givenConfiguredTasksCancelTaskAndAddTaskBackWithRandomCronExpression() throws InterruptedException {
```
chrisbellingham-hmcts, line 39: With all the sleeps, this test took around 56 seconds to run, I think this is adding too much time onto our overall test execution time. Can anything be done to reduce this?

chrisbellingham-hmcts, line 39: I think this test is trying to do too much, so it's quite hard to read. Can be it split up into multiple tests, each testing (and asserting) a specific part of the functionality? E.g:

- A test to demonstrate task cancellation works
- A test to demonstrate updating the cron expression works
- A test to demonstrate reloading works
..etc

src/integrationTest/java/uk/gov/hmcts/darts/common/service/AutomatedTaskServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+        log.info("----- About to cancel task {}", automatedTask.getTaskName());
+        boolean taskCancelled = automatedTaskService.cancelAutomatedTask(automatedTask.getTaskName(), false);
+        assertTrue(taskCancelled);
+
+        Thread.sleep(21_000);
+        log.info("----- Tasks after sleep");
+        scheduledTasks = scheduledTaskHolder.getScheduledTasks();
+        displayTasks(scheduledTasks);
+        automatedTaskService.updateAutomatedTaskCronExpression(automatedTask.getTaskName(), getRandomCronExpression());
```
chrisbellingham-hmcts, line 59: Does the cron expression need to be random? Can we just specify a known cron value instead?

chrisbellingham-hmcts (commented):
Hi Karen, left some feedback. More generally, how does this solution support multiple instances of the application running the cluster? E.g. if we have two instances running, it feels like both will load (and execute) the same cron jobs. Similarly, if one instance updates a cron expression in the database is how will the other instances behave? Feels like it could be easy for instances to get out of sync with the source of truth (database), unless there is a mechanism to address that?

src/main/resources/db/migration/common/V1_110__insert_automated_task_data.sql
karen-hedges: I have raised the question about adding a new field enabled which would mean this could be different for each environment accordingly.

src/integrationTest/java/uk/gov/hmcts/darts/common/service/AutomatedTaskServiceTest.java
karen-hedges: I will modify this test to reduce the times. Initially it was so you could see the tasks running independantly and at different times.

src/main/java/uk/gov/hmcts/darts/common/service/impl/AutomatedTaskServiceImpl.java
karen-hedges: The taskRegister is only available by the configuration and is what contains the list of tasks even if they haven't run yet. The task scheduler which is available to the service can then start and stop tasks that have been loaded by the taskRegister. If you add just to task scheduler, you won't be able to amend it until has run for the first time.

davet1985:
I've been running the locally and I think it does the intended job
- updating the cron when it's changed in the database
- using shedlock lock during execution and releasing the lock appropriately

@karen-hedges can you review the [sonar results](https://sonarcloud.io/summary/new_code?id=uk.gov.hmcts.reform%3Adarts-api&pullRequest=314) please, there are some things that need addressing

I'd rather leave the more details review of the code to the senior java devs, some of whom have already reviewed I think.

src/main/java/uk/gov/hmcts/darts/task/service/impl/AutomatedTaskServiceImpl.java
```diff
@@ -0,0 +1,232 @@
...
+    @Override
+    public boolean cancelAutomatedTaskAndUpdateCronExpression(String taskName, boolean mayInterruptIfRunning, String cronExpression) {
+        boolean cancelled = cancelAutomatedTask(taskName, mayInterruptIfRunning);
+        boolean updatedCronExpresssion = false;
+        if (cancelled) {
+            updatedCronExpresssion = updateAutomatedTaskCronExpression(taskName, cronExpression);
+            if (updatedCronExpresssion) {
+                reloadTaskByName(taskName);
+            } else {
+                throw new DartsApiException(FAILED_TO_FIND_AUTOMATED_TASK);
```
chrisbellingham-hmcts, line 134: Is this `else` branch is reachable?

Looks like the only possible outcomes of `updateAutomatedTaskCronExpression()` are:
- `true`
-  or an exception is thrown

src/main/java/uk/gov/hmcts/darts/common/repository/AutomatedTaskRepository.java
karen-hedges: Changes to the database will be done in a separate ticket.

src/main/java/uk/gov/hmcts/darts/task/service/impl/AutomatedTaskServiceImpl.java
karen-hedges: this has changed

chrisbellingham-hmcts (approved):

PR #335, opened 2023-08-14
DMP-708 ATS Download Zip file - To link the logic of Metadata

richardbull:

[DMP-708](https://tools.hmcts.net/jira/browse/DMP-708)

### Change description ###

* Update generateZipStructure to add readMe.txt
* Update generateZipStructure to add playlist.xml
* Update generateZipStructure to add annotations.xml
* Update ViqMetaData / PlaylistInfo / ViqAnnotationData to use ZonedDateTime [Europe/London]
ensure that playlist and annotations H attribute uses London time
If you need to store data and time, use OffsetDateTime and to present date and time to users, use the ZonedDataTime

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/audio/component/impl/AnnotationXmlGeneratorImpl.java
```diff
@@ -21,7 +23,7 @@
 @Slf4j
 public class AnnotationXmlGeneratorImpl extends AbstractDocumentGenerator {
 
-    private static final String ANNOTATION_FILENAME = "annotations.xml";
+    private static final ZoneId EUROPE_LONDON_ZONE = ZoneId.of("Europe/London");
```
karen-hedges, line 26: I'm wondering we should all be using the same time constants and so this should be in common perhaps?

src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileZipGeneratorImpl.java
```diff
@@ -51,16 +78,71 @@ public Path generateAndWriteZip(List<List<AudioFileInfo>> audioSessions) {
         }
     }
 
-    private Map<Path, Path> generateZipStructure(List<List<AudioFileInfo>> audioSessions) {
+    private ViqMetaData createViqMetaData(MediaRequestEntity mediaRequestEntity) {
+        return ViqMetaData.builder()
+            .courthouse(mediaRequestEntity.getHearing().getCourtroom().getCourthouse().getCourthouseName())
+            .raisedBy(null)
```
karen-hedges, line 84: Should the raised by be null?

richardbull: That's my understanding currently when I refer to the readme in the other ticket [DMP-588](https://tools.hmcts.net/jira/browse/DMP-588) ATS - VIQ Header - Readme.txt file

readMe.txt shows `Raised by: `

https://github.com/hmcts/darts-legacy/blob/2e512caf650aeb532fbdad539a714cd191a7144c/workspace/dartsdev/darts-ats/src/main/java/com/cgi/darts/ats/MojDownloadRequest.java#L259C27-L259C36 shows that it appears to only being populated for Transcription Requests.

It seems to also be missing the empty lines between the fields - just an observation

src/main/java/uk/gov/hmcts/darts/audio/component/impl/AnnotationXmlGeneratorImpl.java
richardbull: Updated in commit e339bbc - add DateConverters (similar to https://github.com/hmcts/darts-gateway/pull/90/files#r1294794974)

BST can be very frustrating!  I came to the conclusion today that using `ZonedDateTime` to present date and time to users correctly in BST was needed in darts-api only for the ATS Download Zip metadata files.

To store data and time, we use `OffsetDateTime` in UTC and to present date and time to users, we should use the `ZonedDateTime` for the legacy interfaces. According to the restful api standards for the darts-api we expect request / response for date-time strings in JSON to be in UTC.

See also [5. Common format for timestamps](https://tools.hmcts.net/confluence/display/DMP/5.+Common+format+for+timestamps?src=contextnavpagetreemode) - for initial decision on choosing types for our database / JPA

chrisbellingham-hmcts (approved):

PR #338, opened 2023-08-15
DMP-746 - stop courthouse being added if one exists already with a di…

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-746

Have also given Test classes their own PMD rules


tonymort (approved):
LGTM

PR #351, opened 2023-08-17
DMP-753_Add_error_response_to_GET_case_hearings_endpoint

marcmollins:
https://tools.hmcts.net/jira/browse/DMP-753

### Change description ###
updated CaseApiError.java with another Enum for Hearing data not found. added error handling within service class for an empty array from repository and throws a DartsAPIException

also updated tests to reflect changes - these pass

```
[ ] Yes
[ x] No
```


src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerGetCaseHearingsTest.java
```diff
@@ -96,7 +96,7 @@ void casesSearchGetEndpointNoObjectsReturned() throws Exception {
 
         MockHttpServletRequestBuilder requestBuilder = get(endpointUrl, "25");
```
cmagowan-hmcts, line 97: Can we have a test where the case exists but no hearings?

PR #354, opened 2023-08-20
DMP-761 Request Audio - Get Event endpoint

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-761

### Change description ###

Added the getEvent endpoint

```
[ ] Yes
[ X] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerGetEventsTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerGetEventsTest.java
@@ -1,0 +1,99 @@
+package uk.gov.hmcts.darts.cases.controller;
+
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.test.web.servlet.MockMvc;
+import org.springframework.test.web.servlet.request.MockHttpServletRequestBuilder;
+import org.springframework.test.web.servlet.result.MockMvcResultMatchers;
+import uk.gov.hmcts.darts.common.entity.CourtCaseEntity;
+import uk.gov.hmcts.darts.common.entity.CourthouseEntity;
+import uk.gov.hmcts.darts.common.entity.CourtroomEntity;
+import uk.gov.hmcts.darts.common.entity.EventEntity;
+import uk.gov.hmcts.darts.common.entity.EventHandlerEntity;
+import uk.gov.hmcts.darts.common.entity.HearingEntity;
+import uk.gov.hmcts.darts.common.entity.JudgeEntity;
+import uk.gov.hmcts.darts.testutils.IntegrationBase;
+import uk.gov.hmcts.darts.testutils.data.EventHandlerTestData;
+
+import java.time.LocalDate;
+import java.time.OffsetDateTime;
+
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static uk.gov.hmcts.darts.testutils.data.CaseTestData.createCaseAt;
+import static uk.gov.hmcts.darts.testutils.data.CourthouseTestData.someMinimalCourthouse;
+import static uk.gov.hmcts.darts.testutils.data.CourtroomTestData.createCourtRoomWithNameAtCourthouse;
+import static uk.gov.hmcts.darts.testutils.data.EventTestData.createEventWith;
+import static uk.gov.hmcts.darts.testutils.data.HearingTestData.createHearingWithDefaults;
+import static uk.gov.hmcts.darts.testutils.data.JudgeTestData.createJudgeWithName;
+
+@AutoConfigureMockMvc
+class CaseControllerGetEventsTest  extends IntegrationBase {
+
+    public static final String EVENT_TEXT1 = "event1";
+    @Autowired
+    private transient MockMvc mockMvc;
+
+    private static String endpointUrl = "/cases/hearings/{hearingId}/events";
+
+    private static final OffsetDateTime SOME_DATE_TIME = OffsetDateTime.parse("2023-01-01T12:00Z");
+    private static final String SOME_COURTHOUSE = "some-courthouse";
+    private static final String SOME_COURTROOM = "some-courtroom";
+
+    private HearingEntity hearing1;
+    private EventEntity event1;
+    private EventHandlerEntity activeHandler;
+
+    @BeforeEach
+    void setupData() {
+        CourthouseEntity swanseaCourthouse = someMinimalCourthouse();
+        swanseaCourthouse.setCourthouseName(SOME_COURTHOUSE);
+
+        CourtroomEntity courtroom1 = createCourtRoomWithNameAtCourthouse(swanseaCourthouse, SOME_COURTROOM);
+
+        CourtCaseEntity case1 = createCaseAt(swanseaCourthouse);
+        case1.setCaseNumber("Case1");
+
+        JudgeEntity judge = createJudgeWithName("aJudge");
+        hearing1 = createHearingWithDefaults(case1, courtroom1, LocalDate.of(2023, 5, 20), judge);
+
+        dartsDatabase.saveAll(hearing1);
+
+        activeHandler = EventHandlerTestData.someMinimalEventHandler();
+
+        dartsDatabase.saveAll(activeHandler);
+
+        event1 = createEventWith("eventName", EVENT_TEXT1, hearing1, OffsetDateTime.now(), activeHandler);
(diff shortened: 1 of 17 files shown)


src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerGetEventsTest.java
```diff
@@ -0,0 +1,111 @@
...
+    @AfterEach
+    void tearDown() {
+        dartsDatabase.addToTrash(activeHandler);
+    }
+
+    @Test
+    void casesGetEventsGetEndpoint() throws Exception {
+        MockHttpServletRequestBuilder requestBuilder = get(endpointUrl, case1.getId(), hearing1.getId());
+        mockMvc.perform(requestBuilder).andExpect(MockMvcResultMatchers.status().isOk());
+    }
```
chrisbellingham-hmcts, line 85: Suggest this be removed as the test case is already covered by `casesGetEvents()`

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerGetEventsTest.java
```diff
@@ -0,0 +1,111 @@
...
+    void casesGetEventsWithInvalidHearingIdReturnsNotFoundEndpoint() throws Exception {
+        MockHttpServletRequestBuilder requestBuilder = get(endpointUrl, case1.getId(), 456);
+        mockMvc.perform(requestBuilder).andExpect(MockMvcResultMatchers.status().isNotFound());
+    }
+
+    @Test
+    void casesGetEvents() throws Exception {
+        MockHttpServletRequestBuilder requestBuilder = get(endpointUrl, case1.getId(), hearing1.getId());
+
+        mockMvc.perform(requestBuilder).andExpect(MockMvcResultMatchers.status().isOk())
```
chrisbellingham-hmcts, line 103: Consider adding an assertion on the length of the returned array (I think expectation is length == 1).

src/main/resources/openapi/cases.yaml
```diff
@@ -450,3 +485,29 @@ components:
...
+        $ref: '#/components/schemas/event_response'
+    event_response:
+      type: object
+      properties:
+        id:
+          type: integer
+          example: 1
+        timestamp:
+          type: string
+          format: date
```
richardbull, line 502: Please use `format: date-time`

src/main/resources/openapi/cases.yaml
```diff
@@ -237,6 +237,41 @@ paths:
               schema:
                 $ref: './problem.yaml'
 
+
+  /cases/{caseId}/hearings/{hearingId}/events:
+    get:
+      tags:
+        - Cases
+      summary: Allows the retrieval of events for a given caseId and hearingId
```
richardbull, line 245: Think we should think about adding a simple unique `operationId: getEvents` and have it the same name in the Controller?  I just find it easier to search for the text "Find in Files..." that way

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
```diff
@@ -130,5 +133,18 @@ public List<AdvancedSearchResult> advancedSearch(GetCasesSearchRequest request)
         return AdvancedSearchResponseMapper.mapResponse(hearings);
     }
 
-
+    public List<EventResponse> getEvents(Integer caseId, Integer hearingId) {
+        List<Hearing> hearings = getCaseHearings(caseId);
+        if (hearings.isEmpty()) {
+            throw new DartsApiException(CaseApiError.CASE_NOT_FOUND);
```
richardbull, line 139: Is this the right error for empty hearings?

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
```diff
@@ -130,5 +133,18 @@ public List<AdvancedSearchResult> advancedSearch(GetCasesSearchRequest request)
...
+        }
+        Hearing hearing = hearings
+            .stream()
+            .filter(hearingEntity -> hearingEntity.getId().equals(hearingId))
```
richardbull, line 143: It feels like we should probably add a new finder service here?

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
```diff
@@ -130,5 +133,18 @@ public List<AdvancedSearchResult> advancedSearch(GetCasesSearchRequest request)
...
+            .findFirst().orElseThrow(
+                () -> new DartsApiException(CaseApiError.HEARING_NOT_FOUND)
+            );
+
+        return casesMapper.mapToEvents(eventRepository.findAllByHearingId(hearing.getId()));
```
richardbull, line 148: I'm not entirely sure why we need the case_id param here if we choose a hearing_id as once we have a hearing, we will know the single case. We should probably simplify the service here.
FYI @hemantasharma1129

src/main/java/uk/gov/hmcts/darts/common/entity/EventEntity.java
```diff
@@ -35,7 +36,7 @@ public class EventEntity extends VersionedEntity {
     @Column(name = "event_object_id", unique = true, length = 16)
     private String legacyObjectId;
 
-    @ManyToOne
+    @ManyToOne(fetch = FetchType.EAGER, optional = false)
     @JoinColumn(name = "evh_id")
     private EventHandlerEntity eventType;
```
richardbull, line 41: Since we're touching this relationship - would you be able to correct the field please so that it represents the new entity name - `eventHandler`?

src/main/java/uk/gov/hmcts/darts/cases/mapper/CasesMapper.java
```diff
@@ -103,6 +106,18 @@ public SingleCase mapToSingleCase(CourtCaseEntity caseEntity) {
...
     }
 
+    public List<EventResponse> mapToEvents(List<EventEntity> eventEntities) {
+        return emptyIfNull(eventEntities).stream().map(this::mapToEvent).collect(Collectors.toList());
+    }
+
+    public EventResponse mapToEvent(EventEntity eventEntity) {
+        EventResponse event = new EventResponse();
+        event.setId(eventEntity.getId());
+        event.setTimestamp(eventEntity.getTimestamp().toLocalDate());
```
richardbull, line 116: This doesn't make sense for `timestamp` - the `toLocalDate()` will need to be removed when you update the openapi yaml to be `date-time` format

src/main/resources/openapi/cases.yaml
```diff
@@ -450,3 +485,29 @@ components:
...
+      properties:
+        id:
+          type: integer
+          example: 1
+        timestamp:
+          type: string
+          format: date
+          example: 2023-07-31T14:32:24.620Z
+        name:
+          type: string
```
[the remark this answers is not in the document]
richardbull: Not sure it's needed here - as it's a response object using the name from the `event_handler` table?

src/main/resources/openapi/cases.yaml
```diff
@@ -450,3 +485,29 @@ components:
...
+          example: Case called on
+        text:
+          type: string
```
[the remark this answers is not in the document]
hemantasharma1129: Is there a length limitation on legacy column? I would suggest to have it without any length constraint and trim it in the dashboard column and show the complete value using a tooltip on the dashboard to save some real estate?

src/main/resources/openapi/cases.yaml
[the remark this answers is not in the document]
hemantasharma1129: Is there a length limitation on legacy column? I would suggest to have it without any length constraint and trim it in the dashboard column and show the complete value using a tooltip on the dashboard to save some real estate?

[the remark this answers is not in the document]
chrisbellingham-hmcts: Indeed it's response attribute so no need for a length constrain, please disregard my comment.

src/main/resources/openapi/cases.yaml
karen-hedges: What is operationId: getEvents?

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
karen-hedges: This has been simplified since we don't need case Id.

src/main/java/uk/gov/hmcts/darts/common/entity/EventEntity.java
karen-hedges: This requires quite a few changes so can be moved to another ticket

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
hemantasharma1129: Agree, we dont need case_id as inout param.

chrisbellingham-hmcts (approved):

PR #367, opened 2023-08-23
DMP-759 Audio Requests API: GET /audio-requests

richardbull:
### JIRA link (if applicable) ###

[DMP-759](https://tools.hmcts.net/jira/browse/DMP-759)

### Change description ###

Audio Requests API [getYourAudio](http://localhost:4550/swagger-ui/index.html#/Audio%20Requests/getYourAudio) 
* `GET /audio-requests` Users can view a summary of their audio requests and status
* Update Table "darts.media_request" to add `expiry_ts timestamp with time zone`
* MediaRequestService.viewAudioRequests - add `EXPIRED` media_request_status

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #359, opened 2023-08-21
DMP-736 Changes to device_register table

karen-hedges:
### JIRA link (if applicable) ###

https://tools.hmcts.net/jira/browse/DMP-736

### Change description ###

Renamed device_register table node_register. 
Removed primary key. 
Made node_id not null. 
Made node_id primary key. 
Created sequence.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[ ] No
```

PR #375, opened 2023-08-24
Add files via upload

barry-jones50410:

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -782,13 +801,15 @@ IS 'inherited from dm_sysobject_r, for r_object_type of moj_media';
 CREATE TABLE media_request
 (mer_id                      INTEGER                       NOT NULL
 ,hea_id                      INTEGER                       NOT NULL
-,requestor                   INTEGER                       NOT NULL  -- FK to moj_user.moj_usr_id
-,request_status              CHARACTER VARYING
-,request_type                CHARACTER VARYING
+,requestor                   INTEGER                       NOT NULL  -- FK to user_account
+,current_owner               INTEGER                       NOT NULL  -- FK to user_account
```
richardbull, line 805: @barry-jones50410 What's this field for? I don't have it currently in my pull request for media_request table updates in flyway.  FYI @hemantasharma1129 

https://github.com/hmcts/darts-api/pull/367/files#diff-8be848af5cd5da457379f14c0bb81ab6e243dbe62ef5377f7c0ac997761a8660

hemantasharma1129: @barry-jones50410 Is the new column meant for change of ownership functionality in Admin?

barry-jones50410: My understanding is that the change, is to allow for the media_request  where the original requestor may now differ from the person expected to receive the media requested, by recording current_owner. That is  the media when its made available, is accessible to the current_owner as opposed to the original requestor,

src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -482,9 +498,9 @@ COMMENT ON COLUMN defendant.cas_id
 IS 'foreign key from court_case';
 
 CREATE TABLE device_register
```
richardbull, line 500: Looking at @karen-hedges PR from earlier - she was expecting the table to be named `node_register`?
https://github.com/hmcts/darts-api/pull/359/files#diff-dfb5b8a34cd01a3fe9a384a7def51f31c027fb5f51c0e3b7c77143f2c219e828R5

src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -482,9 +498,9 @@ COMMENT ON COLUMN defendant.cas_id
 IS 'foreign key from court_case';
 
 CREATE TABLE device_register
-(der_id                      INTEGER                       NOT NULL
+(node_id                    INTEGER                       NOT NULL
```
richardbull, line 501: Are we missing a sequence for this table? It looks like @karen-hedges was expecting 
`CREATE SEQUENCE nod_seq START WITH 10000;`

https://github.com/hmcts/darts-api/pull/359/files#r1304035986

barry-jones50410:
Classification: Official

In naming this table, I believe there was a decision to move away from legacy ( tbl_moj_node) and use "device" in preference to "node".

Device_register has been the modernised equivalent from the outset.  Only  agreed change has been to remove the synthetic sequence generate PK column der_id, and utilise node_id as the PK as this is already accepted column name and known to be unique.

barry-jones50410:
Classification: Official

The table name is "device_register" and the sequence created to create candidate values for the PK was "der_seq".
We have accepted that der_id will be effectively renamed to "node_id".  I would like to minimise the degree to which we deviate from the pattern seen across the schema, by allowing this non-default PK column name.
I was not intending to propagate that exception wider, to include renaming the sequence from der_seq to nod_seq

barry-jones50410:
Classification: Official

Tbl_moj_node has all nullable columns, with the exception of node_id
In practice hostname and ip_address contain no nulls
Can add not null as necessary, to v47 please confirm

richardbull:
RE: https://github.com/hmcts/darts-api/pull/375#issuecomment-1691720649

Yes please @barry-jones50410 - I'd think it makes sense to make columns `NOT NULL` as soon as we know.

barry-jones50410:
Classification: Official

Hostname and ip_address now NN
Mac_address is the only remaining nullable column.
Legacy has 0 null records, but is inconsistent with format, most use "-" to delimit elements of the address, some don't, some have unnecessary whitespace, so I think data could easily be cleaned, def  suitable for not null, so will change this to NN

richardbull (approved):

PR #379, opened 2023-08-24
Dmp 760 delete requested and processed audio

garethmorriscgi:
https://tools.hmcts.net/jira/browse/DMP-760

### Change description ###
Services endpoint that handles the deletion of audio request details and if present any audio blob data from outbound storage container and the linking transient object

```
[ ] Yes
[X] No
```


src/main/resources/openapi/audio.yaml
```diff
@@ -29,6 +29,31 @@ paths:
...
         description: Audio Request item to add
+    delete:
+      tags:
+        - Audio
+      summary: Delete Audio Request and Processed Audio
+      description: Delete Audio Request and Processed Audio
+      operationId: deleteAudioRequest
+      parameters:
+        - in: query
+          name: audioRequestId
```
richardbull, line 40: Query params should be lower snake case
but I think this should really be a path param to be similar to `DELETE
[/courthouses/{courthouse_id}](http://localhost:4550/swagger-ui/index.html#/Courthouses/courthousesCourthouseIdDelete)`

src/main/resources/openapi/audio.yaml
```diff
@@ -29,6 +29,31 @@ paths:
             schema:
               $ref: '#/components/schemas/AudioRequestDetails'
         description: Audio Request item to add
+    delete:
```
richardbull, line 32: I think this endpoint should go along the new one I added this sprint
[audio-requests.yaml](https://github.com/hmcts/darts-api/blob/master/src/main/resources/openapi/audio-requests.yaml)

DELETE `/audio-requests/{id}` ?

FYI @hemantasharma1129

src/main/resources/openapi/audio.yaml
```diff
@@ -29,6 +29,31 @@ paths:
...
+      description: Delete Audio Request and Processed Audio
+      operationId: deleteAudioRequest
+      parameters:
+        - in: query
+          name: audioRequestId
+          required: true
+          schema:
+            $ref: '#/components/schemas/AudioRequestId'
+      responses:
+        '200':
```
richardbull, line 45: I was expecting a `204` response code - but the ticket does have `200`. Maybe worth checking with @hemantasharma1129

src/main/resources/openapi/audio.yaml
```diff
@@ -29,6 +29,31 @@ paths:
...
+          description: audio request has been deleted
+        '500':
```
richardbull, line 47: Do we need to specify `500` response - what errors do you have coming back, is `400` Bad Request missing?

src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
```diff
@@ -38,6 +38,12 @@ public ResponseEntity<Void> addAudioRequest(AudioRequestDetails audioRequestDeta
         return new ResponseEntity<>(HttpStatus.CREATED);
     }
 
+    public ResponseEntity<Void> deleteAudioRequest(Integer audioRequestId) {
```
richardbull, line 41: If Hemanta agrees with the base path being `audio-requests` pluralised - then I think the new services we do this sprint should go under `AudioRequestsController`

src/test/java/uk/gov/hmcts/darts/audio/service/impl/MediaRequestServiceImplTest.java
```diff
@@ -5,18 +5,25 @@
...
+import uk.gov.hmcts.darts.common.repository.TransientObjectDirectoryRepository;
 import uk.gov.hmcts.darts.common.repository.UserAccountRepository;
+import uk.gov.hmcts.darts.datamanagement.api.impl.DataManagementApiImpl;
 
 import java.time.OffsetDateTime;
+import java.util.Optional;
+import java.util.UUID;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.eq;
```
richardbull, line 26: Can you check SonarQube - I have a feeling I did this style before and spotted it was a code smell
https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&sinceLeakPeriod=true&pullRequest=379&id=uk.gov.hmcts.reform%3Adarts-api

src/test/java/uk/gov/hmcts/darts/audio/service/impl/MediaRequestServiceImplTest.java
```diff
@@ -90,4 +100,51 @@ void whenSavingAudioRequestIsSuccessful() {
...
+
+    @Test
+    void whenNoAudioIsPresentOnlyDeleteAudioRequest() {
+        var mediaRequestId = 1;
+        when(transientObjectDirectoryRepository.getTransientObjectDirectoryEntityByMediaRequest_Id(mediaRequestId))
+            .thenReturn(Optional.empty());
+
+        mediaRequestService.deleteAudioRequest(mediaRequestId);
+
+        verify(mockMediaRequestRepository, Mockito.times(1)).deleteById(eq(mediaRequestId));
```
richardbull, line 145: I think 1 times is the default, so can be simplified FYI

src/main/resources/openapi/audio.yaml
hemantasharma1129: It should be part of AudioController which has endpoints for requesting an Audio, downloading an Requested Audio etc. AudioRequestController may no longer be required? We can discuss about it.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/MediaRequestServiceImpl.java
```diff
@@ -74,6 +80,26 @@ public Integer saveAudioRequest(AudioRequestDetails request) {
...
+            UUID blobId = mediaTransientObject.getExternalLocation();
+
+            if (blobId != null) {
+                dataManagementApi.deleteBlobDataFromOutboundContainer(blobId);
+            }
+
+            transientObjectDirectoryRepository.deleteById(mediaTransientObject.getId());
+        }
+
+        mediaRequestRepository.deleteById(mediaRequestId);
```
hemantasharma1129, line 100: what happens if the requestId which we are trying to delete doesnt exist in Media Request table? Which error the JPA repository thows?

hemantasharma1129 (approved):

PR #402, opened 2023-09-04
DMP-841: Support internal user login

mestebanez:

[DMP-841](https://tools.hmcts.net/jira/browse/DMP-841)

### Change description ###

* Add a auth configuration strategy pattern to allow configurations to be obtained based on the incoming payload (url name by default)
* Support for a fallback mechanism if the payload can't match to any auth configuration (Currently configured to error back to the user) 
* Replace feign token endpoint interaction for nimbus. This seems0 consistent with the token validator
* Use of Spring @configurationproperties properties to allow Spring to get all auth configuration properties for us avoiding @value
* All tests have been modified and pass successfully to support this new pattern

```
[ ] Yes
[X ] No
```

**Omissions

* Functional tests need to be written in order to prove that we can talk to the internal user idp


src/main/java/uk/gov/hmcts/darts/authentication/config/AuthProviderConfigurationProperties.java
```diff
@@ -0,0 +1,32 @@
...
+
+    String getLogoutUri();
+
+    String getResetPasswordUri();
+
+    default JWKSource<SecurityContext> getJwkSource() {
+        try {
+            URL jwksUrl = new URL(getJwkSetUri());
+
+            return JWKSourceBuilder.create(jwksUrl).build();
```
mestebanez, line 27: Maybe better to have this in a separate PR. This was changed as the builder is now mandated by nimbus API.  Our previous usage is deprecated. For example:-

new RemoteJWKSet<>(jwksUrl);

src/main/java/uk/gov/hmcts/darts/authentication/client/impl/OAuthClientImpl.java
```diff
@@ -0,0 +1,43 @@
...
+import uk.gov.hmcts.darts.authentication.client.OAuthClient;
+import uk.gov.hmcts.darts.authentication.config.AuthProviderConfigurationProperties;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+
+
+@Component
+public class OAuthClientImpl implements OAuthClient {
```
mestebanez, line 23: Replace the use of feign as:-

1) This class avoids our code being bound to a specific yaml key. 
2) The use of nimbus is consistent with the way in which we validate tokens.

src/main/java/uk/gov/hmcts/darts/authentication/dao/AzureDao.java
```diff
@@ -1,8 +1,11 @@
 package uk.gov.hmcts.darts.authentication.dao;
 
+import uk.gov.hmcts.darts.authentication.config.AuthConfigurationProperties;
+import uk.gov.hmcts.darts.authentication.config.AuthProviderConfigurationProperties;
 import uk.gov.hmcts.darts.authentication.exception.AzureDaoException;
 import uk.gov.hmcts.darts.authentication.model.OAuthProviderRawResponse;
 
 public interface AzureDao {
```
mestebanez, line 8: Maybe good to also make this interface more generic so we remain agnostic of any paticular cloud provider e.g. idpDAO?

src/main/java/uk/gov/hmcts/darts/authentication/dao/impl/AzureDaoImpl.java
```diff
@@ -1,58 +1,51 @@
...
-import java.nio.charset.StandardCharsets;
 
 @Component
 @Slf4j
 @RequiredArgsConstructor
 public class AzureDaoImpl implements AzureDao {
 
-    private final AuthenticationConfiguration authConfig;
-    private final AzureActiveDirectoryB2CClient azureActiveDirectoryB2CClient;
+    private final OAuthClient azureActiveDirectoryB2CClient;
```
mestebanez, line 24: Variable name probably needs to change

src/main/java/uk/gov/hmcts/darts/common/util/RequestMatcher.java
```diff
@@ -0,0 +1,29 @@
...
+
+    String externalUrlMatcher = "/external-user";
+
+    RequestMatcher URL_MAPPER_INTERNAL = (req) -> {
+        if (req.getRequestURL().toString().contains(internalUrlMatcher)) {
+            return true;
+        }
+
+        return false;
+    };
```
chrisbellingham-hmcts, line 18: Consider simplifying to:

```java
    RequestMatcher URL_MAPPER_INTERNAL = (req) -> req.getRequestURL().toString().contains(internalUrlMatcher);

```

src/main/java/uk/gov/hmcts/darts/common/util/RequestMatcher.java
```diff
@@ -0,0 +1,29 @@
...
+
+    RequestMatcher URL_MAPPER_EXTERNAL = (req) -> {
+        if (req.getRequestURL().toString().contains(externalUrlMatcher)) {
+            return true;
+        }
+
+        return false;
+    };
```
chrisbellingham-hmcts, line 26: As above, consider simplifying to a lambda expression.

src/main/java/uk/gov/hmcts/darts/authentication/config/AuthenticationConfiguration.java
```diff
@@ -1,58 +1,16 @@
...
-    @Value("${spring.security.oauth2.client.registration.external-azure-ad.issuer-uri}")
-    private String externalADissuerUri;
-
-    @Value("${spring.security.oauth2.client.provider.external-azure-ad-provider.reset-password-uri}")
-    private String externalADresetPasswordUri;
-
+    @Bean
+    public AuthConfigFallback getDefaultNoAuthConfiguration() {
+        return (req) -> {
+            throw new AuthenticationException("No authentication configuration found");
```
chrisbellingham-hmcts, line 13: Preferred error-handling pattern is to throw a `DartsApiException` with the appropriate enum value as a param.

Can you replace `throw new AuthenticationException` with `throw new DartsApiException(AuthenticationError.FAILED_TO_OBTAIN_AUTHENTICATION_CONFIG)`, and add the new enum value to `src/main/java/uk/gov/hmcts/darts/authentication/exception/AuthenticationError.java` along with a unique error code?

Would also be worth removing (or marking as deprecated) the `AuthenticationException` class as this should no longer be used.

src/main/java/uk/gov/hmcts/darts/authentication/client/impl/OAuthClientImpl.java
chrisbellingham-hmcts: Concious that HMCTS standards have a preference for Feign (https://hmcts.github.io/standards/standards/java.html), so it may be better to keep our http clients consistent with that, unless we get any compelling benefits from using nimbus for this one use case?

chrisbellingham-hmcts (commented):
Hi @mestebanez, I like the pattern. 👍

src/main/java/uk/gov/hmcts/darts/authentication/client/impl/OAuthClientImpl.java
hemantasharma1129: We can use Nimbus for JWT operations. Platops confirmed.

src/main/java/uk/gov/hmcts/darts/common/util/RequestMatcher.java
mestebanez: Done

src/main/java/uk/gov/hmcts/darts/authentication/config/AuthenticationConfiguration.java
mestebanez: Done

hemantasharma1129 (approved):

PR #403, opened 2023-09-04
DMP-783 - Integrated dailylist processing with scheduling framework.

HMCTSDeveloper:
https://tools.hmcts.net/jira/browse/DMP-783


src/main/java/uk/gov/hmcts/darts/task/runner/impl/ProcessDailyListAutomatedTask.java
```diff
@@ -24,13 +29,17 @@ public String getTaskName() {
...
     }
 
     @Override
     protected void handleException(Exception exception) {
-        // TODO this is to be populated as a part of jira ticket https://tools.hmcts.net/jira/browse/DMP-183
         log.error("Exception: {}", exception.getMessage());
     }
 
+    @Autowired
+    public void setDailyListProcessor(DailyListProcessor dailyListProcessor) {
```
karen-hedges, line 41: Would this be better if it was passed in via the constructor and then it can be made final?

karen-hedges (approved):
Looks good to go just need a bit more code coverage

PR #406, opened 2023-09-04
DMP-459 - patch case

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-459

### DIFF
--- a/ModernisedDarts.postman_collection.json
+++ b/ModernisedDarts.postman_collection.json
@@ -243,6 +243,44 @@
 							"response": []
 						}
 					]
+				},
+				{
+					"name": "PATCH /cases/{case_id}",
+					"item": [
+						{
+							"name": "PATCH /cases/{case_id}",
+							"request": {
+								"method": "PATCH",
+								"header": [],
+								"body": {
+									"mode": "raw",
+									"raw": "{\n  \"retain_until\": \"2023-09-06T16:16:57.331Z\"\n}",
+									"options": {
+										"raw": {
+											"language": "json"
+										}
+									}
+								},
+								"url": {
+									"raw": "{{serverUrl}}/cases/:case_id",
+									"host": [
+										"{{serverUrl}}"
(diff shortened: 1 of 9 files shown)


ModernisedDarts.postman_collection.json
```diff
@@ -402,6 +440,43 @@
 							"response": []
 						}
 					]
+				},
```
davet1985, line 443: What's this bit? I think this should probably be removed anyway as we have `GET /cases`

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerPatchCaseTest.java
```diff
@@ -0,0 +1,64 @@
...
+        String actualResponse = response.getResponse().getContentAsString();
+
+        String expectedResponse = """
+            {"case_id":<<case_id>>,"courthouse":"testCourthouse","case_number":"testCaseNumber","retain_until":"2023-09-06T16:16:57.331Z"}""";
+        expectedResponse = expectedResponse.replace("<<case_id>>", createdCase.getId().toString());
+        assertEquals(expectedResponse, actualResponse, JSONCompareMode.NON_EXTENSIBLE);
+    }
+
+    @Test
+    void testFail() throws Exception {
```
davet1985, line 46: Can we also add a test to cover what happens if we try to patch more than just `retain_until`? I assume it will also fail?

davet1985 (commented):
A couple of questions

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerPatchCaseTest.java
cmagowan-hmcts: done

davet1985 (approved):

PR #410, opened 2023-09-05
DMP-364 Data Model - JPA Updates: Transcription Entity and Workflow

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-364

### Change description ###

Added new transcription tables and removed redundant columns

```
[ ] Yes
[X ] No
```


src/main/java/uk/gov/hmcts/darts/common/entity/TranscriptionEntity.java
```diff
@@ -75,18 +79,6 @@ public class TranscriptionEntity extends CreatedModifiedBaseEntity {
     @Column(name = "end_ts")
     private OffsetDateTime end;
 
-    @Column(name = "requested_by")
```
richardbull, line 78: I'm not entirely sure why we are removing these columns - I remember we added them to the LLD a few months ago when it was being thought that we would need a way to do Transcription searches for a particular user view e.g. My Transcriptions (for a requestor, approver, transcriber) was the idea.
Although it looked like Jack thought the `approved_by` field could be changed to be `actioned_by` perhaps which could be used where the approver chose `Rejected` action instead.

@jackmaloney should @barry-jones50410 have removed these columns with the introduction of the new transcription_workflow table, to remove workflow related elements from transcription? How do we find this stuff now

src/main/java/uk/gov/hmcts/darts/common/entity/TranscriptionStatusEntity.java
```diff
@@ -0,0 +1,35 @@
...
+import jakarta.persistence.Table;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+
+@Entity
+@Table(name = TranscriptionStatusEntity.TABLE_NAME)
+@Getter
+@Setter
+@EqualsAndHashCode(callSuper = false)
```
richardbull, line 18: I'm pretty sure that this can be removed as we haven't properly thought about implementing Equals / HashCode for our entities. I have a feeling it probably came about when I tried to introduce the `VersionedEntity` but the `version` column has since been dropped in the schema.

If you could remove it from all the Transcription* Entities please. We just want the Getter / Setter stuff for now.

FYI @cmagowan-hmcts / @hemantasharma1129

src/main/java/uk/gov/hmcts/darts/common/entity/TranscriptionStatusEntity.java
```diff
@@ -0,0 +1,35 @@
...
+public class TranscriptionStatusEntity {
+
+    public static final String TABLE_NAME = "transcription_status";
+    public static final String TRANSACTION_STATUS_ID = "trs_id";
```
richardbull, line 22: Typo on these globals - should be `TRANSCRIPTION_`

I'd prefer the new Transcription* entities to be in similar style to `TranscriptionEntity` - we don't need these columns exposed here if we are using jpa queries - I just find it harder to read the entity class.

This other style came about on the back of the Notification entity

src/main/java/uk/gov/hmcts/darts/common/entity/TranscriptionStatusEntity.java
```diff
@@ -0,0 +1,35 @@
...
+    public static final String TRANSACTION_STATUS_SEQUENCE = "trs_seq";
+    private static final String TRANSACTION_STATUS_TYPE = "status_type";
+
+    @Id
+    @Column(name = TRANSACTION_STATUS_ID)
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = TRANSACTION_STATUS_GENERATOR)
+    @SequenceGenerator(name = TRANSACTION_STATUS_GENERATOR, sequenceName = TRANSACTION_STATUS_SEQUENCE, allocationSize = 1)
+    private Integer id;
+
+    @Column(name = TRANSACTION_STATUS_TYPE)
```
richardbull, line 33: Missing `nullable = false`?

src/main/java/uk/gov/hmcts/darts/common/entity/TranscriptionWorkflowEntity.java
```diff
@@ -0,0 +1,48 @@
...
+
+@Entity
+@Table(name = TranscriptionWorkflowEntity.TABLE_NAME)
+@Getter
+@Setter
+@EqualsAndHashCode(callSuper = false)
+public class TranscriptionWorkflowEntity extends CreatedModifiedBaseEntity {
+
+    public static final String TABLE_NAME = "transcription_workflow";
+    public static final String TRANSACTION_WORKPLACE_ID = "trw_id";
```
richardbull, line 25: Similar typo and EqualsAndHashCode annotation comment as before

richardbull: Actually TRANSACTION -> TRANSCRIPTION
WORKPLACE -> WORKFLOW
...

src/main/java/uk/gov/hmcts/darts/common/entity/TranscriptionWorkflowEntity.java
```diff
@@ -0,0 +1,48 @@
...
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = TRANSACTION_WORKPLACE_GENERATOR)
+    @SequenceGenerator(name = TRANSACTION_WORKPLACE_GENERATOR, sequenceName = TRANSACTION_WORKPLACE_SEQUENCE, allocationSize = 1)
+    private Integer id;
+
+    @ManyToOne(optional = false)
+    @JoinColumn(name = TRANSACTION_ID, nullable = false)
+    private TranscriptionEntity transcription;
+
+    @Column(name = WORKFLOW_STAGE, nullable = false)
+    private String workflowStage;
```
richardbull, line 43: I wish we had a better understanding of what should be here - whether an enum and how that might be different from the Transcription Status which the 5 I chose were based originally on the heritage workflow lifecycle events anyway e.g. Requested, Awaiting Authorisation, With Transcriber, Complete, Rejected.

Wondering, do we want an enum here with the names you have similar to `TranscriptionStatusTypeEnum`?
@jackmaloney / @hemantasharma1129

src/main/resources/db/migration/common/V1_209__alter_transcription_table.sql
```diff
@@ -0,0 +1,17 @@
+--v46 Removed workflow related elements from transcription.
+--v47 Removed column transcription.current_state
+--    added column transcription.trs_id and FK to transcription_status
+
+ALTER TABLE transcription DROP COLUMN requested_by;
+ALTER TABLE transcription DROP COLUMN approved_by;
+ALTER TABLE transcription DROP COLUMN approved_on_ts;
+ALTER TABLE transcription DROP COLUMN transcribed_by;
```
richardbull, line 8: It's tricky - but if we're dropping these columns, we also need to drop the 3 constraints when comparing v43 with v47
* transcription_requested_by_fk
* transcription_approved_by_fk
* transcription_transcribed_by_fk

src/main/resources/db/migration/common/V1_210__insert_transcription_status_data.sql
```diff
@@ -0,0 +1,6 @@
+INSERT INTO darts.transcription_status VALUES(1, 'Requested');
+INSERT INTO darts.transcription_status VALUES(2, 'Awaiting Authorisation');
+INSERT INTO darts.transcription_status VALUES(3, 'With Transcriber');
+INSERT INTO darts.transcription_status VALUES(4, 'Complete');
+INSERT INTO darts.transcription_status VALUES(5, 'Rejected');
+INSERT INTO darts.transcription_status VALUES(6, 'Closed');
```
richardbull, line 6: I guess we need to restart the sequence `trs_seq` to be next value 7 - since we have one!
Even though I don't think we really need to have the sequence for a standing data table

src/main/resources/db/migration/common/V1_209__alter_transcription_table.sql
karen-hedges: If you drop a column with constraints the indexes and table constraints will automatically be dropped: https://www.postgresql.org/docs/7.4/sql-altertable.html#:~:text=DROP%20COLUMN,foreign%20key%20references%20or%20views.

src/main/java/uk/gov/hmcts/darts/common/entity/TranscriptionEntity.java
jackmaloney: My understanding is that now the workflow table will be used to do these lookups. For example, to work out who is assigned to a transcription request, we would need to look at the workflow table and check which user performed the 'with transcriber' (or whatever we call it) task under the workflow_stage field. 

Let me know if any concerns with this approach.

src/main/java/uk/gov/hmcts/darts/common/entity/TranscriptionWorkflowEntity.java
jackmaloney: Yes, that sounds like a plan @richardbull. Mind creating a ticket and making sure they align with what Barry is putting in? We also need 'Closed' as a stage for modernised.

src/main/resources/db/migration/common/V1_210__insert_transcription_status_data.sql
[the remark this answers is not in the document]
richardbull: In answer to my own question, I see the LLD State Model shows 

"Closed" - if a request has open for more than 30 days it will be closed by the [close transcription requests ](https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=1689786831#TaskautomationLLD-Closetranscriptionrequest(Previouslyknownas)task

@karen-hedges Please can you liase with DBA to add the extra status to their standing_data script?

src/main/java/uk/gov/hmcts/darts/common/entity/TranscriptionEntity.java
richardbull: @jackmaloney My main concern is the fields are missing unless we are thinking we should be able to use the jpa auditing fields for a table e.g. created_by / last_modified_by which we're not able to populate currently in our schema. I think this table sounds like we'd only ever be inserting into the workflow table.

It's a bit different now than e.g. media_request which would use the `requestor` field and what is new recently `current_owner`. We also have a conflicting field of same name under transcription table - but that must be a legacy throw away.

src/main/java/uk/gov/hmcts/darts/common/entity/TranscriptionWorkflowEntity.java
richardbull: @karen-hedges Please can you add an enum here to add the similar String enumerated type for this column?

    REQUESTED,
    AWAITING_AUTHORISATION,
    WITH_TRANSCRIBER,
    COMPLETE,
    REJECTED,
    CLOSED;


It would be similar to what we did in `MediaRequestEntity`
```
    @Column(name = ..., nullable = false)
    @Enumerated(EnumType.STRING)
    private WorkflowStageEnum workflowStage;
```

src/main/java/uk/gov/hmcts/darts/common/entity/TranscriptionEntity.java
jackmaloney: When I spoke to Barry yesterday about this, he advised the created_by field within the transcription_workflow table will be used for this purpose. 
The requestor column is used for legacy cases. Barry said "just for completeness, i ought to mention that the table currently also has a requestor column, which exists to support migration, if we subsequently determine that we cannot create an appropriate transcription_workflow record".

hemantasharma1129 (approved):

PR #411, opened 2023-09-05
DMP-172 - created event handler that only logs event information.

HMCTSDeveloper:
created event handler that only logs event information.

https://tools.hmcts.net/jira/browse/DMP-172


HMCTSDeveloper:
> Is there a change we need to make somewhere to link the event type to the handler?

I don't think so the jira is only mentions the possibility of doing it in the future.

tonymort (changes requested):
This won't work.  It will return an error on to api client because it can't find a mapping for the type and subtype.  You are getting a false positive in your test because you using mockito so it's not really an integration test.  The integration test should only mock external API calls not internal calls.  Its worth also using postman to verify behavoiur.

HMCTSDeveloper:
Looking at legacy there is 1 event for which the null handler is used. The event has already been added I will update the handler for it.

tonymort (approved):

PR #431, opened 2023-09-11
BUGFIX (DMP-829): Fix path usage for audio operations

chrisbellingham-hmcts:
These changes fix two feature-breaking bugs with the preview audio flow:

- Update `AudioServiceImpl.preview()` to call `AudioOperationService.reEncode()` with a directory name as the `workspaceDir` rather than a full absolute path. This directory is also explicitly created, as this is the expectation of the `ffmpeg` executable.
- Update `FileOperationService.saveFileToTempWorkspace()` to create a subdirectory within the temp workspace directory.

For both items above, the directory name is randomised on each call to ensure each request has a clean/empty location in which to write any output files. This avoids potential naming conflicts with existing files in the directory.

In genreal, method signatures are updated to favour use of `Path` rather than `String` to avoid future ambiguity around whether a filename or path is expected.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/darts/audio/service/impl/AudioOperationServiceImplTest.java
+++ b/src/test/java/uk/gov/hmcts/darts/audio/service/impl/AudioOperationServiceImplTest.java
@@ -29,7 +29,7 @@ import static org.mockito.Mockito.when;
 @ExtendWith(MockitoExtension.class)
 class AudioOperationServiceImplTest {
 
-    private static final String WORKSPACE_DIR = "requestId";
+    private static final String WORKSPACE_DIR = "44887a8c-d918-4907-b9e8-38d5b1bf9c9c";
     private static final String T_09_00_00_Z = "2023-04-28T09:00:00Z";
     private static final String T_10_30_00_Z = "2023-04-28T10:30:00Z";
     private static final String T_11_00_00_Z = "2023-04-28T11:00:00Z";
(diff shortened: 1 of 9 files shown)


hemantasharma1129 (approved):

src/test/java/uk/gov/hmcts/darts/audio/service/impl/AudioOperationServiceImplTest.java
```diff
@@ -45,113 +48,123 @@ class AudioOperationServiceImplTest {
...
+        when(audioConfigurationProperties.getConcatWorkspace()).thenReturn(tempDirectory.toString());
         when(systemCommandExecutor.execute(any())).thenReturn(Boolean.TRUE);
 
-        AudioFileInfo expectedAudio = new AudioFileInfo(
-            Instant.parse(T_09_00_00_Z),
-            Instant.parse(T_11_00_00_Z),
-            "/tempDir/concatenate/requestId/C1-concatenate-20230510145233697.mp2",
-            1
+        AudioFileInfo audioFileInfo = audioOperationService.concatenate(
+            WORKSPACE_DIR,
```
richardbull, line 109: Should this test example make use of something similar to `UUID.randomUUID().toString()` since we've chosen at this point to not make a subfolder with the `requestId`?

richardbull (approved):

src/test/java/uk/gov/hmcts/darts/audio/service/impl/AudioOperationServiceImplTest.java
chrisbellingham-hmcts: Good point, will update.

PR #433, opened 2023-09-12
DMP-810 add functional tests for cases resource

mmmalik1:

https://tools.hmcts.net/jira/browse/DMP-810

### Change description ###

Functional tests for Cases added

```
[ ] Yes
[x] No
```


tonymort (commented):
How does the data get cleaned up?  There was discussions about these test being idempotent.

mmmalik1:
> How does the data get cleaned up? There was discussions about these test being idempotent.

@tonymort this is a very good point as I do not have a an endpoint or event to remove case or events. This is one for @davet1985 to advise

hemantasharma1129 (approved):

PR #436, opened 2023-09-12
Proof of concept around problem  codes

mestebanez:

THIS IS A PROOF OF CONCEPT

### Change description ###

This idea will probably not go anywhere but thought i'd put the idea out there to get feedback. Its great that we have a common problem structure in out HTTP responses. The use of RFC 7807 is amazing.

To take this further we could bake enums error codes into the open api spec. This would allow clients to easily work out the nature of the responses without having to know the specific error integers. 

This PR ONLY converts the AuditApiError code.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/exception/AuditApiError.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/exception/AuditApiError.java
@@ -3,8 +3,8 @@ package uk.gov.hmcts.darts.common.exception;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import org.springframework.http.HttpStatus;
-import uk.gov.hmcts.darts.audio.model.AuditErrorCode;
-import uk.gov.hmcts.darts.audio.model.AuditTitleErrors;
+import uk.gov.hmcts.darts.audit.model.AuditErrorCode;
+import uk.gov.hmcts.darts.audit.model.AuditTitleErrors;
 
 import java.net.URI;
 
--- a/src/main/java/uk/gov/hmcts/darts/common/exception/AuditApiError.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/exception/AuditApiError.java
@@ -3,20 +3,21 @@ package uk.gov.hmcts.darts.common.exception;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import org.springframework.http.HttpStatus;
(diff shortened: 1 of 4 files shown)


chrisbellingham-hmcts (commented):
Hey @mestebanez thanks for the improvements!

I like the idea of using enums defined in the openapi spec to make the spec self-documenting. I'm wondering can we achieve that goal by defining only `AuditErrorCode` in the spec, along with a descriptive-enough `x-enum-varname`?

E.g. your definition tells the reader exactly which codes to expect, and the varnames give them an idea of what the problem is:

```yaml
AuditErrorCode:
      type: string
      enum:
        - "AUDIT_100"
        - "AUDIT_101"
      x-enum-varnames: [AUDIT_FILTERS_WERE_EMPTY, AUDIT_DATE_EMPTY]
```
Feels like we can then drop `AuditTitleErrors` from the spec, keeping the message as hardcoded String literals in `AuditApiError`, which feels a little less complex?

Key thing is we'd be communicating the error codes and their meaning (which are stable/unchanging) in the openapi spec as part of the API contract. Whereas changing the informative error message should not break the API contract, and so can remain part of the Java code.

What do you think?

mestebanez:
> Hey @mestebanez thanks for the improvements!
> 
> I like the idea of using enums defined in the openapi spec to make the spec self-documenting. I'm wondering can we achieve that goal by defining only `AuditErrorCode` in the spec, along with a descriptive-enough `x-enum-varname`?
> 
> E.g. your definition tells the reader exactly which codes to expect, and the varnames give them an idea of what the problem is:
> 
> ```yaml
> AuditErrorCode:
>       type: string
>       enum:
>         - "AUDIT_100"
>         - "AUDIT_101"
>       x-enum-varnames: [AUDIT_FILTERS_WERE_EMPTY, AUDIT_DATE_EMPTY]
> ```
> 
> Feels like we can then drop `AuditTitleErrors` from the spec, keeping the message as hardcoded String literals in `AuditApiError`, which feels a little less complex?
> 
> Key thing is we'd be communicating the error codes and their meaning (which are stable/unchanging) in the openapi spec as part of the API contract. Whereas changing the informative error message should not break the API contract, and so can remain part of the Java code.
> 
> What do you think?

Totally agree with you're comment, was thinking the same myself

mestebanez:
Closed this PR and added https://tools.hmcts.net/jira/browse/DMP-985 to the backlog

PR #440, opened 2023-09-13
DMP-962 - When the daillylist is created, the status of the Dailylist in the database is null which is not expected.

HMCTSDeveloper:
https://tools.hmcts.net/jira/browse/DMP-962

When an xml is only sent to the daily list post endpoint then we should set the status to NEW for that xml. When the processor handles the daily list it will fail and set the status to FAIL.

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListProcessorImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListProcessorImpl.java
@@ -77,7 +77,7 @@ public class DailyListProcessorImpl implements DailyListProcessor {
                 if (!dailyLists.isEmpty()) {
                     try {
                         processDailyList(dailyLists.get(0));
-                    } catch (JsonProcessingException e) {
+                    } catch (JsonProcessingException | IllegalArgumentException e) {
                         dailyLists.get(0).setStatus(String.valueOf(JobStatusType.FAILED));
                         log.error("Failed to process dailylist for courthouse: {} with dailylist id: {}",
                                   courthouse.getCourthouseName(), dailyLists.get(0).getId(), e
@@ -101,7 +101,7 @@ public class DailyListProcessorImpl implements DailyListProcessor {
     }
 
     @SuppressWarnings("checkstyle:VariableDeclarationUsageDistance")
-    private void processDailyList(DailyListEntity dailyListEntity) throws JsonProcessingException {
+    private void processDailyList(DailyListEntity dailyListEntity) throws JsonProcessingException, IllegalArgumentException {
         DailyListJsonObject dailyList = objectMapper.readValue(dailyListEntity.getContent(), DailyListJsonObject.class);
         JobStatusType statusType = JobStatusType.PROCESSED;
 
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListServiceImpl.java
```diff
@@ -90,6 +91,7 @@ public PostDailyListResponse saveDailyListXmlToDatabase(DailyListPostRequest pos
         dailyListEntity.setCourthouse(courthouse);
         dailyListEntity.setXmlContent(postRequest.getDailyListXml());
         dailyListEntity.setSource(postRequest.getSourceSystem());
+        dailyListEntity.setStatus(JobStatusType.NEW.name());
```
richardbull, line 94: The status field on the DailyListEntity feels odd here around the different ways we try to setStatus in code / test.  Ideally it would be changed to be something like the enum usage in `MediaRequestEntity` to make the code more readable and it looks like really we should be ensuring the column is `not null` as well?

```
    @Column(name = REQUEST_STATUS_COLUMN_NAME, nullable = false)
    @Enumerated(EnumType.STRING)
    private AudioRequestStatus status;
```

src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListProcessorImpl.java
```diff
@@ -136,7 +136,7 @@ private void processDailyList(DailyListEntity dailyListEntity) throws JsonProces
             } else {
                 statusType = JobStatusType.PARTIALLY_PROCESSED;
                 log.error("Unregistered courthouse " + courtHouseName + " daily list entry with id "
-                          + dailyListEntity.getId() + " has not been processed");
+                              + dailyListEntity.getId() + " has not been processed");
```
mestebanez, line 139: Incorrect indentation

src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListServiceImpl.java
richardbull: @HMCTSDeveloper Looks like @barry-jones50410 has made the database change recently in v48 TABLE daily_list.
`job_status                  CHARACTER VARYING             NOT NULL`
Do you think you could add the corresponding flyway script - may need to ask the test team if the dodgy records can be deleted from staging?

src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListProcessorImpl.java
```diff
@@ -150,7 +150,7 @@ private LocalTime getScheduledStartTime(Sitting sitting, Hearing dailyListHearin
                 return getTimeFromTimeMarkingNote(timeMarkingNoteText);
             } catch (DateTimeException dateTimeException) {
                 log.warn("Ignore error and continue, Parsing failed for field TimeMarkingNote with value: "
-                         + timeMarkingNoteText, dateTimeException);
+                             + timeMarkingNoteText, dateTimeException);
```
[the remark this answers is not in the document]
HMCTSDeveloper: Reverted Change.

src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListServiceImpl.java
HMCTSDeveloper: ACCEPT, will create separate jira for Barry's changes.

richardbull (approved):

PR #453, opened 2023-09-15
DMP-991 Transcription request: Error submitting Hearing ID with no Ca…

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-991

### Change description ###

Fixed issue whereby a transcription request is passed with a valid hearing id but no case id

```
[ ] Yes
[ X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/TranscriptionServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/TranscriptionServiceImpl.java
@@ -5,6 +5,7 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
 import uk.gov.hmcts.darts.cases.service.CaseService;
+import uk.gov.hmcts.darts.common.entity.HearingEntity;
 import uk.gov.hmcts.darts.common.entity.TranscriptionEntity;
 import uk.gov.hmcts.darts.common.entity.TranscriptionStatusEntity;
 import uk.gov.hmcts.darts.common.entity.TranscriptionTypeEntity;
@@ -43,6 +44,7 @@ public class TranscriptionServiceImpl implements TranscriptionService {
 
     private final CaseService caseService;
     private final HearingsService hearingsService;
+    private HearingEntity hearing;
 
 
     @Transactional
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/TranscriptionServiceImpl.java
```diff
@@ -75,12 +77,17 @@ private TranscriptionEntity saveTranscription(UserAccountEntity userAccount,
         }
 
         TranscriptionEntity transcription = new TranscriptionEntity();
-
+        HearingEntity hearing = null;
```
richardbull, line 80: This doesn't look right having this property at top level - please remove

src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/TranscriptionServiceImpl.java
```diff
@@ -75,12 +75,17 @@ private TranscriptionEntity saveTranscription(UserAccountEntity userAccount,
...
         if (!isNull(transcriptionRequestDetails.getHearingId())) {
```
richardbull, line 79: There are other methods in the Objects class like `nonNull` that could be used instead of `!isNull`, but I wonder why people use these methods outside of Predicates.  The JavaDoc says `This method exists to be used as a java.util.function.Predicate, filter(Objects::isNull)` and similarly for `nonNull`.
Prefer != null myself.

src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/TranscriptionServiceImpl.java
```diff
@@ -75,12 +75,17 @@ private TranscriptionEntity saveTranscription(UserAccountEntity userAccount,
...
-            transcription.setHearing(hearingsService.getHearingById(transcriptionRequestDetails.getHearingId()));
+            hearing = hearingsService.getHearingById(transcriptionRequestDetails.getHearingId());
```
richardbull, line 80: Really, we just need to getReferenceById to set the FK link - no need to get the full object.
But understand why we're doing that for convenience with the added API errors we have added in those services.

richardbull (approved):

src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/TranscriptionServiceImpl.java
karen-hedges: In my opinion it looks neater.

PR #455, opened 2023-09-15
DMP-933 (Dev Only) Populate Authorisation tables

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-933

### Change description ###

Updated the security roles and permissions.

N.B. THIS IS EXPECTED TO BREAK THE FLYWAY MIGRATIONS WHICH WILL NEED A MANUAL FIX

```
[ X] Yes
[ ] No
```


src/main/resources/db/migration/common/V1_214__update_security_roles_data.sql
```diff
@@ -0,0 +1,22 @@
...
+UPDATE security_role
+set role_name='JUDGE'
+WHERE rol_id = 3;
+
+UPDATE security_role
+set role_name='TRANSCRIBER'
+WHERE rol_id = 4;
+
+UPDATE security_role
+set role_name='LANGUAGE SHOP USER'
```
richardbull, line 18: Should these last 2 roles have screaming snake case names like you have set the permissions?

hemantasharma1129 (approved):

src/main/resources/db/migration/h2/V1_215_1__insert_security_group_reference_data.sql
richardbull, line 1: Will this cause a problem because the old script in common `src/main/resources/db/migration/common/V1_105__insert_security_group_reference_data.sql` no longer exists in the flyway history?

I'm not confident about this change - maybe @hemantasharma1129 might know

karen-hedges: It will break the build but we just need to remove the migration from the flyway table

richardbull: Is it possible to do it so that the scripts are a bit easier to understand - knowing that the DBAs probably need to have a simpler standing data / security_test_data sql themselves. Just everybody would have to deal with the flyway issue.

Can we delete the old security data I had previously?

karen-hedges: We have to do that as well but essentially the original script that populates the groups should not go into live. So this is why we have a manual process to remove and cleanup

PR #465, opened 2023-09-19
DMP-982 Update AddRequest Endpoint Response

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-982

### Change description ###

Added response object

```
[ ] Yes
[X ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/audio/component/impl/AudioResponseMapperImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/component/impl/AudioResponseMapperImpl.java
@@ -2,6 +2,8 @@ package uk.gov.hmcts.darts.audio.component.impl;
 
 import org.springframework.stereotype.Component;
 import uk.gov.hmcts.darts.audio.component.AudioResponseMapper;
+import uk.gov.hmcts.darts.audio.entity.MediaRequestEntity;
+import uk.gov.hmcts.darts.audio.model.AddAudioResponse;
 import uk.gov.hmcts.darts.audio.model.AudioMetadata;
 import uk.gov.hmcts.darts.common.entity.MediaEntity;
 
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/hmcts/darts/audio/component/impl/AudioResponseMapperImpl.java
```diff
@@ -10,6 +12,19 @@
...
 
+    @Override
+    public AddAudioResponse mapToAddAudioResponse(MediaRequestEntity audioRequest) {
+        AddAudioResponse addAudioResponse = new AddAudioResponse();
+        addAudioResponse.setRequestId(audioRequest.getId());
+        addAudioResponse.setCaseId(audioRequest.getHearing().getCourtCase().getCaseNumber());
+        addAudioResponse.setCourthouseName(audioRequest.getHearing().getCourtroom().getCourthouse().getCourthouseName());
+        addAudioResponse.setDefendants(audioRequest.getHearing().getCourtCase().getDefendantStringList());
+        addAudioResponse.setHearingDate(audioRequest.getHearing().getHearingDate());
+        addAudioResponse.setStartTime(audioRequest.getStartTime().toLocalTime().toString());
```
richardbull, line 23: Simply going toLocalTime doesn't feel right to me - I think it would be better to return the full offset string date/time.

In general, I think the team needs to question any use of LocalDateTime / LocalTime in the code as the server system clock will be in UTC.
We deal with UTC timestamps in darts-api and currently you can expect to always get a UTC date stored in the database and returned in the darts-api response.
If we wish get a "local" UK time, then I think it would be less confusing for us to use ZonedDateTime like we have done in some places where we needed the local time in darts-api for some metadata files as part of the audio download.

src/main/resources/openapi/audio.yaml
```diff
@@ -17,6 +17,12 @@ paths:
       operationId: addAudioRequest
       description: Adds a user audio request for processing
       responses:
+        200:
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AddAudioResponse'
         '201':
```
richardbull, line 26: The 201 response should be removed if we want to just have the 200 response now?

hemantasharma1129 (approved):

PR #466, opened 2023-09-19
DMP-932 AudioController / AudioRequestsController @Authorisation

richardbull:

[DMP-932](https://tools.hmcts.net/jira/browse/DMP-932)

### Change description ###

Added custom Authorisation annotation to Audio / Audio Requests endpoints
* GET /audio/download
* GET /audio/hearings/{hearing_id}/audios
* GET /audio/preview/{media_id}
* DELETE /audio-requests/{media_request_id}

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/authorisation/configuration/AuthorisationAspect.java
```diff
@@ -0,0 +1,259 @@
...
+
+        ContextIdEnum contextId = authorisationAnnotation.contextId();
+
+        switch (contextId) {
+            case CASE_ID -> checkAuthorisationByCaseId(request, roles);
+            case HEARING_ID -> checkAuthorisationByHearingId(request, roles);
+            case MEDIA_REQUEST_ID -> checkAuthorisationByMediaRequestId(request, roles);
+            case MEDIA_ID -> checkAuthorisationByMediaId(request, roles);
+            case TRANSCRIPTION_ID -> checkAuthorisationByTranscriptionId(request, roles);
+            default -> log.warn("Unrecognised contextId");
```
karen-hedges, line 93: Should this throw an exception when the context is unknown?

src/main/java/uk/gov/hmcts/darts/authorisation/configuration/AuthorisationAspect.java
```diff
@@ -0,0 +1,259 @@
...
+
+        if (caseIdParamOptional.isEmpty()) {
+            caseIdParamOptional = Optional.ofNullable(request.getParameter(CASE_ID_PARAM));
+            checkAuthorisationByCaseId(caseIdParamOptional, roles);
+        }
+
+        if (caseIdParamOptional.isEmpty()) {
+            caseIdParamOptional = Optional.ofNullable(request.getHeader(CASE_ID_PARAM));
+            checkAuthorisationByCaseId(caseIdParamOptional, roles);
+        }
```
karen-hedges, line 114: If the action should be against case id but there is no case id, should this also fail authorisation?

src/main/java/uk/gov/hmcts/darts/authorisation/component/impl/AuthorisationImpl.java
```diff
@@ -0,0 +1,99 @@
...
+import static uk.gov.hmcts.darts.audio.exception.AudioApiError.MEDIA_NOT_FOUND;
+import static uk.gov.hmcts.darts.audio.exception.AudioRequestsApiError.MEDIA_REQUEST_NOT_FOUND;
+import static uk.gov.hmcts.darts.cases.exception.CaseApiError.CASE_NOT_FOUND;
+import static uk.gov.hmcts.darts.hearings.exception.HearingApiError.HEARING_NOT_FOUND;
+import static uk.gov.hmcts.darts.transcriptions.exception.TranscriptionApiError.TRANSCRIPTION_NOT_FOUND;
+
+@Component
+@RequiredArgsConstructor
+@Slf4j
+public class AuthorisationImpl implements Authorisation {
```
mestebanez, line 30: Just an idea, no need to action it. 

This implementation will end up becoming a 'god' type mediator class. We could have an authorisation interface that each service implements? This way we decentralize this implementation, ensuring that authorisation for a specific resource is the responsibility of a specific service class.

src/main/java/uk/gov/hmcts/darts/authorisation/configuration/AuthorisationAspect.java
richardbull: It should never happen, I only added the default line because of checkstyle `switch without "default" clause. [MissingSwitchDefault]`. Done in commit ab8c4036.

src/main/java/uk/gov/hmcts/darts/authorisation/configuration/AuthorisationAspect.java
richardbull: Thanks, I guess it probably should on reflection - it was a very loose integration. Will correct.

src/main/java/uk/gov/hmcts/darts/authorisation/component/impl/AuthorisationImpl.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Override
+    public void authoriseByHearingId(Integer hearingId, Set<SecurityRoleEnum> securityRoles) {
+        try {
+            final List<CourthouseEntity> courthouses = List.of(hearingRepository.getReferenceById(hearingId)
+                                                                   .getCourtroom().getCourthouse());
+            authorisationApi.checkAuthorisation(courthouses, securityRoles);
+        } catch (EntityNotFoundException e) {
+            log.error("Unable to find Hearing-Courtroom-Courthouse for checkAuthorisation", e);
+            throw new DartsApiException(HEARING_NOT_FOUND);
```
mestebanez, line 59: Probably not,  but are we leaking information here? I'm thinking some kind of enumeration attack.  Should we be informing the client that the record in question does not exist?

src/main/java/uk/gov/hmcts/darts/authorisation/exception/AuthorisationError.java
```diff
@@ -0,0 +1,54 @@
...
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import org.springframework.http.HttpStatus;
+import uk.gov.hmcts.darts.common.exception.DartsApiError;
+
+@Getter
+@RequiredArgsConstructor
+public enum AuthorisationError implements DartsApiError {
+
+    USER_NOT_AUTHORISED_FOR_COURTHOUSE(
```
mestebanez, line 12: Again could be split as per my comments above.

src/main/java/uk/gov/hmcts/darts/authorisation/config/AuthorisationAspect.java
```diff
@@ -0,0 +1,262 @@
...
+
+        if (hearingIdParamOptional.isEmpty()) {
+            hearingIdParamOptional = Optional.ofNullable(request.getParameter(HEARING_ID_PARAM));
+            checkAuthorisationByHearingId(hearingIdParamOptional, roles);
+        }
+
+        if (hearingIdParamOptional.isEmpty()) {
+            hearingIdParamOptional = Optional.ofNullable(request.getHeader(HEARING_ID_PARAM));
+            checkAuthorisationByHearingId(hearingIdParamOptional, roles);
+        }
```
karen-hedges, line 149: I guess the same comment applies about if your authentication is against hearing id, media request id etc and its null, should it fail authentication. Not sure how you handle the requests that need to create the hearing, case, media etc.

src/main/java/uk/gov/hmcts/darts/authorisation/config/AuthorisationAspect.java
```diff
@@ -0,0 +1,259 @@
...
+            case MEDIA_ID -> checkAuthorisationByMediaId(request, roles);
+            case TRANSCRIPTION_ID -> checkAuthorisationByTranscriptionId(request, roles);
+            default -> log.warn("Unrecognised contextId");
+        }
+    }
+
+    private void checkAuthorisationByCaseId(HttpServletRequest request, Set<SecurityRoleEnum> roles) {
+        Optional<String> caseIdParamOptional = Optional.empty();
+
+        Matcher matcher = CASES_ID_PATH_PATTERN.matcher(request.getRequestURI());
```
mestebanez, line 103: Rather than binding ourselves to the URL pattern can we not just check the input parameters of the method? Spring has already extrapolated these for us.

src/main/java/uk/gov/hmcts/darts/authorisation/configuration/AuthorisationAspect.java
richardbull: Done in commit 5af180b8

src/main/java/uk/gov/hmcts/darts/authorisation/config/AuthorisationAspect.java
richardbull: Done in commit https://github.com/hmcts/darts-api/commit/5af180b88f1eb52400920bfeba84714892a24238

src/main/java/uk/gov/hmcts/darts/authorisation/config/AuthorisationAspect.java
richardbull: Thanks @mestebanez - done in commit 023357fd. I missed that, now simplified and clears my headache around the regex matching of our audio endpoints such as "preview". Cheers

src/main/java/uk/gov/hmcts/darts/authorisation/component/impl/AuthorisationImpl.java
richardbull: I had a similar thought myself - it was done here to keep the similar responses on the Cases / Hearing API which responds with these Not Found errors. So I did similar for the other 3 contexts that I couldn't navigate to courthouses from. I wanted to distinguish between the Bad Request / Not Found / Unauthorized client errors to avoid confusion.
@hemantasharma1129 advises to keep it for now.

src/main/java/uk/gov/hmcts/darts/authorisation/exception/AuthorisationError.java
richardbull: @hemantasharma1129 advises to keep it for now. He wanted the same Unauthorised response for all Controller operations with the custom 'Authorisation' annotation across 5 contexts with a contextId and securityRoles parameters.

src/main/java/uk/gov/hmcts/darts/authorisation/component/impl/AuthorisationImpl.java
richardbull: @hemantasharma1129 advises to keep it for now. It was mainly done, so that I could standup a simple test schema to develop against in our integration tests using `AuthorisationStub`. Not all the endpoints can be protected by the custom annotation currently as they may not be acting on a single context ID parameter e.g. POST transcription and GET cases. So they will have to be done on the data layer [DMP-626](https://tools.hmcts.net/jira/browse/DMP-626).

hemantasharma1129 (approved):

PR #505, opened 2023-10-02
dmp-1090 | feat | Keda work

LyleCGI:
# [DMP-1090](https://tools.hmcts.net/jira/browse/DMP-1090)

Allow `processAudioRequest()` to be invoked by KEDA for the oldest `OPEN` media_request upon application startup when the `ATS_MODE` env var is assigned.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/Application.java
+++ b/src/main/java/uk/gov/hmcts/darts/Application.java
@@ -1,16 +1,15 @@
 package uk.gov.hmcts.darts;
 
 import jakarta.annotation.PostConstruct;
+import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.CommandLineRunner;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.autoconfigure.security.servlet.SecurityAutoConfiguration;
 import org.springframework.boot.autoconfigure.security.servlet.UserDetailsServiceAutoConfiguration;
 import org.springframework.transaction.annotation.EnableTransactionManagement;
-import uk.gov.hmcts.darts.audio.enums.AudioRequestStatus;
-import uk.gov.hmcts.darts.audio.service.MediaRequestService;
+import uk.gov.hmcts.darts.audio.api.AudioInternalApi;
 
 import java.util.TimeZone;
 
@@ -19,11 +18,10 @@ import static java.time.ZoneOffset.UTC;
 @SpringBootApplication(exclude = {SecurityAutoConfiguration.class, UserDetailsServiceAutoConfiguration.class})
 @EnableTransactionManagement
 @Slf4j
-@SuppressWarnings("HideUtilityClassConstructor") // Spring needs a constructor, its not a utility class
+@RequiredArgsConstructor
 public class Application implements CommandLineRunner {
 
-    @Autowired
-    private MediaRequestService mediaRequestService;
+    private final AudioInternalApi audioInternalApi;
 
     @PostConstruct
     public void started() {
@@ -42,12 +40,10 @@ public class Application implements CommandLineRunner {
     }
 
     @Override
-    public void run(String... args) throws Exception {
+    public void run(String... args) {
         if (System.getenv("ATS_MODE") != null) {
             log.info("ATS_MODE activated");
-            // Temporary workaround to prevent many media requests in OPEN state
-            var openAudioRequests = mediaRequestService.getMediaRequestsByStatus(AudioRequestStatus.OPEN);
-            mediaRequestService.updateAudioRequestStatus(openAudioRequests.get(0).getId(), AudioRequestStatus.PROCESSING);
+            audioInternalApi.handleKedaInvocationForMediaRequests();
         }
     }
 
--- a/src/main/java/uk/gov/hmcts/darts/audio/service/MediaRequestService.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/service/MediaRequestService.java
@@ -8,10 +8,9 @@ import uk.gov.hmcts.darts.audiorequests.model.AudioRequestDetails;
 
 import java.io.InputStream;
 import java.util.List;
+import java.util.Optional;
 
 public interface MediaRequestService {
-
-    List<MediaRequestEntity> getMediaRequestsByStatus(AudioRequestStatus status);
 
     MediaRequestEntity getMediaRequestById(Integer id);
 
@@ -24,6 +23,8 @@ public interface MediaRequestService {
     void deleteAudioRequest(Integer mediaRequestId);
 
     List<AudioRequestSummaryResult> viewAudioRequests(Integer userId, Boolean expired);
+
+    Optional<MediaRequestEntity> getOldestMediaRequestByStatus(AudioRequestStatus status);
 
     void updateAudioRequestLastAccessedTimestamp(Integer mediaRequestId);
 
--- a/src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
@@ -9,6 +9,7 @@ import org.springframework.stereotype.Service;
 import uk.gov.hmcts.darts.audio.component.OutboundFileProcessor;
 import uk.gov.hmcts.darts.audio.component.OutboundFileZipGenerator;
 import uk.gov.hmcts.darts.audio.entity.MediaRequestEntity;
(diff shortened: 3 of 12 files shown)


src/main/java/uk/gov/hmcts/darts/Application.java
```diff
@@ -45,9 +48,7 @@ public static void main(final String[] args) {
     public void run(String... args) throws Exception {
         if (System.getenv("ATS_MODE") != null) {
             log.info("ATS_MODE activated");
-            // Temporary workaround to prevent many media requests in OPEN state
-            var openAudioRequests = mediaRequestService.getMediaRequestsByStatus(AudioRequestStatus.OPEN);
-            mediaRequestService.updateAudioRequestStatus(openAudioRequests.get(0).getId(), AudioRequestStatus.PROCESSING);
+            audioTransformationService.handleKedaInvocationForMediaRequests();
```
davet1985, line 51: Should we expose this method out via an interface in the `audio` package, instead of using the service directly? CC @chrisbellingham-hmcts

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
```diff
@@ -314,4 +315,12 @@ private void notifyUser(MediaRequestEntity mediaRequestEntity,
         log.debug("Notification scheduled successfully");
     }
 
+    @Override
+    public void handleKedaInvocationForMediaRequests() {
+        Optional<MediaRequestEntity> openAudioRequests = mediaRequestService.getOldestMediaRequestByStatus(AudioRequestStatus.OPEN);
+        if (openAudioRequests.isPresent()) {
+            processAudioRequest(openAudioRequests.get().getId());
```
hemantasharma1129, line 322: Before calling processAudioRequest, we will have to update the media_request table record with PROCESSING status.

src/main/java/uk/gov/hmcts/darts/Application.java
hemantasharma1129: We already have AudioTransformationService Interface which the logic should be using. Since the requirement is to have an asynchronous process which will be triggered by KEDA, so I don't see a reason to expose it as Api. Unless there is a requirement to trigger the process forcefully from portal.

davet1985: I didn't mean exposing it as an API, just an interface into the service. As was proposed very early on in the development of the app. In order to make clear the cross-package calls that are happening.

chrisbellingham-hmcts: My interpretation of @davet1985's comment is that we should follow the established pattern whereby any external usages of a service method should be defined in an interface in the `audio.api` package. This will not be an http api, simply a java interface with an implementation that delegates to the service.

I think this is a good point, and I agree we should follow that pattern here.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
chrisbellingham-hmcts: The existing `processAudioRequest()` method already sets status to PROCESSING, so I don't think we need to do this here.

Snippet:
```java
    public UUID processAudioRequest(Integer requestId) {

        log.debug("Starting processing for audio request id: {}", requestId);
        mediaRequestService.updateAudioRequestStatus(requestId, PROCESSING);
        ...etc
}
```

src/main/java/uk/gov/hmcts/darts/Application.java
hemantasharma1129: I see. I misinterpreted as exposing it as an http api. Yes we can follow the same pattern then.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/MediaRequestServiceImpl.java
```diff
@@ -56,6 +58,12 @@ public List<MediaRequestEntity> getMediaRequestsByStatus(AudioRequestStatus stat
         return mediaRequestRepository.findByStatusOrderByCreatedDateTimeAsc(status);
     }
 
+    @Override
+    @Transactional(propagation = Propagation.SUPPORTS)
+    public Optional<MediaRequestEntity> getOldestMediaRequestByStatus(AudioRequestStatus status) {
```
chrisbellingham-hmcts, line 63: I think this is a good change, but consider removing the `getMediaRequestsByStatus()` set of methods as these are now redundant with this change.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
chrisbellingham-hmcts, line 322: Since we get an Optional from `getOldestMediaRequestByStatus()`, a functional approach is a little cleaner here, e.g:

```java
        mediaRequestService.getOldestMediaRequestByStatus(AudioRequestStatus.OPEN)
            .ifPresent(openMediaRequests -> processAudioRequest(openMediaRequests.getId()));
```

src/main/java/uk/gov/hmcts/darts/audio/service/MediaRequestService.java
```diff
@@ -22,4 +25,7 @@ public interface MediaRequestService {
 
     List<AudioRequestSummaryResult> viewAudioRequests(Integer userId, Boolean expired);
 
+    Optional<MediaRequestEntity> getOldestMediaRequestByStatus(AudioRequestStatus status);
```
chrisbellingham-hmcts, line 28: Good descriptive method name.

src/integrationTest/java/uk/gov/hmcts/darts/audio/service/AudioTransformationServiceProcessAudioRequestTest.java
chrisbellingham-hmcts, line 1: Please add a test case covering AC2 in https://tools.hmcts.net/jira/browse/DMP-1090:

`
AC2:

GIVEN

KEDA is running

WHEN

No record exist in Media Request table

THEN

KEDA should not trigger the ATS process.
`

src/integrationTest/java/uk/gov/hmcts/darts/audio/service/AudioTransformationServiceProcessAudioRequestTest.java
```diff
@@ -52,8 +53,9 @@ void setUp() {
 
     @ParameterizedTest
     @EnumSource(names = {"DOWNLOAD", "PLAYBACK"})
+    @Order(1)
```
chrisbellingham-hmcts, line 56: Is ordering the tests necessary? It's preferable not to order, as it suggest there is an ordering dependency between the tests, and ideally each test should be independent/standalone.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
```diff
@@ -90,16 +89,88 @@ void postConstruct() {
...
+    @Transactional
+    public UUID saveProcessedData(MediaRequestEntity mediaRequest, BinaryData binaryData) {
+        UUID blobId = saveAudioBlobData(binaryData);
+        saveTransientDataLocation(mediaRequest, blobId);
+
+        return blobId;
+    }
+
+    @Override
+    public void handleKedaInvocationForMediaRequests() {
```
richardbull, line 160: This top level service looks like it's missing `@Transactional` for all the various saves nested in `processAudioRequest`?

hemantasharma1129 (approved):

src/integrationTest/java/uk/gov/hmcts/darts/audio/service/AudioTransformationServicehandleKedaInvocationForMediaRequestsGivenBuilder.java
```diff
@@ -28,7 +28,7 @@
 @Setter
 @Scope(scopeName = SCOPE_PROTOTYPE)
 @SuppressWarnings("MethodName")
-public class AudioTransformationServiceProcessAudioRequestGivenBuilder {
+public class AudioTransformationServicehandleKedaInvocationForMediaRequestsGivenBuilder {
```
richardbull, line 31: I guess these 2 files in the change list with `Servicehandle` should be renamed to be `ServiceHandle`...?

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
chrisbellingham-hmcts: I think we explictly want to avoid `@Transactional` for this particular case - if there is an issue/exception thrown during processing we don't want to role-back any changes to `media_request.request_status`. The approach in `processAudioRequest()` is to catch any exception and explictly set state to FAILED.

PR #533, opened 2023-10-06
DMP-154 - Transcription Request handler

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-154


src/main/resources/db/migration/common/V1_221__TranscriptionRequestHandler.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE darts.event_handler SET handler = 'TranscriptionRequestHandler', last_modified_ts = current_timestamp, last_modified_by = 0 where evh_id = 78;
```
richardbull, line 1: Please would you be able to update the standing data for event_handler and user_account? We probably need to add the other system users for Barry's awareness.
[modernised_darts_standing_data.sql](https://github.com/hmcts/darts-api/blob/master/src/main/resources/db/reference/modernised_darts_standing_data.sql#L88C83-L88C83)

richardbull (approved):

PR #536, opened 2023-10-09
DMP-1090: Resolve LazyInitialisationException for processAudioRequest

chrisbellingham-hmcts:
This fix resolves `LazyInitialisationException` thrown at multiple points through the `processAudioRequest` flow.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/audio/entity/MediaRequestEntity.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/entity/MediaRequestEntity.java
@@ -57,7 +57,7 @@ public class MediaRequestEntity extends CreatedModifiedBaseEntity {
     @JoinColumn(name = HEARING_ID_COLUMN_NAME, nullable = false)
     private HearingEntity hearing;
 
-    @ManyToOne(fetch = FetchType.EAGER, optional = false)
+    @ManyToOne(optional = false)
     @JoinColumn(name = REQUESTOR_COLUMN_NAME, nullable = false)
     private UserAccountEntity requestor;
 
--- a/src/main/java/uk/gov/hmcts/darts/audio/entity/MediaRequestEntity.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/entity/MediaRequestEntity.java
@@ -57,7 +57,7 @@ public class MediaRequestEntity extends CreatedModifiedBaseEntity {
     @JoinColumn(name = HEARING_ID_COLUMN_NAME, nullable = false)
     private HearingEntity hearing;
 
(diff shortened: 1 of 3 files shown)


karen-hedges (approved):

hemantasharma1129 (approved):

src/main/java/uk/gov/hmcts/darts/audio/entity/MediaRequestEntity.java
```diff
@@ -57,7 +57,7 @@ public class MediaRequestEntity extends CreatedModifiedBaseEntity {
     @JoinColumn(name = HEARING_ID_COLUMN_NAME, nullable = false)
     private HearingEntity hearing;
 
-    @ManyToOne(fetch = FetchType.LAZY, optional = false)
+    @ManyToOne(fetch = FetchType.EAGER, optional = false)
```
richardbull, line 60: This fetch attribute can be omitted since EAGER is the default for `ManyToOne`

src/main/java/uk/gov/hmcts/darts/common/entity/HearingEntity.java
```diff
@@ -78,7 +78,7 @@ public class HearingEntity extends CreatedModifiedBaseEntity {
     @Transient
     private boolean isNew; //helper flag to indicate that the entity was just created, and so to notify DAR PC
 
-    @ManyToMany
+    @ManyToMany(fetch = FetchType.EAGER)
```
richardbull, line 81: I think there's something not right with the modelling of this ManyToMany relationship.
It looks strange to me because one side should be the owning parent side, so you would expect to have the JoinTable annotations like we have here, but if you navigate to the other side. It has similar, whereas I'd expect to see `mappedBy` attribute of ManyToMany being used by the other side of the bidirectional relationship.

chrisbellingham-hmcts: It does look a bit odd - do you think that could be covered under a separate tech debt ticket so as to not impede DMP-1090?

cmagowan-hmcts: I'd be worried about performance here. Looks like this would affect the Advanced Search etc, which is not necessarily interested in events?

chrisbellingham-hmcts: As dicussed let's handle this as a separate task.

chrisbellingham-hmcts: This change is reverted per https://github.com/hmcts/darts-api/pull/543.

PR #544, opened 2023-10-09
increase advanced search performance

cmagowan-hmcts:
increase search performance by modifying the main search query to only bring back a list of caseIds, instead of a list of CourtCaseEntities, as when we were bringing back the entities, hibernate was going and fetching judges, prosecutors etc, which was not required for the functionality.


davet1985 (commented):
Could you please add a bit to the description about what has changed to increase search performance.

cmagowan-hmcts:
> Could you please add a bit to the description about what has changed to increase search performance.

@davet1985 
done

PR #546, opened 2023-10-10
DMP-626 data layer auth

garethmorriscgi:
https://tools.hmcts.net/jira/browse/DMP-626 

### Change description ###
Added security group filter to advance search queries

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/common/entity/CourthouseEntity.java
```diff
@@ -37,4 +42,10 @@ public class CourthouseEntity extends CreatedModifiedBaseEntity {
 
     @OneToMany(mappedBy = "courthouse")
     private List<CourtroomEntity> courtrooms;
+
+    @ManyToMany
```
richardbull, line 46: If we need to define a bidirectional ManyToMany relationship here - then I think we should use mappedBy attribute using `SecurityGroupEntity_.COURTHOUSE_ENTITIES`
 So no need to specify the JoinTable on the child side.
e.g. `@ManyToMany(mappedBy = COURTHOUSE_ENTITIES)`

src/main/java/uk/gov/hmcts/darts/common/entity/SecurityGroupEntity.java
```diff
@@ -68,4 +68,10 @@ public class SecurityGroupEntity {
         inverseJoinColumns = {@JoinColumn(name = "cth_id")})
     private Set<CourthouseEntity> courthouseEntities = new LinkedHashSet<>();
 
+    @ManyToMany
```
richardbull, line 71: UserAccountEntity_.SECURITY_GROUP_ENTITIES
 `@ManyToMany(mappedBy = SECURITY_GROUP_ENTITIES)`
 So no need to have JoinTable annotation on child side.

src/main/java/uk/gov/hmcts/darts/cases/helper/AdvancedSearchRequestHelper.java
```diff
@@ -35,10 +41,13 @@
 
 @Component
 @SuppressWarnings({"PMD.TooManyMethods"})
+@RequiredArgsConstructor
 public class AdvancedSearchRequestHelper {
     @PersistenceContext
     private EntityManager entityManager;
```
richardbull, line 47: Should this be made `final` also for the RequiredArgsConstructor?

src/main/java/uk/gov/hmcts/darts/cases/helper/AdvancedSearchRequestHelper.java
```diff
@@ -134,6 +143,19 @@ private List<Predicate> addCourthouseCriteria(GetCasesSearchRequest request, Cri
         return predicateList;
     }
 
+    private List<Predicate> addPermissionsCriteria(GetCasesSearchRequest request, CriteriaBuilder criteriaBuilder, Root<CourtCaseEntity> caseRoot) {
+        String userAccount = userIdentity.getEmailAddress();
```
richardbull, line 147: Do we need this variable - strange name. Probably better to put the RHS down into line 153, since that's the only place that uses this emailAddress.

src/main/java/uk/gov/hmcts/darts/cases/helper/AdvancedSearchRequestHelper.java
```diff
@@ -190,8 +224,12 @@ private Join<HearingEntity, CourtroomEntity> joinCourtroom(Root<CourtCaseEntity>
 
     }
 
+    @SuppressWarnings("unchecked")
     private Join<CourtCaseEntity, CourthouseEntity> joinCourthouse(Root<CourtCaseEntity> caseRoot) {
-        return caseRoot.join(CourtroomEntity_.COURTHOUSE, JoinType.INNER);
+        Optional<Join<CourtCaseEntity, ?>> foundJoin = caseRoot.getJoins().stream().filter(join -> join.getAttribute().getName().equals(
```
richardbull, line 229: Why do we need to do all this - looking at Table "darts.court_case" - we would always expect to have this relationship?
`cth_id           | integer                  |           | not null `

src/main/java/uk/gov/hmcts/darts/cases/helper/AdvancedSearchRequestHelper.java
garethmorriscgi: Yep, updated entity manager to be final

src/main/java/uk/gov/hmcts/darts/common/entity/CourthouseEntity.java
garethmorriscgi: Applying above changes caused following errors: 
Caused by: org.hibernate.AnnotationException: Collection 'uk.gov.hmcts.darts.common.entity.CourthouseEntity.securityGroups' is 'mappedBy' a property named 'securityGroupEntities' which does not exist in the target entity 'uk.gov.hmcts.darts.common.entity.SecurityGroupEntity'

richardbull: It should be `SecurityGroupEntity_.COURTHOUSE_ENTITIES`
That generated file has the Java property `courthouseEntities` - so it should be ok.
Think you tried the wrong thing?

src/main/java/uk/gov/hmcts/darts/cases/helper/AdvancedSearchRequestHelper.java
```diff
@@ -138,6 +148,18 @@ private List<Predicate> addCourthouseCriteria(GetCasesSearchRequest request, Cri
         return predicateList;
     }
 
+    private List<Predicate> addUserSecurityRolesCriteria(GetCasesSearchRequest request, CriteriaBuilder criteriaBuilder, Root<CourtCaseEntity> caseRoot) {
```
richardbull, line 151: Parameter 'request' is never used

src/main/java/uk/gov/hmcts/darts/cases/helper/AdvancedSearchRequestHelper.java
```diff
@@ -138,6 +148,18 @@ private List<Predicate> addCourthouseCriteria(GetCasesSearchRequest request, Cri
...
+
+        List<Predicate> predicateList = new ArrayList<>();
+        Join<CourtCaseEntity, UserAccountEntity> userJoin = joinUser(caseRoot);
+        predicateList.add(criteriaBuilder.equal(
+                              userJoin.get(UserAccountEntity_.EMAIL_ADDRESS),
```
richardbull, line 156: Think this criteria will need to use equal lower both sides like e.g.
https://github.com/hmcts/darts-api/blob/a595b7cb11d92dfe7bdc64293d749c7dc359d0a8/src/main/java/uk/gov/hmcts/darts/authorisation/service/impl/AuthorisationServiceImpl.java#L84

src/main/java/uk/gov/hmcts/darts/admin/test/TestSupportController.java
```diff
@@ -75,12 +79,22 @@ public void cleanUpDataAfterFunctionalTests() {
...
+        removeUserCourthousePermissions(session, courthouseTrash);
+
         removeCourtHouses(session);
 
         session.getTransaction().commit();
         session.close();
     }
 
+    private void removeUserCourthousePermissions(Session session,List cthIds) {
+        session.createNativeQuery("""
```
richardbull, line 91: No need for a multi-line string here?
I'm also not feeling it is right to simply remove the links in table security_group_courthouse_ae, unless we can be certain that it belongs only to a TEST security_group / courthouse?
I can't easily see this to know currently

src/main/java/uk/gov/hmcts/darts/common/entity/CourthouseEntity.java
richardbull: Please try using `@ManyToMany(mappedBy = SecurityGroupEntity_.COURTHOUSE_ENTITIES)`

src/main/java/uk/gov/hmcts/darts/admin/test/TestSupportController.java
johnrace0: Don't actually need multi-line here - but it's consistent with the rest of the code in this class.

src/main/java/uk/gov/hmcts/darts/cases/helper/AdvancedSearchRequestHelper.java
johnrace0: Changed criteriaBuilder to lower case

PR #557, opened 2023-10-11
Running daily list housekeeping cron job every day at 4:30pm

tonymort:
Running daily list housekeeping cron job every day at 4:30pm

```
[ ] Yes
[x] No
```


src/main/resources/application.yaml
```diff
@@ -116,7 +116,7 @@ darts:
   daily-list:
     housekeeping:
       enabled: true
-      cron: ${DAILY_LIST_HOUSEKEEPING_SCHEDULER_CRON:4 4 2 * * SAT}
+      cron: '0 30 16 * * *'
```
davet1985, line 119: I think we should keep the `DAILY_LIST_HOUSEKEEPING_SCHEDULER_CRON` in case we want to override the value per environment

davet1985 (approved):

PR #590, opened 2023-10-18
DMP 1193: Redis cache config & bank holiday caching 

tonymort:
Added caching support with connections to Redis by default.  A simple in memory cache is also configured using the profile `in-memory-cache` if required for running the application without access to a Redis instance.  The Redis connection configuration required some low level configuration due to an issue with Netty, Lettuce's (Redis Client Lib) underlying http client with loading a functioning DNS resolver.  Basic cache usage has been implemented for retrieving and caching of bank holidays.  This may change as use case are implemented.  

A testing support endpoint has been added to verify the cache is working in staging. 

Redis has been added to the docker compose config for local testing.

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/service/bankholidays/BankHolidaysService.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/service/bankholidays/BankHolidaysService.java
@@ -1,0 +1,21 @@
+package uk.gov.hmcts.darts.common.service.bankholidays;
+
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+
+import java.util.List;
+
+import static java.util.stream.Collectors.toList;
+
+@Service
+@RequiredArgsConstructor
+public class BankHolidaysService {
+
+    private final BankHolidaysApi bankHolidaysApi;
+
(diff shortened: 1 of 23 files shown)


chrisdavecm (approved):

src/main/java/uk/gov/hmcts/darts/common/service/bankholidays/BankHolidaysService.java
```diff
@@ -0,0 +1,21 @@
...
+
+import static java.util.stream.Collectors.toList;
+
+@Service
+@RequiredArgsConstructor
+public class BankHolidaysService {
+
+    private final BankHolidaysApi bankHolidaysApi;
+
+    public List<Event> getBankHolidaysFor(int year) {
```
jackmaloney, line 16: What's the reason for passing a year in here? I don't think there are 100's of dates that would be stored, so not sure you're saving much by doing it this way. The requirement for DARTS is going to be along the lines of 'tell me if there is a bank holiday in the next X number of days', so not sure the logic implemented here will actually be used in real life, correct me if i've missed something here?

PR #610, opened 2023-10-25
DMP-666: Changing ttl representation from days to minutes

tonymort:
Changing ttl representation from days to minutes

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/config/BankHolidayCacheConfig.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/config/BankHolidayCacheConfig.java
@@ -23,12 +23,12 @@ import java.time.Duration;
 @Slf4j
 public class BankHolidayCacheConfig {
 
-    @Value("${darts.cache.bank-holiday.expiry-days}")
-    private int bankHolidayExpiry;
+    @Value("${darts.cache.bank-holiday.expiry-minutes}")
+    private int bankHolidayExpiryMinutes;
 
     @Bean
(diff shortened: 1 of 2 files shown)


davet1985 (approved):

src/main/java/uk/gov/hmcts/darts/common/config/BankHolidayCacheConfig.java
```diff
@@ -23,12 +23,12 @@
 @Slf4j
 public class BankHolidayCacheConfig {
 
-    @Value("${darts.cache.bank-holiday.expiry-days}")
-    private int bankHolidayExpiry;
+    @Value("${darts.cache.bank-holiday.expiry-minutes}")
+    private int bankHolidayExpiryMinutes;
```
chrisbellingham-hmcts, line 27: Consider using Spring's ability to inject Durations to better leverage the type system here.

E.g:
- Change property name to `darts.cache.bank-holiday.expiry`
- In the yaml specify: `darts.cache.bank-holiday.expiry: PT10080M`, which defines 10080 minutes
- Change the type on L27: `private Duration bankHolidayExpiry`
- Use that duration directly on L41 (i.e no need for the ofMinutes wrapper)

The advantage is the unit name doesn't need to be described in the property or variable name, and the config value can be expressed in different units without the code changing.

tonymort: Cheers, updated

PR #625, opened 2023-10-31
DMP-593 - added automated task to delete outbound audio.

HMCTSDeveloper:


src/main/java/uk/gov/hmcts/darts/task/service/impl/AutomatedTaskServiceImpl.java
```diff
@@ -254,6 +281,24 @@ private void rescheduleCloseNonCompletedTranscriptionsAutomatedTask() {
         }
     }
 
+    private void rescheduleOutboundAudioDeleterAutomatedTask() {
+        OutboundAudioDeleterAutomatedTask outboundAudioDeleterAutomatedTask;
+        Trigger trigger;
```
tonymort, line 286: These don't need to be defined outside there usage scope, I see its been copied from above. I would change those too.

src/main/java/uk/gov/hmcts/darts/audio/config/AudioConfiguration.java
```diff
@@ -1,10 +1,22 @@
...
 import org.springframework.context.annotation.Configuration;
 
+import java.time.Clock;
+
 @Configuration
 @EnableConfigurationProperties(AudioConfigurationProperties.class)
 public class AudioConfiguration {
 
+    @Value("${darts.audio.outbounddeleter.last-accessed-deletion-day:2}")
+    private long lastAccessedDeletionDays;
```
tonymort, line 15: Should this be used by something?

src/main/java/uk/gov/hmcts/darts/common/service/bankholidays/Event.java
```diff
@@ -2,9 +2,13 @@
...
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Getter;
+import lombok.Setter;
 
 import java.time.LocalDate;
 
+@Getter
+@Setter
 @JsonIgnoreProperties(ignoreUnknown = true)
```
tonymort, line 12: My mistake from previous ticket but as you added lombok can the fields be private?

src/integrationTest/java/uk/gov/hmcts/darts/task/service/OutboundAudioDeleterProcessorTest.java
```diff
@@ -0,0 +1,385 @@
...
+import static org.mockito.ArgumentMatchers.anyInt;
+import static uk.gov.hmcts.darts.audio.enums.AudioRequestStatus.COMPLETED;
+import static uk.gov.hmcts.darts.audio.enums.AudioRequestStatus.OPEN;
+import static uk.gov.hmcts.darts.audio.enums.AudioRequestStatus.PROCESSING;
+import static uk.gov.hmcts.darts.common.enums.ObjectDirectoryStatusEnum.MARKED_FOR_DELETION;
+import static uk.gov.hmcts.darts.common.enums.ObjectDirectoryStatusEnum.STORED;
+
+@Transactional
+@SuppressWarnings("PMD.ExcessiveImports")
+class OutboundAudioDeleterProcessorTest extends IntegrationBase {
```
tonymort, line 45: If this test fails without @Transactional it might mean that the code its testing will fail in production.  Worth having a look at exactly why this is needed to make the test pass and see if you can restructure it so that its not needed.

tonymort (changes requested):
Few comments

src/main/java/uk/gov/hmcts/darts/task/service/impl/AutomatedTaskServiceImpl.java
HMCTSDeveloper: ACCEPT

src/main/java/uk/gov/hmcts/darts/audio/config/AudioConfiguration.java
HMCTSDeveloper: its being used in the OutboundAudioDeleterProcessorImpl

src/main/java/uk/gov/hmcts/darts/common/service/bankholidays/Event.java
HMCTSDeveloper: ACCEPT

src/main/java/uk/gov/hmcts/darts/audio/config/AudioConfiguration.java
tonymort: It is, but its being injected directly into the OutboundAudioDeleterProcessorImpl's constructor.  I'm fairly confident that if you were to remove the line above it would still be available in the OutboundAudioDeleterProcessorImpl.

src/integrationTest/java/uk/gov/hmcts/darts/task/service/OutboundAudioDeleterProcessorTest.java
HMCTSDeveloper: This is happening because some entities being used have properties that are lazily fetched. In order to fetch them we always need a session open to the database. In the code we are using @transactional and OSIV which means we usually don't see this issue.

HMCTSDeveloper: Sorted it out it was because the object under test was being created manually instead of being autowried. Added a comment explain why we are doing that.

src/main/java/uk/gov/hmcts/darts/audio/config/AudioConfiguration.java
HMCTSDeveloper: ACCEPT.

tonymort (approved):

PR #626, opened 2023-10-31
DMP-938: added darts-gateway and stub to docker compose

tonymort:
Added the gateway and stub to docker compose.  Also added convenience scripts for bring up docker compose with the additional services.

```
[ ] Yes
[ ] No
```

### DIFF
--- a/bin/dbuild
+++ b/bin/dbuild
@@ -3,11 +3,16 @@
 project_dir=$(pwd)
 mkdir -p ./bin/logs
 
-git clone git@github.com:hmcts/"$1".git &>> "$project_dir"/bin/logs/"$1".log
+logfile="$project_dir"/bin/logs/"$1".log
+if [ -f "$logfile" ]; then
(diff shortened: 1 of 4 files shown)


bin/dbuild
```diff
@@ -0,0 +1,19 @@
...
+
+project_dir=$(pwd)
+mkdir -p ./bin/logs
+
+logfile="$project_dir"/bin/logs/"$1".log
+if [ -f "$logfile" ]; then
+  rm "$logfile"
+fi
+
+git clone git@github.com:hmcts/"$1".git &>> "$logfile"
```
karen-hedges, line 11: How much space is this going to add to a persons and where is this being cloned into?

tonymort: None, the repo is deleted after the image is built and log file is recreated on each build so it will always be quite short

richardbull (approved):

PR #628, opened 2023-10-31
postgres_app_v53_sec_v17

barry-jones50410:

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -154,6 +154,13 @@
 --    added: GRANT SELECT,UPDATE ON  nod_seq TO darts_user;
 --    updated annotation_document - all columns NOT NULL
 --    updated transcription_document - all columns NOT NULL
+--v53 amend transaction.hearing_date from TIEMSTAMP to DATE, to be consistent with hearing.hearing_date
```
richardbull, line 157: Quite a few typos in this comment

src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -438,6 +446,7 @@ CREATE TABLE courthouse
 (cth_id                      INTEGER                       NOT NULL
 ,courthouse_code             INTEGER                       NOT NULL          UNIQUE
```
richardbull, line 447: Is INTEGER the correct type for this column - `courthouse_code`?

src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -795,12 +804,17 @@ IS 'primary key of judge';
...
+,met_id                      INTEGER                       NOT NULL
 ,media_object_id             CHARACTER VARYING(16)
 ,channel                     INTEGER                       NOT NULL -- 1,2,3,4 or rarely 5
 ,total_channels              INTEGER                       NOT NULL --99.9% are "4" in legacy, occasionally 1,2,5 
 ,reference_id                CHARACTER VARYING             --all nulls in legacy
 ,start_ts                    TIMESTAMP WITH TIME ZONE      NOT NULL
 ,end_ts                      TIMESTAMP WITH TIME ZONE      NOT NULL
+,media_file                  CHARACTER VARYING             NOT NULL
+,media_format                CHARACTER VARYING             NOT NULL
+,file_size                   INTEGER                       NOT NULL
```
richardbull, line 816: Should this column be `bigint`?

src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -795,12 +804,17 @@ IS 'primary key of judge';
 CREATE TABLE media
 (med_id                      INTEGER                       NOT NULL
 ,ctr_id                      INTEGER
+,met_id                      INTEGER                       NOT NULL
```
richardbull, line 807: Will you be adding the INSERTS into the standing data script we have? So that we know what should be in this table

src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -883,6 +897,10 @@ IS 'filename of the requested media object, possibly migrated from moj_transform
 COMMENT ON COLUMN media_request.output_format
 IS 'format of the requested media object, possibly migrated from moj_transformation_s';
 
+CREATE TABLE media_type
+(met_id                      INTEGER                       NOT NULL
+,media_type                  CHARACTER VARYING             NOT NULL
```
richardbull, line 902: Wondering whether this would be any different from the mime type presumably in media table
`media_format                CHARACTER VARYING             NOT NULL`?

src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -1194,6 +1213,8 @@ CREATE TABLE user_account
 ,description                 CHARACTER VARYING
 ,user_state                  INTEGER
 ,last_login_ts               TIMESTAMP WITH TIME ZONE
+,is_system_user              BOOLEAN                       NOT NULL
```
richardbull, line 1216: Will you be updating the standing data scripts for system users?

src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -1194,6 +1213,8 @@ CREATE TABLE user_account
 ,description                 CHARACTER VARYING
 ,user_state                  INTEGER
```
richardbull, line 1214: Wondering should this column be `BOOLEAN` like you have introduced in security_group and security_role tables.
Assuming it would make it easier to understand - is enabled TRUE etc?

richardbull (approved):

src/main/resources/db/reference/modernised_darts_schema.sql
barry-jones50410: transaction replaced by transcription and TIEMSTAMP replace by TIMESTAMP.  Does 2 equate to a few, but two too many for a single line. correction will be in v55

src/main/resources/db/reference/modernised_darts_schema.sql
barry-jones50410: what data type is expected ?
moj_courthouse_s.c_code is varchar in legacy and contains the full text courthouse_name, hence is being mapped to courthouse_name ( c_id in legacy is an incomplete list of the same full courthouse names).
The numeric courthouse code is found in moj_daily_list_s.c_crown_court_code and this is the intended content of courthouse_code, as per the column level comment defined on this column in the schema creation script. In legacy it takes integer values between 401 and 762.

src/main/resources/db/reference/modernised_darts_schema.sql
barry-jones50410: assuming you are referring to "file_size" this is currently INTEGER, which has an upper limit of 2,147,483,647 and I believe is intended to record the file size in bytes.  This implies a maximum recorded file size of 2048Mb or 2Gb.
The largest recorded object in Legacy is 797,993,232 ( according to both dm_sysobject_s.r_content_size and dmr_content_s.full_content_size).  I'm not sure whether there is an upper limit on the size of the external objects ( inc media), but looks like we're currently at about 40% of what INTEGER can accommodate. What's the anticipated maximum size of media in bytes ?

src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -795,12 +804,17 @@ IS 'primary key of judge';
...
 ,media_object_id             CHARACTER VARYING(16)
 ,channel                     INTEGER                       NOT NULL -- 1,2,3,4 or rarely 5
 ,total_channels              INTEGER                       NOT NULL --99.9% are "4" in legacy, occasionally 1,2,5 
 ,reference_id                CHARACTER VARYING             --all nulls in legacy
 ,start_ts                    TIMESTAMP WITH TIME ZONE      NOT NULL
 ,end_ts                      TIMESTAMP WITH TIME ZONE      NOT NULL
+,media_file                  CHARACTER VARYING             NOT NULL
+,media_format                CHARACTER VARYING             NOT NULL
+,file_size                   INTEGER                       NOT NULL
+,checksum                    CHARACTER VARYING
```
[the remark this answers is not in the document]
barry-jones50410: the correct location for checksum is still being considered, but from my understanding should be present on {<external/transient>_object_directory}  or { <annotation/transcription>_document, media}, but not both.  Hence intention would be to remove checksum from 2 or 3 tables once determined.

src/main/resources/db/reference/modernised_darts_schema.sql
barry-jones50410: As I understand it, this is being added to accommodate future enhancement, and currently this value will have 1 distinct value, corresponding to AUDIO. 
If there is a realistic chance that media_type might be extended to include VIDEO, we would need a second row, and at that point, i suspect the upper limit of INTEGER might become an issue for file_size

src/main/resources/db/reference/modernised_darts_schema.sql
richardbull: Done in https://github.com/hmcts/darts-api/pull/633

src/main/resources/db/reference/modernised_darts_schema.sql
barry-jones50410: re-visiting in light of comment below, regarding media_type, i note that the datatype of the two fields in legacy database are defined in both dm_sysobject_s and dmr_content_s, with a content_size of type number(10) and a full_content_size of float(126).  Throughout the live data, both fields are populated, as no objects exceed the capacity of number(10).  However, to ensure we aren't introducing any future limitation, not present in legacy, I agree BIGINT is a better choice, but still far lower capacity than that offered by FLOAT(126).  That is 1e18 vs 1e38 ! But based on current usage should be more than adequate.

src/main/resources/db/reference/modernised_darts_schema.sql
barry-jones50410: I agree, media_format should be sufficient to determine what might possibly be held in media_type.  If absolutely necessary, we might normalise media_format, to have an attribute to include acceptable values of { audio/video/... }. Awaiting clarification.

src/main/resources/db/reference/modernised_darts_schema.sql
barry-jones50410: i have no list of proposed system users, so no plan currently.
That said i note the standing data script, modernised_darts_standing_data.sql, has been updated to include is_system_user so appears the question has already been addressed

src/main/resources/db/reference/modernised_darts_schema.sql
barry-jones50410: the legacy data has four acceptable values for user_state, analysis would need to be performed on the 3 states that would effectively be merged to boolean FALSE, given we have one legacy user_state that permits access.
Also, if the change to a BOOL should be acceptable, the column name should change to user_access or something similar, as user_state = TRUE, seems a little unintuitive.

PR #633, opened 2023-11-01
postgres_app_v54_sec_v17

jackmaloney:


src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -1132,6 +1136,7 @@ CREATE TABLE transcription_document
 ,file_name                   CHARACTER VARYING             NOT NULL
 ,file_type                   CHARACTER VARYING             NOT NULL
 ,file_size                   INTEGER                       NOT NULL
+,checksum                    CHARACTER VARYING             NOT NULL
```
richardbull, line 1139: Similar to my comment on the v53 PR media checksum...
Why do we need the checksum column on `transcription_document` table as well as `external_object_directory`?

Is this expected to be calculated to correspond to the file hash value in the CONTENT-MD5 tag as a base64 encoded representation of the binary MD5 hash value

e.g. openssl md5 -binary "Test Document.doc" | base64
KQ9vVogyRdsnEvxyNQz77g==

FYI @hemantasharma1129

src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -160,6 +160,9 @@
 --    add media_type table, primary key and fk from media
 --    add is_system_user and account_guid to user_account
 --    add display_name to courthouse
+--v54 add checksum to annotation_document and transcription_document
+--    add display_state to transcription_type and transcription_urgency
+--    add display_name to transcription_status
```
richardbull, line 165: Why do we need this? What needs to happen to our https://github.com/hmcts/darts-api/blob/master/src/main/resources/db/reference/modernised_darts_standing_data.sql that keeps getting neglected with new NOT NULL columns

src/main/resources/db/reference/modernised_darts_schema.sql
hemantasharma1129: All checksum fields are moved to respective tables i.e media, transcription_document, annotation_document. Yes the checksum will be the MD5 hash value. I agree the checksum field in external_object_directory should be removed.

src/main/resources/db/reference/modernised_darts_schema.sql
richardbull: Done in commit a6f7b9e9

richardbull (approved):

PR #644, opened 2023-11-03
Adding `/preview2` audio preview endpoint

davet1985:

N/A

### Change description ###

Introducing a second audio preview endpoint (/preview2) which returns the file in bytes and allows for a range header to be specified.

I'd like to try this on the portal to see if it makes any difference to the streaming issues we've been having.

```
[] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
@@ -119,8 +119,14 @@ public class AudioController implements AudioApi {
     }
 
     public byte[] readByteRange(byte[] wholeFile, long start, long end) {
+        int srcPos;
+        if (start > Integer.MIN_VALUE && start < Integer.MAX_VALUE)
+            srcPos = (int)start;
+        else
+            throw new IllegalArgumentException("Invalid input: start bytes truncated");
+
         byte[] result = new byte[(int) (end - start) + 1];
-        System.arraycopy(wholeFile, (int) start, result, 0, result.length);
+        System.arraycopy(wholeFile, srcPos, result, 0, result.length);
         return result;
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
```diff
@@ -77,12 +80,62 @@ public ResponseEntity<Resource> preview(Integer mediaId) {
         return new ResponseEntity<>(new InputStreamResource(audioMediaFile), HttpStatus.OK);
     }
 
+    @SneakyThrows
+    @Override
+    @SecurityRequirement(name = SECURITY_SCHEMES_BEARER_AUTH)
+    @Authorisation(contextId = MEDIA_ID,
+        securityRoles = {JUDGE, REQUESTER, APPROVER, TRANSCRIBER, LANGUAGE_SHOP_USER, RCJ_APPEALS})
+    public ResponseEntity<byte[]> preview2(Integer mediaId, @RequestHeader(value = "Range", required = false) String httpRangeList) {
```
richardbull, line 88: Should this be `range` to match the audio.yaml `in: header`?

src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
```diff
@@ -77,12 +80,62 @@ public ResponseEntity<Resource> preview(Integer mediaId) {
...
+                rangeEnd = Long.parseLong(ranges[1]);
+            } else {
+                rangeEnd = fileSize - 1;
+            }
+            if (fileSize < rangeEnd) {
+                rangeEnd = fileSize - 1;
+            }
+            String contentLength = String.valueOf((rangeEnd - rangeStart) + 1);
+            String contentRange = "bytes" + " " + rangeStart + "-" + rangeEnd + "/" + fileSize;
+            return ResponseEntity.status(HttpStatus.OK)
```
richardbull, line 113: @davet1985 Not sure it will help, but wondering if the status here should be PARTIAL_CONTENT (206)?
It might do something in combination with the special Content-Range header perhaps...

davet1985: @richardbull yes it should be, thank you for pointing that out 🤦

src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
davet1985: I'm not sure it makes a difference

richardbull: Sorry, think what you have here is correct - it was the yaml that needed to be the same case, then no need to override the generated param stuff. That's all I was meaning.

PR #649, opened 2023-11-06
Dmp 1409 v53 db

chrisdavecm:
https://tools.hmcts.net/jira/browse/DMP-1409

### Change description ###
Updated schema

```
[ ] Yes
[x] No
```


src/main/resources/openapi/transcriptions.yaml
```diff
@@ -437,6 +437,9 @@ components:
           type: string
           format: date-time
           example: '2023-07-31T14:32:24.000Z'
+        is_manual:
```
richardbull, line 440: I'm not sure why this was added to the transcriptions api endpoint?
Pretty sure I asked this in a refinement the other week and @jackmaloney didn't want the request body to be changed as the idea was that all users using this endpoint would be manual requests as the automatic requests just use the underlying service instead.

chrisdavecm: Ah didn't know where the auto was coming from, e.g. through the API or an internal job

src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/TranscriptionServiceImpl.java
```diff
@@ -297,6 +297,7 @@ private TranscriptionEntity saveTranscription(UserAccountEntity userAccount,
         transcription.setEndTime(transcriptionRequestDetails.getEndDateTime());
         transcription.setCreatedBy(userAccount);
         transcription.setLastModifiedBy(userAccount);
+        transcription.setIsManual(false);
```
richardbull, line 300: This Boolean will need to be parametised - please can you change, the `saveTranscriptionRequest`
https://github.com/search?q=repo%3Ahmcts%2Fdarts-api%20saveTranscriptionRequest&type=code

to introduce the new boolean separate from `TranscriptionRequestDetails` as then the automatic handler [TranscriptionRequestHandler.java](https://github.com/hmcts/darts-api/blob/2ee0701f9d71c5ef5b6c0545253d5df52f98e0da/src/main/java/uk/gov/hmcts/darts/event/service/impl/TranscriptionRequestHandler.java#L39) can pass in FALSE whereas the [TranscriptionController.java](https://github.com/hmcts/darts-api/blob/2ee0701f9d71c5ef5b6c0545253d5df52f98e0da/src/main/java/uk/gov/hmcts/darts/transcriptions/controller/TranscriptionController.java#L72) would pass in TRUE.

richardbull (approved):

src/main/resources/db/migration/common/V1_233__v53_updates.sql
```diff
@@ -0,0 +1,21 @@
+ALTER TABLE transcription ADD COLUMN hearing_date_tmp DATE;
+UPDATE transcription SET hearing_date_tmp = CAST(hearing_date AS DATE);
+ALTER TABLE transcription DROP COLUMN hearing_date;
+ALTER TABLE transcription RENAME COLUMN hearing_date_tmp TO hearing_date;
+ALTER TABLE transcription ADD COLUMN is_manual boolean NOT NULL DEFAULT FALSE;
```
richardbull, line 5: Ideally the old staging data would have the transcriptions updated to be is_manual=TRUE where the transcription_type is `999	Other`

richardbull: Sorry - yep it was supposed to be NOT 999

PR #651, opened 2023-11-07
Amending audio/preview2 to use different headers

davet1985:

N/A

### Change description ###

Trying with non-audio header to see if that's causing some issues.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
@@ -88,14 +88,14 @@ public class AudioController implements AudioApi {
     public ResponseEntity<byte[]> preview2(Integer mediaId, @RequestHeader(value = "Range", required = false) String httpRangeList) {
         InputStream audioMediaFile = audioService.preview(mediaId);
 
+        byte[] bytes = IOUtils.toByteArray(audioMediaFile);
         if (httpRangeList == null) {
-            byte[] bytes = IOUtils.toByteArray(audioMediaFile);
             return ResponseEntity.status(HttpStatus.OK)
-                .header("Content-Type", "audio/mpeg")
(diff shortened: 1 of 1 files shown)


richardbull (approved):

src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
```diff
@@ -88,14 +88,14 @@ public ResponseEntity<Resource> preview(Integer mediaId) {
     public ResponseEntity<byte[]> preview2(Integer mediaId, @RequestHeader(value = "Range", required = false) String httpRangeList) {
         InputStream audioMediaFile = audioService.preview(mediaId);
 
+        byte[] bytes = IOUtils.toByteArray(audioMediaFile);
         if (httpRangeList == null) {
-            byte[] bytes = IOUtils.toByteArray(audioMediaFile);
             return ResponseEntity.status(HttpStatus.OK)
-                .header("Content-Type", "audio/mpeg")
+                .header("Content-Type", "application/octet-stream")
+                .header("Content-Disposition", "attachment; filename=\"" + mediaId.toString() + ".mp3\"")
```
richardbull, line 95: I think this header will cause the browser to handle a file download - as this is what I'm using for the download transcript endpoint

test.mp3
richardbull, line 1: I'm guessing you didn't mean to commit this file as you have it under `src/main/resources/test.mp3`?

src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
davet1985: Yeah, I expect so, I'm just trying a different header to see if that's what's causing the request to be blocked.

test.mp3
davet1985: nope, removed

PR #663, opened 2023-11-09
Dmp 1297 add authorisation for xhibitcpp

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-1297

### Change description ###

Add Authorisation for endpoints used by XHIBIT/CPP

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/authorisation/annotation/Authorisation.java
```diff
@@ -18,4 +18,6 @@
 
     SecurityRoleEnum[] securityRoles();
 
+    SecurityRoleEnum[] globalAccessSecurityRoles();
```
richardbull, line 21: Think this should probably be `default {};`
So that we don't necessarily need to add this parameter, if not needed?

src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
```diff
@@ -45,7 +45,8 @@ public class AudioController implements AudioApi {
     @Override
     @SecurityRequirement(name = SECURITY_SCHEMES_BEARER_AUTH)
     @Authorisation(contextId = HEARING_ID,
-        securityRoles = {JUDGE, REQUESTER, APPROVER, TRANSCRIBER, LANGUAGE_SHOP_USER, RCJ_APPEALS})
+        securityRoles = {JUDGE, REQUESTER, APPROVER, TRANSCRIBER, LANGUAGE_SHOP_USER, RCJ_APPEALS},
```
richardbull, line 48: Wondering why we would have JUDGE also under securityRoles, if we're thinking that they will never actually have a link to courthouses via `security_group_courthouse_ae`?

src/main/java/uk/gov/hmcts/darts/authorisation/component/impl/UserIdentityImpl.java
```diff
@@ -24,43 +31,89 @@ public class UserIdentityImpl implements UserIdentity {
...
+            }
+        }
+        return null;
+    }
+
     public UserAccountEntity getUserAccount() {
-        return userAccountRepository.findByEmailAddressIgnoreCase(getEmailAddress())
-            .orElseThrow(() -> new DartsApiException(USER_DETAILS_INVALID));
+        UserAccountEntity userAccount = null;
+        String guid = getGuidFromToken();
```
richardbull, line 84: The getPrincipal JWT stuff inside the 2 private methods should probably move into this top level method and be passed in

src/main/java/uk/gov/hmcts/darts/authorisation/config/AuthorisationAspect.java
```diff
@@ -52,13 +54,25 @@ public Object handleRequestBodyAuthorisationAdvice(ProceedingJoinPoint joinPoint
...
         }
 
-        JsonNode jsonNode = objectMapper.valueToTree(body);
+        boolean hasGlobalAccess = false;
 
-        controllerAuthorisationFactory.getHandler(authorisationAnnotation.contextId()).checkAuthorisation(jsonNode, roles);
+        if (!globalAccessRoles.isEmpty()) {
+            hasGlobalAccess = userIdentity.userHasGlobalAccess(globalAccessRoles);
+        }
+        //TODO
```
richardbull, line 68: Remember to remove this comment

src/main/resources/db/migration/h2/V1_235_1__AddExternalTestSecurityGroups.sql
```diff
@@ -0,0 +1,9 @@
+INSERT INTO darts.user_account (usr_id, user_name, user_email_address, account_guid, is_system_user) values (-40, 'Xhibit', 'xhibit@hmcts.net', '43cb3e21-b03a-4629-9908-b0ddaf7448b9', true);
```
richardbull, line 1: Do we need to update the `modernised_darts_standing_data.sql` with these new system user details?

karen-hedges: no as these are out test users and not ones for production.

src/main/resources/db/migration/h2/V1_235_1__AddExternalTestSecurityGroups.sql
```diff
@@ -0,0 +1,9 @@
...
+INSERT INTO darts.user_account (usr_id, user_name, user_email_address, account_guid, is_system_user) values (-41, 'Cpp', 'cpp@hmcts.net', '43cb3e21-b03a-4629-9908-b0ddaf7448b9', true);
+
+insert into darts.security_group(grp_id, rol_id,group_name, global_access) values (-14, 7, 'Xhibit Group', true);
+insert into darts.security_group(grp_id, rol_id,group_name, global_access) values (-15, 8, 'Cpp Group', true);
+
+UPDATE darts.user_account
+SET account_guid = '43cb3e21-b03a-4629-9908-b0ddaf7448b9'
+WHERE usr_id = '-999';
```
richardbull, line 9: Not sure about the single quotes around the integer -999... but if you're sure it works in both places?

src/main/java/uk/gov/hmcts/darts/authorisation/component/impl/UserIdentityImpl.java
karen-hedges: Just started changing this and it would still need to be in 2 places as it needs to be in 2 places, it also needs to in userHasGlobalAccess. I don't want to be calling the getUserAccount() and then making another call to get the useraccount for the global  access query

src/functionalTest/java/uk/gov/hmcts/darts/configuration/AzureAdB2CAuthenticationProperties.java
```diff
@@ -17,5 +17,7 @@ public class AzureAdB2CAuthenticationProperties {
     private String password;
     private String clientId;
     private String clientSecret;
+    private String globalUsername;
```
richardbull, line 20: Do we need these 2 properties here - or have they been replaced by `AzureAdB2CGlobalAuthenticationProperties` ?

src/main/java/uk/gov/hmcts/darts/dailylist/controller/DailyListController.java
```diff
@@ -51,6 +58,10 @@ public class DailyListController implements DailyListsApi {
     private DailyListProcessor processor;
 
     @Override
+    @SecurityRequirement(name = SECURITY_SCHEMES_BEARER_AUTH)
+    @Authorisation(contextId = ANY_ENTITY_ID,
+        securityRoles = {},
```
richardbull, line 63: Will you be removing the empty `securityRoles = {},` now that we have a default in the annotation?

src/main/java/uk/gov/hmcts/darts/transcriptions/controller/TranscriptionController.java
```diff
@@ -80,7 +81,8 @@ public ResponseEntity<RequestTranscriptionResponse> requestTranscription(
 
     @Override
     @SecurityRequirement(name = SECURITY_SCHEMES_BEARER_AUTH)
-    @Authorisation(contextId = TRANSCRIPTION_ID, securityRoles = {APPROVER, TRANSCRIBER})
+    @Authorisation(contextId = TRANSCRIPTION_ID, securityRoles = {APPROVER, TRANSCRIBER},
+        globalAccessSecurityRoles = {})
```
richardbull, line 85: Will you be removing the empty `globalAccessSecurityRoles = {}` now that we have a default value in the annotation?

src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
karen-hedges: Its for failover in case a judge is not in a global security group.

richardbull (approved):

PR #669, opened 2023-11-10
DMP-1417 Save filename & fileformat at the end of ATS process

johnrace0:
https://tools.hmcts.net/jira/browse/DMP-1417

### Change description ###
Save filename & fileformat at the end of ATS process

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
@@ -48,6 +48,7 @@ import java.util.concurrent.ExecutionException;
 
 import static uk.gov.hmcts.darts.audio.enums.AudioRequestStatus.FAILED;
 import static uk.gov.hmcts.darts.audio.enums.AudioRequestStatus.PROCESSING;
+import static uk.gov.hmcts.darts.audiorequests.model.AudioRequestType.DOWNLOAD;
 import static uk.gov.hmcts.darts.common.enums.ExternalLocationTypeEnum.UNSTRUCTURED;
 import static uk.gov.hmcts.darts.common.enums.ObjectDirectoryStatusEnum.STORED;
 
@@ -205,10 +206,9 @@ public class AudioTransformationServiceImpl implements AudioTransformationServic
             throw new DartsApiException(AudioApiError.FAILED_TO_PROCESS_AUDIO_REQUEST, e);
         }
 
-        AudioRequestOutputFormat audioRequestOutputFormat;
-        switch (mediaRequestEntity.getRequestType()) {
-            case DOWNLOAD -> audioRequestOutputFormat = AudioRequestOutputFormat.ZIP;
-            default -> audioRequestOutputFormat = AudioRequestOutputFormat.MP3;
+        AudioRequestOutputFormat audioRequestOutputFormat = AudioRequestOutputFormat.MP3;
+        if (mediaRequestEntity.getRequestType().equals(DOWNLOAD)) {
+            audioRequestOutputFormat = AudioRequestOutputFormat.ZIP;
         }
         String fileName = mediaRequestEntity.getHearing().getCourtCase().getCaseNumber() + "_" + mediaRequestEntity.getHearing().getHearingDate();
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
```diff
@@ -205,7 +206,13 @@ private UUID processAudioRequest(Integer requestId) {
...
         }
 
-        mediaRequestService.updateAudioRequestStatus(mediaRequestEntity.getId(), COMPLETED);
+        AudioRequestOutputFormat audioRequestOutputFormat = AudioRequestOutputFormat.MP3;
+        if (mediaRequestEntity.getRequestType().equals(DOWNLOAD)) {
+            audioRequestOutputFormat = AudioRequestOutputFormat.ZIP;
+        }
+        String fileName = mediaRequestEntity.getHearing().getCourtCase().getCaseNumber() + "_" + mediaRequestEntity.getHearing().getHearingDate();
+
+        mediaRequestService.updateAudioRequestCompleted(requestId, fileName, audioRequestOutputFormat);
```
chrisbellingham-hmcts, line 215: The added code could throw an exception, e.g. if the `hearing` does not have a `courthouse` assigned, or if there's a hibernate connectivity issue. Could you move the code into the try block (L166-L192)?

The background - we need to be careful with exception handling in this method (`processAudioRequest`). When the method starts, the `AudioRequestStatus` is updated to `PROCESSING`. If an uncaught exception is thrown, the processing will quit, and the media request record in the DB will remain in the `PROCESSING` state. This will perpetually retried by KEDA, blocking processing of any other media request. As such we need to ensure all exceptions are caught in this method and handled by setting `AudioRequestStatus` to `FAILED`.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
```diff
@@ -205,7 +206,13 @@ private UUID processAudioRequest(Integer requestId) {
             throw new DartsApiException(AudioApiError.FAILED_TO_PROCESS_AUDIO_REQUEST, e);
         }
 
-        mediaRequestService.updateAudioRequestStatus(mediaRequestEntity.getId(), COMPLETED);
+        AudioRequestOutputFormat audioRequestOutputFormat = AudioRequestOutputFormat.MP3;
+        if (mediaRequestEntity.getRequestType().equals(DOWNLOAD)) {
+            audioRequestOutputFormat = AudioRequestOutputFormat.ZIP;
+        }
+        String fileName = mediaRequestEntity.getHearing().getCourtCase().getCaseNumber() + "_" + mediaRequestEntity.getHearing().getHearingDate();
```
chrisbellingham-hmcts, line 213: A `hearing` is already declared on L163, can we use that here rather than invoking `mediaRequestEntity.getHearing()` twice?

chrisbellingham-hmcts, line 213: The ticket suggests the date format in the string should be in the form `20_OCT_2023`, but string conversion of `getHearingDate`'s `LocalDate` will produce `YYYY-MM-DD` (e.g. `2023-10-20`).

I agree what you have here is better, but worth clarifying if that format is important.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
chrisbellingham-hmcts, line 1: I think a test of the processAudioRequest method, either unit or integration, would be useful to verify that:

- If an exception is thrown by the new code that AudioRequestStatus is set to FAILED
- That the MP3/DOWNLOAD logic works, and that the `fileName` is constructed properly.

chrisbellingham-hmcts (changes requested):

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
johnrace0: Moved code inside try block. Moved near start so if it fails it will fail early.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
johnrace0: Now using declared hearing object

johnrace0: Changed date format to same as Jira ticket - can always amend it later if required 👍

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
johnrace0: Added a couple of unit tests for null request type / null case

chrisbellingham-hmcts (approved):

PR #674, opened 2023-11-13
Reporting duplicate transcription id in error detail

davet1985:

https://tools.hmcts.net/jira/browse/DMP-762

### Change description ###

Adding the conflicting transcription ID to the error response.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/repository/TranscriptionRepository.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/repository/TranscriptionRepository.java
@@ -5,6 +5,7 @@ import org.springframework.data.jpa.repository.Query;
 import org.springframework.stereotype.Repository;
 import uk.gov.hmcts.darts.common.entity.TranscriptionEntity;
 import uk.gov.hmcts.darts.common.entity.TranscriptionStatusEntity;
+import uk.gov.hmcts.darts.common.entity.TranscriptionTypeEntity;
 
 import java.time.OffsetDateTime;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/darts/common/repository/TranscriptionRepository.java
```diff
@@ -39,4 +40,20 @@ List<TranscriptionEntity> findAllByTranscriptionStatusNotInWithCreatedDateTimeBe
...
     )
     List<TranscriptionEntity> findByHearingId(Integer hearingId);
+
+    @Query("""
+        SELECT t
+        FROM TranscriptionEntity t, HearingEntity hearing
+        WHERE hearing.id = :hearingId
+        AND t.hearing = hearing
+        AND t.transcriptionType = :transcriptionType
+        AND t.startTime = :startTime
```
richardbull, line 50: I'm wondering if this query copes with the startTime or endTime being NULL?
Thinking we might need to tweak the condition to check `IS NULL`

jackmaloney: @davet1985 could you write a test to cover Richards scenario and see what happens? In fact, looks like you might need to write some unit tests to cover all scenarios?

src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/TranscriptionServiceImpl.java
```diff
@@ -137,20 +136,28 @@ public RequestTranscriptionResponse saveTranscriptionRequest(
...
                 getTranscriptionTypeById(transcriptionRequestDetails.getTranscriptionTypeId()),
-                getTranscriptionUrgencyById(transcriptionRequestDetails.getUrgencyId()),
-                isManual
+                transcriptionRequestDetails.getStartDateTime(),
+                transcriptionRequestDetails.getEndDateTime()
             );
-        } catch (DataIntegrityViolationException e) {
-            throw new DartsApiException(TranscriptionApiError.DUPLICATE_TRANSCRIPTION);
+            if (!matchingTranscriptions.isEmpty()) {
+                String errorDetail = "{ \"transcription_id\": " + matchingTranscriptions.get(0).getId().toString() + " }";
```
chrisbellingham-hmcts, line 147: Does the transcriptionId need to be represented in the response as JSON?

If so, we could enhance the DartsApiException to allow us to add arbitrary fields to the error response. So here it could be something like:

```json
{
  "type": "TRANSCRIPTION_107",
  "title": "A transcription already exists with these properties",
  "status": 409,
  "duplicate_transcription_id": "2"
}
```

chrisbellingham-hmcts: I'll draft up a quick patch and will send it your way.

chrisbellingham-hmcts (approved):

PR #678, opened 2023-11-13
DMP-1525 DateTime Issue in Get Transcription endpoint

mmmalik1:

https://tools.hmcts.net/jira/browse/DMP-1525

### Change description ###
There are 2 Get transcriptions endpoint which returns the data using CaseId (for Case screen) and Hearing Id (for hearing screen)

The 'requested_on' parameter in cases.yaml and hearings.yaml have been updated to date-time. 
–  generated file on Transcript.java model for both hearings and cases updated from LocalDate to OffsetDateTime return types
–  the map method in TranscriptionMapper. java for both hearings and cases return requestedOn as OffsetDateTime.

All associated java and json integration test files updated from LocalDate to OffsetDateTime return types for both hearings and cases.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/resources/openapi/cases.yaml
+++ b/src/main/resources/openapi/cases.yaml
@@ -625,8 +625,8 @@ components:
           example: "Sentencing remarks"
         requested_on:
           type: string
-          format: date
-          example: 2022-05-22
+          format: date-time
+          example: 2023-06-20T10:00:00Z
         requested_by_name:
           type: string
           example: "Joe Bloggs"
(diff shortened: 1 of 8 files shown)


src/main/resources/openapi/cases.yaml
```diff
@@ -625,7 +625,7 @@ components:
           example: "Sentencing remarks"
         requested_on:
           type: string
-          format: date
+          format: date-time
           example: 2022-05-22
```
richardbull, line 629: Looks like the 2 examples need to change to be date-time also?

mmmalik1: hearings and cases yaml files updated to reflect 'requested_on' examples as 2023-06-20T10:00:00Z

richardbull (approved):

PR #686, opened 2023-11-14
Amending transcriptions openapi spec

davet1985:

N/A

### Change description ###

- to include the extra property returning the POST /transcriptions 409 error

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/openapi/transcriptions.yaml
+++ b/src/main/resources/openapi/transcriptions.yaml
@@ -60,6 +60,23 @@ paths:
                         $ref: '#/components/schemas/RequestTranscription400ErrorCode'
         '401':
           description: Unauthorised Error
+        '403':
+          description: Forbidden Error
+          content:
+            application/json+problem:
+              schema:
+                allOf:
+                  - $ref: './problem.yaml'
+                  - type: object
+                    required:
+                      - type
+                    properties:
+                      type:
+                        $ref: '#/components/schemas/TranscriptionAuthorisation403ErrorCode'
+              example:
+                type: "AUTHORISATION_100"
+                title: "User is not authorised for the associated courthouse"
+                status: 403
         '404':
           description: Not Found Error
         '409':
--- a/src/main/resources/openapi/transcriptions.yaml
+++ b/src/main/resources/openapi/transcriptions.yaml
@@ -62,6 +62,22 @@ paths:
           description: Unauthorised Error
         '404':
           description: Not Found Error
+        '409':
+          description: Duplicate transcription request
+          content:
+            application/json+problem:
+              schema:
+                allOf:
+                  - $ref: './problem.yaml'
+                  - type: object
+                    required:
+                      - type
+                    properties:
+                      type:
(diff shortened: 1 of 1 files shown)


src/main/resources/openapi/transcriptions.yaml
```diff
@@ -60,8 +60,43 @@ paths:
                         $ref: '#/components/schemas/RequestTranscription400ErrorCode'
         '401':
           description: Unauthorised Error
+          content:
```
richardbull, line 63: Don't think the 401 and 404 errors need expanding here.
They don't look like the standard problem error responses, the body is just empty.

401 is when the Azure access token has expired for example.
404 looks to be similar, so not the one you have added to the schema https://github.com/hmcts/darts-api/blob/master/src/main/java/uk/gov/hmcts/darts/transcriptions/controller/TranscriptionController.java#L77

src/main/resources/openapi/transcriptions.yaml
```diff
@@ -60,8 +60,43 @@ paths:
...
+            application/json+problem:
+              schema:
+                $ref: './problem.yaml'
+              example:
+                type: "AUTHORISATION_100"
+                title: "User is not authorised for the associated courthouse"
+                status: 401
```
richardbull, line 70: I think this problem / example should be moved under 403 Forbidden Error - it looks to have been missed when @karen-hedges added the `@Authorisation` to the TranscriptionController

src/main/resources/openapi/transcriptions.yaml
davet1985: Removed.

src/main/resources/openapi/transcriptions.yaml
davet1985: Removed

src/main/resources/openapi/transcriptions.yaml
```diff
@@ -469,7 +504,17 @@ components:
         - "TRANSCRIPTION_104"
         - "TRANSCRIPTION_106"
         - "TRANSCRIPTION_107"
-      x-enum-varnames: [ FAILED_TO_VALIDATE_TRANSCRIPTION_REQUEST, BAD_REQUEST_TRANSCRIPTION_TYPE, BAD_REQUEST_TRANSCRIPTION_URGENCY, DUPLICATE_TRANSCRIPTION ]
+      x-enum-varnames: [ FAILED_TO_VALIDATE_TRANSCRIPTION_REQUEST, BAD_REQUEST_TRANSCRIPTION_TYPE, BAD_REQUEST_TRANSCRIPTION_URGENCY ]
+    RequestTranscription404ErrorCode:
```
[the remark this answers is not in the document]
davet1985: Removed

richardbull (approved):

src/main/resources/openapi/transcriptions.yaml
davet1985: Re-added back in the correct place.

PR #688, opened 2023-11-14
DMP-1068 audio playback audit

davet1985:

https://tools.hmcts.net/jira/browse/DMP-1068

### Change description ###

The playback audit was already completed, but I changed it to record the `user_id` from the JWT (UserIdentity) rather than the `requestor` on the audio request.

I've also added an internal interface for the `Audit` package and refactored all usages to point to this rather than the service.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/audit/service/AuditService.java
```diff
@@ -11,8 +11,6 @@
 
 public interface AuditService {
 
-    void recordAuditRequestAudio(AuditActivityEnum activity, Integer userAccountEntity, Integer courtCase);
-
     @Transactional
```
richardbull, line 14: This `@Transactional` can be removed from the interface, since we have it on the implementing class which I think is more normal

richardbull (approved):

src/main/java/uk/gov/hmcts/darts/audit/service/AuditService.java
davet1985: Done

src/main/java/uk/gov/hmcts/darts/audio/controller/AudioRequestsController.java
```diff
@@ -102,9 +102,7 @@ public ResponseEntity<AddAudioResponse> addAudioRequest(AudioRequestDetails audi
         try {
             audioRequest = mediaRequestService.saveAudioRequest(audioRequestDetails);
             addAudioResponse = audioRequestResponseMapper.mapToAddAudioResponse(audioRequest);
-            auditService.recordAuditRequestAudio(AuditActivityEnum.REQUEST_AUDIO,
-                                                 audioRequestDetails.getRequestor(), audioRequestDetails.getHearingId()
-            );
+            auditApi.recordAudit(AuditActivityEnum.REQUEST_AUDIO, audioRequest.getRequestor(), audioRequest.getHearing().getCourtCase());
```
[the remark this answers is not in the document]
davet1985: I've brought this inline with other usages, moving it into the service.

davet1985:
@richardbull would you mind giving this another once over please, I've made a few changes since your last review.

src/main/java/uk/gov/hmcts/darts/audit/api/AuditActivity.java
```diff
@@ -1,11 +1,11 @@
-package uk.gov.hmcts.darts.audit.enums;
+package uk.gov.hmcts.darts.audit.api;
```
richardbull, line 1: Why do we want to change the package from `enums` and the enum name?
Only saying because we have it the other way in more places as we like a subpackage

src/main/java/uk/gov/hmcts/darts/audit/service/AuditService.java
```diff
@@ -11,10 +10,7 @@
 
 public interface AuditService {
 
-    void recordAuditRequestAudio(AuditActivityEnum activity, Integer userAccountEntity, Integer courtCase);
-
-    @Transactional
-    void recordAudit(AuditActivityEnum activity, UserAccountEntity userAccountEntity, CourtCaseEntity courtCase);
+    void recordAudit(AuditActivity activity, UserAccountEntity userAccountEntity, CourtCaseEntity courtCase);
```
richardbull, line 13: It's annoying that we have a mix of parameter naming conventions in the same method
e.g. CourtCaseEntity - but it was like this before...

src/main/java/uk/gov/hmcts/darts/audit/api/AuditActivity.java
davet1985: I wanted to keep it within the "api" package since they are used by other packages internally. Should I add a sub-package so it would be `audit.api.enums`?

src/main/java/uk/gov/hmcts/darts/audit/service/impl/AuditServiceImpl.java
```diff
@@ -1,57 +1,35 @@
...
 import uk.gov.hmcts.darts.common.repository.AuditActivityRepository;
 import uk.gov.hmcts.darts.common.repository.AuditRepository;
-import uk.gov.hmcts.darts.common.repository.HearingRepository;
-import uk.gov.hmcts.darts.common.repository.UserAccountRepository;
 
 import java.time.OffsetDateTime;
 import java.util.List;
-import java.util.Optional;
 
 @AllArgsConstructor
```
richardbull, line 23: I think we're generally trying to use `@RequiredArgsConstructor` now with `private final` fields that are used in the constructor arg injection

davet1985: This hasn't changed...

src/main/java/uk/gov/hmcts/darts/audit/api/AuditActivity.java
richardbull: I'm happy to keep it as is - I get annoyed at the best of time with our love for subpackages.
Seems fine to me!

src/main/java/uk/gov/hmcts/darts/audit/service/impl/AuditServiceImpl.java
richardbull: Yep it was a FYI as it jumped out at me

PR #710, opened 2023-11-17
DMP-1363 Admin User Authorisation

richardbull:

[DMP-1363](https://tools.hmcts.net/jira/browse/DMP-1363)

### Change description ###

Update security schema and example use of Authorisation annotation against `GET /users` endpoint
- security_role 11 | ADMIN
- security_group 1 | ADMIN 

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/usermanagement/controller/UserController.java
```diff
@@ -0,0 +1,26 @@
...
+import static uk.gov.hmcts.darts.authorisation.constants.AuthorisationConstants.SECURITY_SCHEMES_BEARER_AUTH;
+import static uk.gov.hmcts.darts.authorisation.enums.ContextIdEnum.ANY_ENTITY_ID;
+import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.ADMIN;
+
+@RestController
+public class UserController implements UserApi {
+
+    @Override
+    @SecurityRequirement(name = SECURITY_SCHEMES_BEARER_AUTH)
+    @Authorisation(contextId = ANY_ENTITY_ID, globalAccessSecurityRoles = ADMIN)
```
karen-hedges, line 21: Would this be more suited to a new context courthouse id?

richardbull: I'm not sure it needs a new contextId?  I only used `ANY_ENTITY_ID` here as I needed to pick one in the current Authorisation annotation. Unless we change the contextId to be an array with a default like the others so that I can just have `@Authorisation(globalAccessSecurityRoles = ADMIN)`
but then I'd have to mess about with the validation we have in our AuthorisationAspect or introduce a new annotation for `@GlobalAuthorisation`.

I didn't think it was worth me changing the User parameter endpoints for this example, although I do think the parameters should be renamed to be more consistent with other endpoints now that they are integer e.g. `courthouse_id`

karen-hedges (approved):

PR #720, opened 2023-11-21
Dmp 1296 add authorisation for dar pc mid tier

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-1296

### Change description ###

Add Authorisation for endpoints used by DAR_PC/MID_TIER

```
[ ] Yes
[x ] No
```


src/main/resources/db/migration/local/V1_243_2__add_external_test_security_groups.sql
```diff
@@ -0,0 +1,49 @@
...
+
+INSERT INTO darts.security_group_user_account_ae
+(usr_id, grp_id)
+VALUES(-60, -16);
+
+UPDATE darts.user_account
+SET account_guid = '9797a4fc-8e62-4095-bda6-04e7747b05e5'
+WHERE usr_id = -46;
+
+-- due to a clash of user accounts whereby there are now 2 darts_global_test_users, rename the one on test -44 and the one on staging -48 N.B. can't remove users due to foreign keys
```
richardbull, line 40: @hemantasharma1129 / @karen-hedges I don't understand why there are 2 users tbh. Would we be expecting to set this global user for testing purposes to be an ADMIN group. I did this locally to be able to use the new user search endpoint: /users/search

```
-- darts.global.user@hmcts.net
INSERT INTO security_group_user_account_ae (usr_id, grp_id)
VALUES (-46, 1);
```

richardbull: Locally, I appear to have usr_id: 44, 46 as darts.global.user@hmcts.net?

src/main/resources/db/migration/local/V1_243_2__add_external_test_security_groups.sql
```diff
@@ -0,0 +1,49 @@
+UPDATE darts.user_account
```
richardbull, line 1: I'm confused with what we have locally
`select * from user_account where is_system_user is false;`

It looks like we should be setting this flag for all these users (usr_id):
```
0
-1
-2
-3
-4
-40
-41
-42
-43
```
Please can you check this?

src/main/resources/db/migration/local/V1_243_2__add_external_test_security_groups.sql
karen-hedges: Local has 2 -44 and -46 and staging has 2 -46 and -48. This was because I added the first global user with -44 but only checked localhost id's instead of staging whereby someone else had created a user with that same id.  The OID I set was for the old azure settings. However, Hemanta was changing the azure stuff and consequently he created a new global user with the new azure OID. This then allowed testers to use both users. -46 appears on both DB's so it makes sense to make that one the main global user.

src/main/resources/db/migration/local/V1_243_2__add_external_test_security_groups.sql
karen-hedges: I set these specific system users and changed the OID. This was a change I did on staging with one of the testers. I figured it would be needed on our local db's. I didn't check if the others should also be set as system user.

src/functionalTest/java/uk/gov/hmcts/darts/configuration/AccessTokenClientConfiguration.java
```diff
@@ -11,11 +12,13 @@
 @EnableConfigurationProperties
 @RequiredArgsConstructor
 @Profile("functionalTest")
+@Slf4j
 public class AccessTokenClientConfiguration {
 
     private final AzureAdAuthenticationProperties adAuthenticationProperties;
     private final AzureAdB2CAuthenticationProperties b2cAuthenticationProperties;
     private final AzureAdB2CGlobalAuthenticationProperties b2cGlobalAuthenticationProperties;
+    private final AzureAdB2CDarPcMidtierGlobalAuthenticationProperties b2CDarPcMidtierGlobalAuthenticationProperties;
```
richardbull, line 21: Should this be b2cDarPcMidtierGlobalAuthenticationProperties to be more like the other b2c fields?

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerTest.java
```diff
@@ -123,7 +136,7 @@ void setupData() {
 
     @Test
     void casesGetEndpoint() throws Exception {
-
+        setupExternalDarPcUserForCourthouse(null);
```
richardbull, line 139: I'm finding it hard to follow why we pass in null here for this private method - it looks like you could do that inside that other method if you need to pass null through to the stub?

richardbull (approved):

PR #735, opened 2023-11-23
DMP-1531 - raise an error if an event mapping is not in the database.

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-1615
https://tools.hmcts.net/jira/browse/DMP-1558

database query has been moved up to the dispatcherImpl so that the difference between "no entry in the database" and "an entry exists in the database, but no handler matching it" can be determined.


src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerPostEventsTest.java
```diff
@@ -35,7 +35,6 @@
 @Transactional
```
tonymort, line 35: Can't remember if we discussed this particular instance of Transaction already, but if not can we avoid this on an int test?

tonymort (changes requested):

tonymort (approved):

PR #740, opened 2023-11-24
Added postman collection changes

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-1296

### Change description ###

Added postman collection changes

```
[ ] Yes
[x ] No
```

### DIFF
--- a/ModernisedDarts.postman_collection.json
+++ b/ModernisedDarts.postman_collection.json
@@ -1,8 +1,9 @@
 {
 	"info": {
-		"_postman_id": "5f17c956-c279-4474-a254-fe91f08073a5",
+		"_postman_id": "1d8ae6aa-4024-415a-a164-ce73657d03f3",
 		"name": "ModernisedDarts",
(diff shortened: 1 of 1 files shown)


ModernisedDarts.postman_collection.json
```diff
@@ -1940,6 +2006,64 @@
 					]
 				}
 			]
+		},
+		{
+			"name": "NodeRegistration",
```
richardbull, line 2011: Should this folder be `Devices` in keep with the current tag in the OAS?

richardbull (approved):

ModernisedDarts.postman_collection.json
karen-hedges: I used the name of the yaml file but can change it if you prefer?

richardbull: I guess so, but up to you :) - annoying history with the good old `node_register`.
But the path is also currently `/register-devices` so no mention of node

PR #743, opened 2023-11-24
DMP-1581: allow manual and auto transcription duplicates

tonymort:
Allow transcription request duplicates if one is automatic and the other manual

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/DuplicateRequestDetector.java
```diff
@@ -0,0 +1,50 @@
...
+        if (requestDetails.getStartDateTime() != null && requestDetails.getEndDateTime() != null) {
+
+            List<TranscriptionEntity> matchingTranscriptions = transcriptionRepository.findByHearingIdTypeStartAndEnd(
+                requestDetails.getHearingId(),
+                getTranscriptionTypeById(requestDetails.getTranscriptionTypeId()),
+                requestDetails.getStartDateTime(),
+                requestDetails.getEndDateTime()
+            );
+            if (!matchingTranscriptions.isEmpty() && matchesSourceOfTranscriptionRequest(isManual, matchingTranscriptions)) {
+                Integer duplicateTranscriptionId = matchingTranscriptions.get(0).getId();
```
cmagowan-hmcts, line 33: is this necessarily correct?
the matchingTranscriptions could contain a manual and a automatic, and the matchesSourceOfTranscriptionRequest just tells you if one of them matches the manual/automatic, but not which one, but you are assuming its the first one returned, which i dont think is necessarily the case?

tonymort: Not sure about this.  We only need 1 match on 'isManual' to declare that a duplicate exists.  We don't care which 1 do we? Just that a match exists.

cmagowan-hmcts (approved):

PR #765, opened 2023-11-30
DMP-1635 added validation on creating users with same email 

tonymort:
Added validation on creating users with same email address that takes into account existing user state

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/UserManagementServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/UserManagementServiceImpl.java
@@ -1,16 +1,16 @@
 package uk.gov.hmcts.darts.usermanagement.service.impl;
 
+import jakarta.transaction.Transactional;
 import lombok.RequiredArgsConstructor;
 import org.apache.commons.lang3.BooleanUtils;
 import org.springframework.stereotype.Service;
 import uk.gov.hmcts.darts.authorisation.api.AuthorisationApi;
 import uk.gov.hmcts.darts.common.entity.SecurityGroupEntity;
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/UserManagementServiceImpl.java
```diff
@@ -36,9 +37,14 @@ public class UserManagementServiceImpl implements UserManagementService {
     private final SecurityGroupRepository securityGroupRepository;
     private final AuthorisationApi authorisationApi;
     private final UserSearchQuery userSearchQuery;
+    private final List<UserCreationValidation> userCreationValidations;
+    private final List<UserModifyValidation> userModifyValidations;
```
chrisbellingham-hmcts, line 41: On reflection, I'm wondering about the flexibility of this approach. E.g. how easy would it be to add a second `UserCreationValidation` validator that accepted a different type? One example might be that we wish to validate a header passed into the `createUser` method, or something else that's not part of the `User` object.

Also in some cases it might make sense to perform validations in a specific known order, and the `List` injection makes it a little unclear in which order the validations would happen.

I do like the idea of having the validators separate as it makes them easier to unit test. One approach that would maximise flexibility and I think readability at the cost of a little more code would be to define a generic interface such as:
```java
public interface Validator<T> {
    void validate(T validatable);
}
```

And then create your impls as needed, e.g:

```java
public class DuplicateEmailValidation implements Validator<User> {
   @Override
    public void validate(User user) {
       ....    
   }
}
```

And then inject each impl into the service class here (rather than using the lists):
```
private final DuplicateEmailValidation duplicateEmailValidation;
```
And call the `.validate()` method(s) where needed.

What do you think, or are there other benefits to the List approach?

PR #785, opened 2023-12-06
DMP-1666-Fix-HearingEvents

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-1666

### Change description ###

Fixed hearing events

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/audio/service/impl/ViqHeaderServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/service/impl/ViqHeaderServiceImpl.java
@@ -17,6 +17,7 @@ import uk.gov.hmcts.darts.audio.util.XmlUtil;
 import uk.gov.hmcts.darts.common.entity.EventEntity;
 import uk.gov.hmcts.darts.common.entity.HearingEntity;
 import uk.gov.hmcts.darts.common.exception.DartsApiException;
+import uk.gov.hmcts.darts.common.repository.EventRepository;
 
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/darts/audio/service/impl/ViqHeaderServiceImpl.java
```diff
@@ -163,8 +166,8 @@ private List<EventEntity> getHearingEventsByStartAndEndTime(
         HearingEntity hearingEntity,
         ZonedDateTime startTime,
         ZonedDateTime endTime) {
-
-        return hearingEntity.getEventList().stream()
+        List<EventEntity> hearingEvents = eventRepository.findAllByHearingId(hearingEntity.getId());
```
richardbull, line 169: What needs to happen to the `HearingEntity#eventList` that's marked Deprecated? I notice @cmagowan-hmcts also left a comment on the ticket about it.  I'd expect that the JoinTable definition on the parent Hearing side to work ok if the service using the hearingEntity was marked Transactional scope - but tricky to follow through the layers. I guess the other workaround here is to find the Hearing by PK and then navigate to events. Probably best to check with the 2 Chris's... FYI @chrisbellingham-hmcts

cmagowan-hmcts: looks like the HearingEntity#eventList is being used by the advanced search, so might be tricky to remove.

PR #787, opened 2023-12-06
Test enhancements

mestebanez:

no jira for this as of yet

### Change description ###

Some enhancements to tests:-

* eliminate the order annotation. We should not be order dependant. This creates brittle tests
* lets use an asynchronous test utility rather than Thread.sleep for the following reasons:-

1) Thread.sleep time is not guaranteed as threads can be interrupted. This can create less reliable tests. 
2) Thread.sleep will always wait for the maximum wait time. By polling for a known state we can potentially wait for less time than we need too.
3) An async test framework will make assertions failures more clear when they do happen

```
[ ] Yes
[x] No
```


src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/wiremock/DartsGatewayStub.java
```diff
@@ -28,10 +29,11 @@ public void verifyDoesntReceiveDarEvent() {
 
     public void verifyReceivedNotificationType(int type) {
         var notificationType = "\"notification_type\":\"" + type + "\"";
-        wait(1000);
-        verify(exactly(1), postRequestedFor(urlEqualTo(DAR_NOTIFY_PATH))
-            .withRequestBody(containing(notificationType)));
-
+        waitForMax10SecondsWithOneSecondPoll(() -> {
```
richardbull, line 32: Do we need this in the method above also? - verifyDoesntReceiveDarEvent

src/main/java/uk/gov/hmcts/darts/common/config/security/SecurityConfig.java
```diff
@@ -51,7 +50,6 @@ public class SecurityConfig {
     private final InternalAuthProviderConfigurationProperties internalAuthProviderConfigurationProperties;
 
     @Bean
-    @Order(1)
```
richardbull, line 54: This is a different annotation than the others removed - `org.junit.jupiter.api.Order`
Comment for awareness - hopefully it's ok to be removed

richardbull (approved):

davet1985 (approved):

src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/wiremock/DartsGatewayStub.java
mestebanez: Kind of impossible i guess. How do we know when something is never going to happen. Only way is to wait the longest time and then check

src/integrationTest/java/uk/gov/hmcts/darts/testutils/AwaitabilityUtil.java
```diff
@@ -0,0 +1,29 @@
...
+import static org.awaitility.Awaitility.await;
+
+public final class AwaitabilityUtil {
+
+    private AwaitabilityUtil() {
+
+    }
+
+    public static void waitForMax10SecondsWithOneSecondPoll(Callable<Boolean> callable) {
+        await().atMost(Duration.of(10, ChronoUnit.SECONDS))
```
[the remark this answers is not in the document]
mestebanez: Done

src/main/java/uk/gov/hmcts/darts/common/config/security/SecurityConfig.java
mestebanez: Good point but I don't believe this will have a negative impact. We need a very good reason to order. We should heavily document any ordering that is required.

PR #789, opened 2023-12-06
Dmp 652 unstructured to arm push

garethmorriscgi:
https://tools.hmcts.net/jira/browse/DMP-652) 

### Change description ###
Automated job to move files from unstructured storage to ARM storage. 

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmProcessorImpl.java
```diff
@@ -0,0 +1,194 @@
...
+        }
+        if (externalObjectDirectoryEntity.getAnnotationDocumentEntity() != null) {
+            documentId = externalObjectDirectoryEntity.getAnnotationDocumentEntity().getId();
+        }
+
+        return String.format("%s_%s_%s", entityId, documentId, transferAttempts);
+    }
+
+    private void updateTransferAttempts(ExternalObjectDirectoryEntity externalObjectDirectoryEntity) {
+        int currentNumberOfAttempts = externalObjectDirectoryEntity.getTransferAttempts();
```
karen-hedges, line 191: Looking at staging this value seems to be set as null in some cases. Might be worth a null check

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmProcessorImpl.java
```diff
@@ -0,0 +1,194 @@
...
+
+    @Override
+    @Transactional
+    public void processUnstructuredToArm() {
+        processPendingUnstructured();
+    }
+
+    private void processPendingUnstructured() {
+
+        ObjectDirectoryStatusEntity storedStatus = objectDirectoryStatusRepository.getReferenceById(
```
karen-hedges, line 59: These are constants so would it make sense to initialise these in the constructor so there are less database calls made?

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmProcessorImpl.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    private void saveToArm(ExternalObjectDirectoryEntity unstructuredExternalObjectDirectoryEntity,
+                           ExternalObjectDirectoryEntity armExternalObjectDirectoryEntity) {
+        try {
+            String filename = generateFilename(armExternalObjectDirectoryEntity);
+            BinaryData inboundFile = dataManagementApi
+                .getBlobDataFromUnstructuredContainer(unstructuredExternalObjectDirectoryEntity.getExternalLocation());
+
+            armDataManagementApi.saveBlobDataToArm(filename, inboundFile);
```
chrisbellingham-hmcts, line 135: Is this just copying from one Blob container to another, by downloading and then uploading?

Would the following be more appropriate: https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-copy-java. From a quick reading it looks like it does the copy server-side which would be a more robust and performant approach?

If that's not appropriate, can you consider using the `blobClient` methods that use streams per https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=1694706043#BestPractices(Backend)-2.1Favourjava.ioStreamsOverLoadingLargeDataintoMemory?

This will help avoid reading the full binary data into memory, which might be considerable given the size of an audio file.

hemantasharma1129:
Functional Tests are required to prove the saving into the Blob storage with a directory structure is working.

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmProcessorImpl.java
karen-hedges: Its copying from our storage container to an arm storage container. I am in the process of trying to write a functional test for it to see if this does work. This functionality is going to be changed as it needs to use a different way of connecting so it might be best that the optimisation is moved to the ticket when its reworked so that other tickets dependent on this at the very minimum is working. We might have to do something like this: https://stackoverflow.com/questions/61117069/how-to-move-azure-blob-from-one-storage-container-to-another-using-java-rest?

PR #809, opened 2023-12-11
DMP-1346: replace `repository.findById` with `repository.getReferenceById`

mario-paniccia:
Note:

I've analyzed all the production code usages of `findById`.
Apart from one case, I believe all the current usages are legit and should not be replaced by `getReferenceById`. This is because:

- in most of the cases the (non-ID) entity fields are accessed with setters/getters, which would trigger a SQL operation anyway
- in a couple of cases replacing would break the API contract by changing the error currently returned in case the entity does not exist, which should be avoided


jackmaloney (approved):

jackmaloney:
Merged with https://github.com/hmcts/darts-api/pull/832

PR #823, opened 2023-12-13
DMP-1658: Apply authorisation to admin endpoints

chrisbellingham-hmcts:
# [DMP-1658](https://tools.hmcts.net/jira/browse/DMP-1658)

Restrict admin endpoints to users with the `ADMIN` role and update integration tests.

Assign the `darts_global_test_user` the `ADMIN` role to facilitate functional tests.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/UserAccountStub.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/UserAccountStub.java
@@ -6,9 +6,11 @@ import org.springframework.transaction.annotation.Transactional;
 import uk.gov.hmcts.darts.common.entity.CourthouseEntity;
 import uk.gov.hmcts.darts.common.entity.SecurityGroupEntity;
 import uk.gov.hmcts.darts.common.entity.UserAccountEntity;
+import uk.gov.hmcts.darts.common.enums.SecurityRoleEnum;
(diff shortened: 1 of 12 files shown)


src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/UserAccountStub.java
```diff
@@ -186,4 +188,14 @@ public UserAccountEntity createExternalUser(String guid, SecurityGroupEntity sec
         return testUser;
     }
 
+    public UserAccountEntity createAdminUser() {
+        var adminGroup = securityGroupRepository.findByGroupName(SecurityRoleEnum.ADMIN.name())
```
richardbull, line 192: This looks to me like it should be changed to be something like `findBySecurityGroupRole`? We use the security role enum id as it is a standing data table - security_role(rol_id). Since it's a test and we know that we have  a security_group with group_name = ADMIN - I guess it could just be a String constant here?

chrisbellingham-hmcts: Good spot, yeah I think `"ADMIN"` as a string literal is a reasonable approach here. If for some reason that name is changed, the associated tests will fail and should be quite obvious as to why.

richardbull (approved):

PR #824, opened 2023-12-13
dmp 1599 fixed type

chrisdavecm:

dmp-1599 bug fix

### Change description ###
Fix bug

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/common/entity/HearingReportingRestrictionsEntity.java
```diff
@@ -56,7 +58,7 @@ public class HearingReportingRestrictionsEntity {
     OffsetDateTime eventDateTime;
 
     @Column(name = "case_number")
-    String caseNumber;
+    private List<String> caseIdList = new ArrayList<>();
```
cmagowan-hmcts, line 61: this should be called caseNumberList so as not to be confused with caseIds which are different

cmagowan-hmcts (approved):

PR #868, opened 2023-12-22
DMP-1853: 

mario-paniccia:
Notes:
the ticket requests using a sas token, but the `BlobServiceClientBuilder` requires also an endpoint when the connection string is not provided (see the builder javadoc). Therefore I've used the sas endpoint, which contains both the endpoint and the sas token.

I've created a vault secret on staging for the sas endpoint: [ARMSasEndpoint](https://portal.azure.com/#@HMCTS.NET/asset/Microsoft_Azure_KeyVault/Secret/https://darts-stg.vault.azure.net/secrets/ARMSasEndpoint/6c3a45c092c441079af63ea234d8d77e), but I wasn't sure whether the secret needs to be configured on other environments too


mario-paniccia:
Closing the PR as there's a functional test failing, I will investigate once back from holidays

PR #870, opened 2024-01-02
Update User Management API spec per latest LLD

chrisbellingham-hmcts:
Changes to API spec to reflect Create User use-case per: https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=1722059332#AdminPortalLLDUserManagement-CreateaUserAccount.

See associated Figma design: https://www.figma.com/file/F9hGLvJg18Hya6gUodDXIZ/DARTS-Admin-(SPEC)?type=design&node-id=106-1901&mode=design&t=HyOxMsarFhLgwOON-4

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/openapi/usermanagement.yaml
+++ b/src/main/resources/openapi/usermanagement.yaml
@@ -43,7 +43,7 @@ paths:
           $ref: '#/components/responses/UserSuccessResponseWithId'
         '400':
           $ref: '#/components/responses/BadUserRequest'
-  /admin/users/{userId}:
+  /admin/users/{user_id}:
     get:
       tags:
         - User
@@ -288,7 +288,7 @@ components:
   parameters:
     UserId:
       in: path
-      name: userId
+      name: user_id
       required: true
       schema:
         type: integer
@@ -414,7 +414,7 @@ components:
       properties:
         id:
           $ref: '#/components/schemas/SecurityGroupId'
-        role_id:
+        security_role_id:
           $ref: '#/components/schemas/RoleId'
         global_access:
           $ref: '#/components/schemas/SecurityGroupGlobalAccess'
--- a/src/main/resources/openapi/usermanagement.yaml
+++ b/src/main/resources/openapi/usermanagement.yaml
@@ -18,7 +18,7 @@ paths:
       operationId: getUsers
       parameters:
         - in: query
-          name: courthouse
+          name: courthouse_id
           required: false
           schema:
             $ref: "#/components/schemas/CourthouseId"
--- a/src/main/resources/openapi/usermanagement.yaml
+++ b/src/main/resources/openapi/usermanagement.yaml
@@ -373,6 +373,8 @@ components:
       properties:
         full_name:
           $ref: '#/components/schemas/UserName'
+        email_address:
+          $ref: '#/components/schemas/UserEmailAddress'
         description:
           $ref: '#/components/schemas/UserDescription'
         active:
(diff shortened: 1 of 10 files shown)


src/main/resources/openapi/usermanagement.yaml
```diff
@@ -19,12 +19,20 @@ paths:
       parameters:
         - in: query
           name: courthouse
-          required: true
+          required: false
           schema:
             $ref: "#/components/schemas/CourthouseId"
+        - in: header
+          name: Email-Address
```
richardbull, line 26: I think this should be lower snake-case, so `email_address`

src/main/resources/openapi/usermanagement.yaml
```diff
@@ -19,12 +19,20 @@ paths:
       parameters:
         - in: query
           name: courthouse
```
richardbull, line 21: This query param should probably be renamed to be `courthouse_id`?

src/main/resources/openapi/usermanagement.yaml
chrisbellingham-hmcts: Hyphenated-Pascal-Case appears to be the expectation per HMCTS standards (https://hmcts.github.io/restful-api-standards/#132). Concious we use snake_case elsewhere for headers but I've chosen to align with the standard here.

src/main/resources/openapi/usermanagement.yaml
chrisbellingham-hmcts: Agree, updated.

richardbull (approved):

src/main/resources/openapi/usermanagement.yaml
```diff
@@ -124,6 +142,14 @@ paths:
         '400':
           $ref: '#/components/responses/SecurityGroupBadRequest'
 
+  /admin/roles:
```
richardbull, line 145: Perhaps this path (context) should be `security-roles` in keep with `security-groups`?
If you agree, then I think the tags (line 149) should also be updated to be similar

src/main/resources/openapi/usermanagement.yaml
```diff
@@ -18,13 +18,19 @@ paths:
       operationId: getUsers
       parameters:
         - in: query
-          name: courthouse
-          required: true
+          name: courthouse_id
```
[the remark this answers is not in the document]
chrisbellingham-hmcts: Have changed to `courthouse_id` for `GET /admin/security-groups`, didn't spot any other instances either.

Have also updated the `name` fields in the schema for query parameters `courthouse` and `user` to be `courthouseId` and `userId` respectively.

src/main/resources/openapi/usermanagement.yaml
chrisbellingham-hmcts: Yep fair point, this will also align us with naming in the DB schema as an added bonus. Updated.

src/main/resources/openapi/usermanagement.yaml
```diff
@@ -288,16 +288,10 @@ components:
   parameters:
     UserId:
       in: path
-      name: id
+      name: userId
       required: true
       schema:
         type: integer
-    CourthouseId:
```
chrisbellingham-hmcts, line 260: Unused, so removed.

src/main/resources/openapi/usermanagement.yaml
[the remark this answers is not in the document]
richardbull: Thanks (Prefer the path param to be underscore like the query params. Also thought the GET response objects would also have a similar underscore id.)

richardbull (approved):

PR #872, opened 2024-01-02
DMP-1625-ARM-CreateMediaManifestFile

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-1625

### Change description ###

Updated the way null pointer is handled in arm task

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/task/runner/impl/UnstructuredToArmAutomatedTask.java
+++ b/src/main/java/uk/gov/hmcts/darts/task/runner/impl/UnstructuredToArmAutomatedTask.java
@@ -34,6 +34,10 @@ public class UnstructuredToArmAutomatedTask extends AbstractLockableAutomatedTas
 
     @Override
     protected void handleException(Exception exception) {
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/darts/task/runner/impl/UnstructuredToArmAutomatedTask.java
```diff
@@ -32,8 +32,13 @@ protected void runTask() {
         unstructuredToArmProcessor.processUnstructuredToArm();
     }
 
+    @SuppressWarnings("java:S4507")
     @Override
     protected void handleException(Exception exception) {
+        if (exception instanceof NullPointerException npe) {
+            log.error("Exception with null: {}", npe.getMessage());
```
richardbull, line 39: I think you might need to add the `npe` as the last param on the log.error to see the stacktrace alongside your error message? Not sure about adding the printStackTrace bit below - unless just temporary

richardbull (approved):

PR #873, opened 2024-01-02
DMP-1397 - dont show automatic transcriptions

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-1397

### DIFF
--- a/src/integrationTest/resources/tests/cases/CaseControllerGetCaseTranscriptsTest/ignoreAutomaticTranscripts.json
+++ b/src/integrationTest/resources/tests/cases/CaseControllerGetCaseTranscriptsTest/ignoreAutomaticTranscripts.json
@@ -1,0 +1,23 @@
+[
+  {
+    "hearing_date": "2023-01-01",
+    "type": "SENTENCING_REMARKS",
+    "requested_on": "2023-06-20T10:00:00Z",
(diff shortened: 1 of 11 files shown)


src/integrationTest/resources/tests/cases/CaseControllerGetCaseTranscriptsTest/ignoreAutomaticTranscripts.json
```diff
@@ -0,0 +1,23 @@
+[
+  {
+    "hearing_date": "2023-01-01",
+    "type": "SENTENCING_REMARKS",
+    "requested_on": "2023-06-20T10:00:00Z",
```
richardbull, line 5: I think it would be better if each of these cases/hearings test data had different times, say minutes to make it more obvious that it is excluded

richardbull (approved):

src/integrationTest/resources/tests/cases/CaseControllerGetCaseTranscriptsTest/ignoreAutomaticTranscripts.json
cmagowan-hmcts: good point, have fixed

PR #888, opened 2024-01-05
DMP-1854_json_document converted to json_string for API

mmmalik1:
https://tools.hmcts.net/jira/browse/DMP-1854

### Change description ###
The gateway feign library can not work with the json_document header i.e. it does not marshall the json, instead it uses the object string. This likely relates to the fact that headers do not have a specific content type within HTTP.

```
[ ] Yes
[ x] No
```


mmmalik1:
New PR created

PR #899, opened 2024-01-08
DMP-1873 Added check for well formed audio

johnrace0:
https://tools.hmcts.net/jira/browse/DMP-1873

### Change description ###
ATS Gap - Handle Discrepancy in the Audio Count between channels

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/darts/audio/model/ChannelAudio.java
```diff
@@ -20,6 +20,10 @@ public ChannelAudio(Integer channel) {
         this.channel = channel;
     }
 
+    public ChannelAudio(List<AudioFileInfo> audioFiles) {
```
richardbull, line 23: I'm a bit confused about what we're doing recently with the audio model files.
There is also `AudioSession` that has no usages.

Please can you check that we need all the lombok annotations on this `ChannelAudio` class - I think it is preferable in a custom class to use the required args constructor and getter methods just? Rather than having the setter methods as well?

`AudioFileInfo` - why do we need both fileName and path if they look to be the same?
Not sure we really want `@Setter` - it would be better to just use the constructor

I know we use the fileName for the audio operation commands currently. I wasn't sure if we were intending maybe to use the media.media_file, but perhaps that got forgotten about.

src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
```diff
@@ -157,7 +195,7 @@ private List<List<AudioFileInfo>> concatenateByChannelWithGaps(List<AudioFileInf
                 audioFileInfosForChannel,
                 allowableAudioGap
```
richardbull, line 196: The property `darts.audio.allowable_audio_gap_duration` has underscores whereas the others are hyphenated.
I think it should be made hyphenated and added to the `AudioConfigurationProperties` class we have already. So we can get rid of the `@Value` annotation on line 35 OutboundFileProcessorImpl#allowableAudioGap?

richardbull: Another observation `AudioTransformationServiceImpl#handlePlaybacks` - I think this public method should be refactored (pushed up) into the private handlePlaybacks as it's a bit redundant and shouldn't be public access anyway?  IntelliJ shows a whinge on Line 350, so I think it's probably worth another a look to tidyup playback a bit.

src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
```diff
@@ -94,6 +105,33 @@ public List<AudioFileInfo> processAudioForPlaybacks(Map<MediaEntity, Path> media
         return concatenatedAndMergedAudioFileInfos;
     }
 
+    private boolean isWellFormedAudio(List<AudioFileInfo> audioFileInfos) {
```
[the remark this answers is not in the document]
johnrace0: gonna leave as is. there's 2 reasons:
1) if there's only 1 channel for audio data then it can be considered "well formed" regardless of any other criteria
2) if all the counts for start times match - it's only "well formed" if none of the counts = 1

src/main/java/uk/gov/hmcts/darts/audio/model/ChannelAudio.java
johnrace0: AudioSession/ChannelAudio were added by ChrisM. I don’t know what the intended use is for AudioSession so will leave as is. ChannelAudio is in use so I will tidy up as per suggestion.

Tidied up AudioFileInfo - removed old fileName in favour of Path - plus other recommendations

richardbull (approved):

PR #917, opened 2024-01-11
DMP-1856 - Mark audios to be deleted from Outbound data store [Task Automation]

HMCTSDeveloper:
https://tools.hmcts.net/jira/browse/DMP-1856

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/audio/service/impl/OutboundAudioDeleterProcessorImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/service/impl/OutboundAudioDeleterProcessorImpl.java
@@ -15,7 +15,6 @@ import uk.gov.hmcts.darts.common.entity.TransientObjectDirectoryEntity;
 import uk.gov.hmcts.darts.common.entity.UserAccountEntity;
 import uk.gov.hmcts.darts.common.exception.DartsApiException;
 import uk.gov.hmcts.darts.common.helper.SystemUserHelper;
-import uk.gov.hmcts.darts.common.repository.MediaRequestRepository;
 import uk.gov.hmcts.darts.common.repository.ObjectRecordStatusRepository;
 import uk.gov.hmcts.darts.common.repository.TransformedMediaRepository;
 import uk.gov.hmcts.darts.common.repository.TransientObjectDirectoryRepository;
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/hmcts/darts/audio/service/impl/OutboundAudioDeleterProcessorImpl.java
```diff
@@ -66,15 +72,36 @@ public List<MediaRequestEntity> markForDeletion() {
...
             entity.getTransformedMedia().getMediaRequest().setLastModifiedBy(systemUser);
 
             entity.setLastModifiedBy(systemUser);
             entity.setStatus(deletionStatus);
-
+            entity.getTransformedMedia().setExpiryTime(OffsetDateTime.now());
             deletedValues.add(entity.getTransformedMedia().getMediaRequest());
         }
 
+        for (Integer mediaRequestId : mediaRequests) {
```
cmagowan-hmcts, line 83: I'm thinking this class would be better to be rewritten, starting with line 51.
instead of getting a list of media requests that have transformed medias that are expired, i think we should be just getting a list of transformed media that are expired, then loop through that list.

using SQL that is very roughly somethgin like this:-

```
    @Query("""
        SELECT tm FROM MediaRequestEntity mr, TransformedMediaEntity tm
        WHERE tm.mediaRequest = mr
        AND ((tm.lastAccessed < :lastAccessedDateTime
        AND mr.status = "COMPLETED")
        OR (tm.createdDate < :lastAccessedDateTime
        AND mr.status <> "PROCESSING")
        AND tm.lastAccessed IS NULL)

        """)
```



The current implementation has me concerned that if a media request has 2 transformed media, and 1 of those was accessed today, and one hasn;t been accessed for a week, then this might deelte the one accessed recently also.

HMCTSDeveloper: ACCEPT

src/main/java/uk/gov/hmcts/darts/common/repository/TransformedMediaRepository.java
```diff
@@ -17,6 +18,13 @@ public interface TransformedMediaRepository extends JpaRepository<TransformedMed
         """)
     List<TransformedMediaEntity> findByMediaRequestId(Integer mediaRequestId);
 
+
+    @Query("""
+        SELECT tm FROM TransformedMediaEntity tm
+        WHERE tm.mediaRequest.id IN :ids
+        """)
+    List<TransformedMediaEntity> findByIdIn(List<Integer> ids);
```
cmagowan-hmcts, line 26: should probably call this `findByMediaRequestIdIn` or `findByMediaRequestIds`

HMCTSDeveloper: It's not being used so I will delete it.

cmagowan-hmcts (approved):

PR #920, opened 2024-01-12
DMP-1792: Lift-and-shift validation logic from TranscriptionController

chrisbellingham-hmcts:
# [DMP-1792](https://tools.hmcts.net/jira/browse/DMP-1792)

```
[ ] Yes
[X] No
```


tonymort (approved):
Regarding the BDD comments.  You think this something we should adopt?  I feel it would be an improvement as current readability of tests is low.  I've used this before but it tends to divide opinion.

chrisbellingham-hmcts:
> Regarding the BDD comments. You think this something we should adopt? I feel it would be an improvement as current readability of tests is low. I've used this before but it tends to divide opinion.

Personally I quite like it, as it does a better job of demarcating setup, invocation and assertion compared to just using whitespace.

Perhaps I can take the BDD approach here and we can use it as an example for further discussion in Best Practices?

src/test/java/uk/gov/hmcts/darts/transcriptions/component/impl/TranscriptionRequestDetailsValidatorTest.java
```diff
@@ -0,0 +1,237 @@
...
+        var hearingEntity = new HearingEntity();
+        hearingEntity.setMediaList(createMediaList());
+        Mockito.when(hearingsServiceMock.getHearingById(DUMMY_HEARING_ID))
+            .thenReturn(hearingEntity);
+
+        var requestDetails = createRequestDetails(MEDIA_1_START_TIME, MEDIA_1_END_TIME);
+        requestDetails.setTranscriptionTypeId(TranscriptionTypeEnum.SPECIFIED_TIMES.getId());
+        requestDetails.setTranscriptionUrgencyId(TranscriptionUrgencyEnum.STANDARD.getId());
+
+        // When
```
tonymort, line 183: Should this be then? Or maybe doesn't it does need it at all

tonymort (approved):

PR #930, opened 2024-01-16
DMP-2063 - v59 and v60 flyway changes

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-2063


src/main/resources/db/migration/common/V1_255_3_1_v59_v60_updates_user_account.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE user_account ADD COLUMN user_full_name  CHARACTER VARYING;
+update darts.user_account dl
```
tonymort, line 2: not doing any harm but dont think alias is needed

tonymort (approved):

src/main/resources/db/migration/common/V1_255_3_1_v59_v60_updates_user_account.sql
cmagowan-hmcts: fixed

PR #937, opened 2024-01-16
DMP-483 initial implementation

garethmorriscgi:
https://tools.hmcts.net/jira/browse/DMP-483 

### Change description ###
Link audio that has not already been linked to a hearing that is associated with hearings for event/case logs that occurred within a 30 minute window of the audio start and end time.

```
[ ] Yes
[X ] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/EventStub.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/EventStub.java
@@ -24,24 +24,16 @@ public class EventStub {
     }
 
     public EventEntity createEvent(HearingEntity hearing, int eventHandlerId) {
-        EventEntity eventEntity = new EventEntity();
-        eventEntity.setEventText("testEventText");
-        eventEntity.setEventName("testEventName");
-        EventHandlerEntity eventHandlerEntity = eventHandlerRepository.findById(eventHandlerId).get();
-        eventEntity.setEventType(eventHandlerEntity);
-        eventEntity.setTimestamp(OffsetDateTime.of(2020, 6, 20, 10, 0, 0, 0, ZoneOffset.UTC));
-        eventEntity.setCreatedBy(userAccountStub.getIntegrationTestUserAccountEntity());
-        eventEntity.addHearing(hearing);
-        eventEntity.setLastModifiedBy(userAccountStub.getIntegrationTestUserAccountEntity());
-        eventEntity.setLastModifiedDateTime(OffsetDateTime.of(2020, 6, 20, 10, 0, 0, 0, ZoneOffset.UTC));
-        eventRepository.saveAndFlush(eventEntity);
-        return eventEntity;
+        return createEvent(hearing,
+                    eventHandlerId,
+                    OffsetDateTime.of(2020, 6, 20, 10, 0, 0, 0, ZoneOffset.UTC),
+                    "testEventName");
     }
 
-    public EventEntity createEvent(HearingEntity hearing, int eventHandlerId, OffsetDateTime eventTimestamp) {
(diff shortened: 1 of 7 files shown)


src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/EventStub.java
```diff
@@ -38,4 +38,19 @@ public EventEntity createEvent(HearingEntity hearing, int eventHandlerId) {
         return eventEntity;
     }
 
+    public EventEntity createEvent(HearingEntity hearing, int eventHandlerId, OffsetDateTime eventTimestamp) {
```
johnrace0, line 41: maybe could combine this with existing createEvent method?

src/test/java/uk/gov/hmcts/darts/audio/service/impl/AudioServiceImplTest.java
```diff
@@ -218,6 +230,48 @@ void addAudio() throws IOException {
...
+        when(audioConfigurationProperties.getHandheldAudioCourtroomNumbers())
+            .thenReturn(Arrays.asList(addAudioMetadataRequest.getCourtroom()));
+
+        audioService.linkAudioToHearingByEvent(addAudioMetadataRequest, mediaEntity);
+        verify(hearingRepository, times(0)).saveAndFlush(any());
+        assertEquals(0, hearing.getMediaList().size());
+    }
+
+    @Test
+    void linkAudioToHearingByEvent() {
```
johnrace0, line 254: does this test cover pre-amble only? is it worth having a test for post-amble / both?

johnrace0 (changes requested):

src/main/resources/application.yaml
```diff
@@ -116,6 +116,10 @@ darts:
       - "mpeg2"
       - "mp2"
     max-file-size: 524288000
+    preAmbleSize: 30
+    postAmbleSize: 30
```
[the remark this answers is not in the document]
garethmorriscgi: Good suggestion - have updated the names

src/test/java/uk/gov/hmcts/darts/audio/service/impl/AudioServiceImplTest.java
garethmorriscgi: The response from searching for events is mocked so a post-amble test would in theory be a duplicate. I have however added an additional test to cover when the query brings back more than one court log for the same hearing to check that the audio is only linked once - and have used post amble times for the court log events.

src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/EventStub.java
garethmorriscgi: I've amended the method so the others call the newly created one

src/main/java/uk/gov/hmcts/darts/audio/config/AudioConfigurationProperties.java
```diff
@@ -34,5 +34,7 @@ public class AudioConfigurationProperties {
     private List<String> allowedMediaFormats = new ArrayList<>();
     private Integer maxFileSize;
     private Duration allowableAudioGapDuration;
```
richardbull, line 36: @garethmorriscgi This is an example of a `Duration` field in the config, like what @johnrace0 was mentioning earlier

johnrace0 (changes requested):
re: comment from RB about switching from Integer to Duration. is this something that can/will be done for this ticket?

garethmorriscgi:
> re: comment from RB about switching from Integer to Duration. is this something that can/will be done for this tic

The way that this configuration value is being used i.e. reducing/adding minutes to an OffsetDateTime which expects a long) I don't see the need to use the Duration type here. You'd need to do unnecessary additional work and lose code readability.

johnrace0 (approved):

PR #944, opened 2024-01-18
dmp 2017 v46 changes

chrisdavecm:
https://tools.hmcts.net/jira/browse/DMP-2017

### Change description ###

```
[ ] Yes
[x] No
```


jackmaloney (approved):

src/main/resources/db/migration/common/V1_258__v46_updates.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE event ADD COLUMN is_log_entry  BOOLEAN    NOT NULL  default false;
+
+ALTER TABLE external_object_directory ADD COLUMN external_file_id CHARACTER VARYING;
+ALTER TABLE external_object_directory ADD COLUMN external_record_id CHARACTER VARYING;
```
davet1985, line 4: These changes are on the Jira, are they intended to be in here?

jackmaloney: You don't need to make this change @chrisdavecm. I had this comment on the jira originally "review the external_file_id and external_record_id in external_object_directory table." but after a conversation with @hemantasharma1129 it was deemed not necessary to make that change.

hemantasharma1129: I had it removed because those 2 fields were blocking other tickets from development. They are now getting added part of dmp-1737 so please remove it from this PR. @chrisdavecm.

davet1985: @hemantasharma1129 This PR has been merged so would need to add another flyway migration removing the columns. @jackmaloney said they are required so can we leave this as it is and fix forward as necessary?

PR #948, opened 2024-01-19
DMP-2042 Daily list courtroom string fix

davet1985:

https://tools.hmcts.net/jira/browse/DMP-2042

### Change description ###

- the courtroom value coming in from the daily list is a string
- removing the cast
- adjusting testing

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListUpdater.java
```diff
@@ -68,7 +68,7 @@ public void processDailyList(DailyListEntity dailyListEntity) throws JsonProcess
                         String caseNumber = getCaseNumber(dailyListEntity, dailyListHearing);
 
                         HearingEntity hearing = retrieveCoreObjectService.retrieveOrCreateHearing(
-                            courtHouseName, String.valueOf(sitting.getCourtRoomNumber()),
```
chrisdavecm, line 71: Could that Integer come through as a null?

davet1985: Shouldn't do as it's required in the XSD.

chrisdavecm (approved):

PR #949, opened 2024-01-19
Dmp 2032 retention policies

chrisdavecm:
https://tools.hmcts.net/jira/browse/DMP-2032
https://tools.hmcts.net/jira/browse/DMP-2076
https://tools.hmcts.net/jira/browse/DMP-2098

### Change description ###
Same thing in both tickets

```
[ ] Yes
[x] No
```


src/main/resources/db/migration/common/V1_259__retention_policies.sql
```diff
@@ -0,0 +1,25 @@
+ALTER TABLE case_management_retention ADD COLUMN is_manual_override   BOOLEAN NOT NULL;
```
cmagowan-hmcts, line 1: not sure if its needed, as we may not have any data in the tables, but if introducing a "NOT NULL" column to an existing table and be careful and maybe add it as nullable, populate it with default values, then add the NOT NULL part.

cmagowan-hmcts (approved):

PR #956, opened 2024-01-22
DMP-1994 - Add GET /admin/security-roles

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-1994

and fied a flaky test from https://tools.hmcts.net/jira/browse/DMP-2068

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/usermanagement/controller/SecurityRoleControllerIntTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/usermanagement/controller/SecurityRoleControllerIntTest.java
@@ -18,7 +18,7 @@ import static org.springframework.test.web.servlet.request.MockMvcRequestBuilder
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
 @AutoConfigureMockMvc
-public class SecurityRoleControllerIntTest extends IntegrationBase {
+class SecurityRoleControllerIntTest extends IntegrationBase {
 
     public static final String ENDPOINT_URL = "/admin/security-roles";
 
--- a/src/integrationTest/java/uk/gov/hmcts/darts/usermanagement/controller/SecurityRoleControllerIntTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/usermanagement/controller/SecurityRoleControllerIntTest.java
@@ -11,9 +11,12 @@ import org.springframework.test.web.servlet.MockMvc;
 import org.springframework.test.web.servlet.MvcResult;
 import org.springframework.test.web.servlet.request.MockHttpServletRequestBuilder;
 import uk.gov.hmcts.darts.authorisation.component.UserIdentity;
+import uk.gov.hmcts.darts.common.entity.UserAccountEntity;
 import uk.gov.hmcts.darts.testutils.IntegrationBase;
 import uk.gov.hmcts.darts.testutils.stubs.AdminUserStub;
 
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.when;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
@@ -32,7 +35,7 @@ class SecurityRoleControllerIntTest extends IntegrationBase {
     private transient MockMvc mockMvc;
 
     @Test
-    void createSecurityGroupShouldSucceedWhenProvidedWithValidValuesForMinRequiredFields() throws Exception {
+    void getSecurityRolesShouldSucceedAndReturnAllRoles() throws Exception {
         adminUserStub.givenUserIsAuthorised(userIdentity);
 
         MockHttpServletRequestBuilder requestBuilder = get(ENDPOINT_URL)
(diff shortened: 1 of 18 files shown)


src/main/java/uk/gov/hmcts/darts/common/repository/SecurityRoleRepository.java
```diff
@@ -4,11 +4,14 @@
...
 
+import java.util.List;
 import java.util.Optional;
 
 @Repository
 public interface SecurityRoleRepository extends JpaRepository<SecurityRoleEntity, Integer> {
 
     Optional<SecurityRoleEntity> findByRoleName(String string);
 
+    List<SecurityRoleEntity> findAllByOrderById();
```
johnrace0, line 15: this name sounds weird - is it right?

src/integrationTest/java/uk/gov/hmcts/darts/usermanagement/controller/SecurityRoleControllerIntTest.java
```diff
@@ -0,0 +1,107 @@
...
+    void createSecurityGroupShouldSucceedWhenProvidedWithValidValuesForMinRequiredFields() throws Exception {
+        adminUserStub.givenUserIsAuthorised(userIdentity);
+
+        MockHttpServletRequestBuilder requestBuilder = get(ENDPOINT_URL)
+            .contentType(MediaType.APPLICATION_JSON_VALUE);
+
+        MvcResult mvcResult = mockMvc.perform(requestBuilder).andExpect(status().isOk()).andReturn();
+
+        String actualJson = mvcResult.getResponse().getContentAsString();
+        String expectedJson = """
```
johnrace0, line 44: is there a standard/recommendation for response strings in tests - a lot of the integration tests have them defined in an expectedResponse.json file?

src/main/java/uk/gov/hmcts/darts/common/repository/SecurityRoleRepository.java
cmagowan-hmcts: yes, its in the standard JPA format, if there was a mistake, it wouldn't work as the name infers the functionality.
See here for reference
https://stackoverflow.com/questions/25486583/how-to-use-orderby-with-findall-in-spring-data

src/integrationTest/java/uk/gov/hmcts/darts/usermanagement/controller/SecurityRoleControllerIntTest.java
cmagowan-hmcts: no standard, but there was a discussion last week on darts-modernisation, one person does suggest using there strings instead of response files:-
https://hmcts-reform.slack.com/archives/C04T0CVJ28N/p1705662974658419?thread_ts=1705658275.075549&cid=C04T0CVJ28N

src/integrationTest/java/uk/gov/hmcts/darts/usermanagement/controller/SecurityRoleControllerIntTest.java
```diff
@@ -0,0 +1,107 @@
...
+    private AdminUserStub adminUserStub;
+
+    @MockBean
+    private UserIdentity userIdentity;
+
+    @Autowired
+    private transient MockMvc mockMvc;
+
+    @Test
+    void createSecurityGroupShouldSucceedWhenProvidedWithValidValuesForMinRequiredFields() throws Exception {
```
chrisbellingham-hmcts, line 35: Update name to something like `getSecurityRolesShouldSucceedAndReturnAllRoles`

src/integrationTest/java/uk/gov/hmcts/darts/usermanagement/controller/SecurityRoleControllerIntTest.java
```diff
@@ -0,0 +1,107 @@
...
+import org.springframework.test.web.servlet.request.MockHttpServletRequestBuilder;
+import uk.gov.hmcts.darts.authorisation.component.UserIdentity;
+import uk.gov.hmcts.darts.testutils.IntegrationBase;
+import uk.gov.hmcts.darts.testutils.stubs.AdminUserStub;
+
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@AutoConfigureMockMvc
+class SecurityRoleControllerIntTest extends IntegrationBase {
```
chrisbellingham-hmcts, line 21: Could you add an int test to cover AC2 "I have obtained an access token as a non-ADMIN user, and the security_roles table is populated." please?

chrisbellingham-hmcts (changes requested):
Can you also add a functional test to cover the happy path (AC1) please? Whilst it will be similar to what's already covered in the int test, there is additional value in testing against Postgres and the non-mocked auth system in func test.

johnrace0 (approved):

src/integrationTest/java/uk/gov/hmcts/darts/usermanagement/controller/SecurityRoleControllerIntTest.java
cmagowan-hmcts: ok, although that would be testing more the Authorisation code and not my endpoint per-se. I.e. if authorisation throws a different HTTP response code, then its not really in the remit of this ticket to change that

chrisbellingham-hmcts: Purpose is to verify that the correct access rights are applied to the controller method via the `@Authorisation` annotation. I don't think we have any other mechanism to verify that only admins can access admin endpoints, on an endpoint-by-endpoint basis. If the `@Authorisation` annotation is ever accidently removed, then the test should catch that.

src/main/resources/openapi/usermanagement.yaml
```diff
@@ -111,7 +111,7 @@ paths:
                       - type
                     properties:
                       type:
-                        $ref: '#/components/schemas/UserAuthorisation403ErrorCode'
+                        $ref: '#/components/schemas/UserAuthorisation403107ErrorCode'
```
karen-hedges, line 114: Bit nik-picky but should you be blending 2 numbers to make 1, its a bit confusing?

cmagowan-hmcts: changed

karen-hedges (approved):

PR #965, opened 2024-01-25
DMP-2049 - API - create new GET /audios/{media_id}/preview endpoint to use Server-Sent events

HMCTSDeveloper:
https://tools.hmcts.net/jira/browse/DMP-2049

Palo Alto firewall is killing long term connections which are not sending data. Any connection to the server that does not send data back will be closed. The work around for this is to use send server events(mime type even/stream) which is a uni directional connection where the server can send data without closing the connection. The alternative api will send heartbeat data and preview is ready. 

Unfortunately swagger does not allow generation off SSE endpoints, thus this api is not created using swagger.


src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
```diff
@@ -50,6 +59,14 @@ public ResponseEntity<List<AudioMetadata>> getAudioMetadata(Integer hearingId) {
         return new ResponseEntity<>(audioMetadata, HttpStatus.OK);
     }
 
+
+    @Override
+    @SecurityRequirement(name = SECURITY_SCHEMES_BEARER_AUTH)
+    public ResponseEntity<Void> addAudio(MultipartFile file, AddAudioMetadataRequest metadata) {
```
cmagowan-hmcts, line 65: is this suppsoed to be added as part of this ticket?

src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
```diff
@@ -61,11 +78,28 @@ public ResponseEntity<byte[]> preview(Integer mediaId, String httpRangeList) {
...
+    public Flux<ResponseEntity<byte[]>> previewAlternative(
+        @Parameter(name = "media_id", description = "Internal identifier for media", required = true, in = ParameterIn.PATH)
+        @PathVariable("media_id") Integer mediaId,
+        @Parameter(name = "range", description = "Range header, required for streaming audio.", in = ParameterIn.HEADER)
+        @RequestHeader(value = "range", required = false) String range
+    ) {
+        Flux<ResponseEntity<byte[]>> work = audioService.getAudioPreviewFlux(mediaId, range);
+
+        Flux<ResponseEntity<byte[]>> heartBeat = Flux.interval(Duration.ofSeconds(5))
+            .map(sequence -> new ResponseEntity<byte[]>(HttpStatus.CONTINUE)).takeUntilOther(work);
```
mestebanez, line 101: Does this break the HTTP semantics of continue. HTTP is request response and continue has a specific meaning. This seems like a bespoke scenario for the client connection to stay open?

src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
```diff
@@ -61,11 +78,28 @@ public ResponseEntity<byte[]> preview(Integer mediaId, String httpRangeList) {
...
+    )
+    @SneakyThrows
     @SecurityRequirement(name = SECURITY_SCHEMES_BEARER_AUTH)
-    public ResponseEntity<Void> addAudio(MultipartFile file, AddAudioMetadataRequest metadata) {
-        audioService.addAudio(file, metadata);
-        return new ResponseEntity<>(HttpStatus.OK);
+    @Authorisation(contextId = MEDIA_ID,
+        securityRoles = {JUDGE, REQUESTER, APPROVER, TRANSCRIBER, TRANSLATION_QA, RCJ_APPEALS},
+        globalAccessSecurityRoles = {JUDGE})
+    public Flux<ResponseEntity<byte[]>> previewAlternative(
```
mestebanez, line 92: Flux will allow us to manage back pressure by putting everything on a single event loop.  This is probably a good idea in order to scale out (and avoid denial of service) but will quickly get to a point whereby we are synchronously processing each request depending on the backend thread pool. Just calling it out.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioServiceImpl.java
```diff
@@ -105,6 +109,20 @@ public InputStream preview(Integer mediaId) {
         return mediaBinaryData.toStream();
     }
 
+    @Override
+    public Flux<ResponseEntity<byte[]>> getAudioPreviewFlux(Integer mediaId, String range) {
```
cmagowan-hmcts, line 113: what subject is this publishing to, as it will need to match what the front-end is looking out for

src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
mestebanez: An alternate would be to add some asynchronous endpoints so that clients can check in (poll) on the progress of the backend operations. That way client connections do not need to be maintained, plus clients are less likely to wait an ever increasing amount of time as more clients wait for a response.

cmagowan-hmcts: that was covered as part of the spike and they didn't want to do that

mestebanez, line 92: Also sorry I'm naive to the specific implemenation of flux. Spring provides a flux RestController annotation as part of the dependnecy spring-boot-starter-webflux. Do we not need that instead of the MVC one?

mestebanez: Did we try using https://www.baeldung.com/spring-deferred-result? DId that help with the timeout problem. Conscious that this allows us to come away from a react approach

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioServiceImpl.java
```diff
@@ -105,6 +109,20 @@ public InputStream preview(Integer mediaId) {
...
+        return Flux.just(mediaId)
+            .publishOn(Schedulers.boundedElastic()).map(m -> {
+                InputStream audioMediaFile = preview(m);
+                try {
```
mestebanez, line 117: put the input stream inside the try block to get the IS close for free

src/main/java/uk/gov/hmcts/darts/audio/service/AudioService.java
```diff
@@ -13,6 +15,8 @@ public interface AudioService {
 
     InputStream preview(Integer mediaId);
 
+    Flux<ResponseEntity<byte[]>> getAudioPreviewFlux(Integer mediaId, String range);
```
mestebanez, line 18: May be worth keep the flux calls in a seperate place/package just to allow them to be easily identifiable

src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioControllerPreviewWithSendServerEventsIntTest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Test
+    void previewShouldReturnSuccess() {
+        FluxExchangeResult<String> result = webClient.get().uri(uriBuilder -> uriBuilder.path(
+                URL + mediaEntity.getId()).build())
+            .header(RANGE, RANGE_1024)
+            .accept(TEXT_EVENT_STREAM)
+            .exchange()
+            .expectStatus().isOk().returnResult(String.class);
```
mestebanez, line 77: Can we assert the interval events coming through and also the file contents.

src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
HMCTSDeveloper: Looking at the definition of it I think matches our usage - "The HTTP 100 Continue informational status response code indicates that everything so far is OK and that the client should continue with the request or ignore it if it is already finished."

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioServiceImpl.java
HMCTSDeveloper: ACCEPT

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioServiceImpl.java
HMCTSDeveloper: I have changed it to SSE events as the front end is looking at the event with name "heartbeat" to determine.

src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
HMCTSDeveloper: > Also sorry I'm naive to the specific implemenation of flux. Spring provides a flux RestController annotation as part of the dependnecy spring-boot-starter-webflux. Do we not need that instead of the MVC one?

Can you show where the Webflux rest controller is located I can't seem to find it. I also suspect in that case the annotation would apply to the whole controller which is what we don't want.

HMCTSDeveloper: > Did we try using https://www.baeldung.com/spring-deferred-result? DId that help with the timeout problem. Conscious that this allows us to come away from a react approach

I don't think this will help as we not looking for a non blocking call but a call that will constantly return data to stop the firewall from closing the connection.

src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioControllerPreviewWithSendServerEventsIntTest.java
HMCTSDeveloper: ACCEPT

src/main/java/uk/gov/hmcts/darts/audio/service/AudioService.java
HMCTSDeveloper: I think it's okay here as it's still related to the audio service.

src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
mestebanez: ok i guess that has to be a limitation with any solution we adopt then. If we use an event based SSE approach to keep the connection open then any similar mechanism is going to get to a point whereby we exhaust the download requests on the container.

mestebanez: I did read about web flux schedulers to avoid such a situation https://www.baeldung.com/spring-webflux-concurrency

src/main/java/uk/gov/hmcts/darts/audio/service/AudioService.java
mestebanez: yeah agreed. just a suggestion

src/main/java/uk/gov/hmcts/darts/audio/controller/AudioController.java
```diff
@@ -61,11 +77,22 @@ public ResponseEntity<byte[]> preview(Integer mediaId, String httpRangeList) {
...
+    @Authorisation(contextId = MEDIA_ID,
+        securityRoles = {JUDGE, REQUESTER, APPROVER, TRANSCRIBER, TRANSLATION_QA, RCJ_APPEALS},
+        globalAccessSecurityRoles = {JUDGE})
+    public SseEmitter previewAlternative(
+        @Parameter(name = "media_id", description = "Internal identifier for media", required = true, in = ParameterIn.PATH)
+        @PathVariable("media_id") Integer mediaId,
+        @Parameter(name = "range", description = "Range header, required for streaming audio.", in = ParameterIn.HEADER)
+        @RequestHeader(value = "range", required = false) String range
+    ) {
+        return audioService.startStreamingPreview(mediaId, range, new SseEmitter(AudioServiceImpl.EMITTER_TIMEOUT));
```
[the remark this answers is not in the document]
HMCTSDeveloper: ACCEPT

cmagowan-hmcts (approved):

src/main/java/uk/gov/hmcts/darts/common/sse/SentServerEventsHeartBeatEmitter.java
```diff
@@ -0,0 +1,64 @@
...
+        } catch (IOException e) {
+            DartsApiException dartsApiException = new DartsApiException(AudioApiError.FAILED_TO_PROCESS_AUDIO_REQUEST);
+            emitter.completeWithError(dartsApiException);
+            log.error("Error while emitting heartbeat", e);
+            throw dartsApiException;
+        }
+    }
+
+    public void startHeartBeat(SseEmitter emitter) {
+        ExecutorService emitterExecutor = Executors.newSingleThreadExecutor();
```
mestebanez, line 52: Not something for you to fix but at some point we may want to use a pool here. The way this currently works is that if we had enough calls coming through we could blow the whole VM.

src/main/resources/application.yaml
```diff
@@ -220,11 +220,14 @@ darts:
...
     application-server: ${CLUSTER_NAME:not available}
+  sse:
+    heartbeat: 5 # in seconds
+    sse-preview-timeout: 120 # in seconds
 
 dbMigration:
   # When true, the app will run DB migration on startup.
   # Otherwise, it will just check if all migrations have been applied (and fail to start if not).
-  runOnStartup: ${RUN_DB_MIGRATION_ON_STARTUP:true}
+  runOnStartup: ${RUN_DB_MIGRATION_ON_STARTUP:true}z
```
mestebanez, line 230: Should there be a z at the end here

src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/DataManagementServiceStubImpl.java
```diff
@@ -21,12 +21,18 @@
...
 @RequiredArgsConstructor
 @Profile("intTest")
+@SuppressWarnings("PMD.AvoidThrowingRawExceptionTypes")
 public class DataManagementServiceStubImpl implements DataManagementService {
 
     @Override
     public BinaryData getBlobData(String containerName, UUID blobId) {
         logStubUsageWarning();
 
+        try {
```
mestebanez, line 31: not sure whats going on here?

src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioControllerPreviewWithSendServerEventsIntTest.java
```diff
@@ -0,0 +1,132 @@
...
+        );
+        webClient = webClient.mutate()
+            .responseTimeout(Duration.ofMillis(TIMEOUT))
+            .build();
+
+    }
+
+    @Test
+    void previewShouldReturnSuccess() {
+        FluxExchangeResult<ServerSentEvent<String>> result = webClient.get().uri(uriBuilder -> uriBuilder.path(
```
mestebanez, line 83: Flux usgae. Do we need to remove this?

PR #970, opened 2024-01-25
Dmp 2040 add viq header in audio files

mario-paniccia:


src/main/java/uk/gov/hmcts/darts/audio/model/ViqHeader.java
```diff
@@ -0,0 +1,53 @@
...
+    private static final int HOUR_OFFSET = 42;
+    private static final int SECONDS_OFFSET = 44;
+
+    byte[] viqHeader;
+
+    public ViqHeader(Instant instant) {
+
+        viqHeader = new byte[VIQ_HEADER_SIZE];
+        viqHeader[DAY_OFFSET] = (byte) instant.atOffset(ZoneOffset.UTC).getDayOfMonth();
+        viqHeader[MONTH_OFFSET] = (byte) (instant.atOffset(ZoneOffset.UTC).getMonthValue() + 1);
```
richardbull, line 41: Please can you check this value - it looks wrong to me, no need to add 1 to the month here?
`Gets the month-of-year field from 1 to 12`

I think it was being done in legacy because of using the old `Calendar` code

src/main/java/uk/gov/hmcts/darts/audio/model/AudioFileInfo.java
```diff
@@ -23,5 +23,6 @@ public class AudioFileInfo {
     private Integer channel;
     @NotNull
     private Path path;
+    private boolean isTrimmed;
```
richardbull, line 26: I'm not sure about the need to add this here, I think the ViqHeader needs adding to all the files in the download zip.

This only works currently because all the files go through `trimAllToPeriod:237, OutboundFileProcessorImpl`.

Really, I think we should only need to trim the first/last audio session - so this is an existing issue, but if we change that then I think this will mess up the Viq files if they are missing the header bytes.

I think @cmagowan-hmcts is the only one that has tried to play something in a Viq Player.

src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileZipGeneratorHelperImpl.java
```diff
@@ -143,6 +146,26 @@ public String generateReadme(ViqMetaData viqMetaData, String fileLocation) {
         return readmeFile.getAbsolutePath();
     }
 
+    @Override
+    public Path generateViqFile(AudioFileInfo audioFileInfo, Path viqOutputFile) {
+
+        if (!audioFileInfo.isTrimmed()) {
```
richardbull, line 152: I don't think we need this bit tbh - what do you think @hemantasharma1129 ?

src/main/java/uk/gov/hmcts/darts/audio/model/ViqHeader.java
mario-paniccia: you're totally right, nice catch Richard. Will fix this

src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileZipGeneratorHelperImpl.java
mario-paniccia: the ticket requirement was to add the header just for trimmed files. Let's see if Hemanta confirms, if not I'll change this. Cheers

hemantasharma1129: Its only for the Audio we manipulated i.e. trimmed.

richardbull: Which is all the files currently :)

src/main/java/uk/gov/hmcts/darts/audio/model/ViqHeader.java
mario-paniccia: done

richardbull (approved):
I've checked a DOWNLOAD request locally and can confirm that the zip structure looks as expected for the audio sessions along with the playlist.xml now contains the full path to the item with back slashes for Windows e.g. `daudio\localaudio\T2023\1009-1\0001\`

mario-paniccia:
> I've checked a DOWNLOAD request locally and can confirm that the zip structure looks as expected for the audio sessions along with the playlist.xml now contains the full path to the item with back slashes for Windows e.g. `daudio\localaudio\T2023\1009-1\0001\`

thanks @richardbull

hemantasharma1129 (approved):

PR #974, opened 2024-01-25
DMP-2151: Implement ArmTokenClient

chrisbellingham-hmcts:
# [DMP-2151](https://tools.hmcts.net/jira/browse/DMP-2151)

Implement API client to obtain an access token for ARM.

KeyVault entries for `ArmTokenBaseUrl` have been added in Azure for `stg` and `demo`.

Associated wiremock stub is located here: https://github.com/hmcts/darts-stub-services/pull/71

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/arm/client/ArmClientConfig.java
+++ b/src/main/java/uk/gov/hmcts/darts/arm/client/ArmClientConfig.java
@@ -1,0 +1,36 @@
+package uk.gov.hmcts.darts.arm.client;
+
+import feign.RequestTemplate;
+import feign.codec.EncodeException;
+import feign.codec.Encoder;
+import groovy.util.logging.Slf4j;
+import org.springframework.context.annotation.Bean;
+import uk.gov.hmcts.darts.arm.client.model.ArmTokenRequest;
+
+@Slf4j
+public class ArmClientConfig {
+
+    @Bean
+    public Encoder armTokenClientEncoder() {
+        // Normally SpringFormEncoder would be used to encode form data, but this doesn't support the "text/plain" Content-Type required by ARM API. Therefore,
+        // we must implement our own encoder here.
+        return (o, type, requestTemplate) -> {
+            if (o instanceof ArmTokenRequest armTokenRequest) {
+                encodeArmTokenRequest(armTokenRequest, requestTemplate);
+            } else {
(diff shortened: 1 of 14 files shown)


hemantasharma1129 (approved):

src/main/resources/application-local.yaml
```diff
@@ -25,6 +25,9 @@ darts:
     enabled: ${TESTING_SUPPORT_ENDPOINTS_ENABLED:true}
   redis:
     ssl-enabled: false
+  storage:
+    arm:
+      token-base-url: ${ARM_TOKEN_BASE_URL:localhost:4551}
```
richardbull, line 30: Is this missing the protocol prefix - `http://`?

src/main/java/uk/gov/hmcts/darts/arm/client/ArmClientConfig.java
```diff
@@ -0,0 +1,36 @@
...
+
+import feign.RequestTemplate;
+import feign.codec.EncodeException;
+import feign.codec.Encoder;
+import groovy.util.logging.Slf4j;
+import org.springframework.context.annotation.Bean;
+import uk.gov.hmcts.darts.arm.client.model.ArmTokenRequest;
+
+@Slf4j
+public class ArmClientConfig {
```
richardbull, line 11: Is this class missing the spring `@Configuration`, since it uses the `@Bean` annotations below?
Also wondering, should it be under a `config` sub package as we normally like doing that

chrisbellingham-hmcts: Not in this case - adding `@Configuration` would make this Encoder configration apply to all Feign clients, whereas here we only wish to override the Encoder behaviour for the Arm Token client only.

src/main/resources/application-local.yaml
chrisbellingham-hmcts: I don't think the protocol is required here, have tested locally with no issues.

src/main/java/uk/gov/hmcts/darts/arm/client/ArmClientConfig.java
chrisbellingham-hmcts: Happy to move into a `config` dir.

PR #997, opened 2024-01-30
DMP-1912 ARM integrate Disposition endpoint

richardbull:

[DMP-1912](https://tools.hmcts.net/jira/browse/DMP-1912)

### Change description ###

* Add ArmApiServiceImpl#updateMetadata
    * Remove `darts.storage.arm.token-base-url` - replaced by `darts.storage.arm-api.url`
* Update ArmApiServiceIntTest to mock the client calls as opposed to using the darts-stub-services
    * staging darts-stub-services doesn't seem to be very stable?
* Add Arm secrets / Environment Variables

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/arm/config/ArmDataManagementConfiguration.java
```diff
@@ -3,9 +3,11 @@
 import lombok.Getter;
 import lombok.Setter;
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Configuration;
 
 @Configuration
+@EnableConfigurationProperties(ArmApiConfigurationProperties.class)
```
mario-paniccia, line 10: from my knowledge this is needed only if you didn't have `@Configuration` already

mario-paniccia (approved):
LGTM, just one minor comment

src/main/java/uk/gov/hmcts/darts/arm/api/impl/ArmDataManagementApiImpl.java
```diff
@@ -43,6 +47,11 @@ public void deleteResponseBlob(String blobName) {
         armService.deleteResponseBlob(getArmContainerName(), blobName);
     }
 
+    @Override
+    public UpdateMetadataResponse updateMetadata(String externalRecordId, OffsetDateTime eventTimestamp) {
+        return armApiService.updateMetadata(externalRecordId, eventTimestamp);
```
chrisbellingham-hmcts, line 52: I'm not quite sure what this part of the Jira ticket means:

> Input parameter will be RecordId from External Object Directory table and calculated event_date. For now pass the event date as 7 years from current Date. There will be a separate ticket ([DMP-2057](https://tools.hmcts.net/jira/browse/DMP-2057)) to calculate event_date as it depends upon Retention Service.

Does this mean we should be ignoring the `eventTypestamp` param here temporarily, and evaulating the second arg to `updateMetadata` as current time + 7 years, or are we ok to just accept the value from the caller?

src/main/java/uk/gov/hmcts/darts/arm/client/ArmApiClient.java
```diff
@@ -0,0 +1,27 @@
...
+import uk.gov.hmcts.darts.arm.client.model.UpdateMetadataResponse;
+
+import static org.springframework.http.HttpHeaders.AUTHORIZATION;
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+
+@FeignClient(
+    name = "arm-api-client",
+    url = "${darts.storage.arm-api.url}"
+)
+public interface ArmApiClient {
```
chrisbellingham-hmcts, line 18: Consider writing a dedicated integration test to cover the client functionality, using embedded Wiremock to verify the request similar to `ArmTokenClientIntTest`. The added value is to ensure the client is making the request in the expected way, in this case a `POST` with a bearer token in the header, and a JSON request body.

When I was writing the token client I found this test useful because Feign client was silently making unwanted changes to the request, which prompted some necessary workarounds.

src/main/java/uk/gov/hmcts/darts/arm/api/impl/ArmDataManagementApiImpl.java
richardbull: I was hoping it would be ok to accept the date/time from the caller as we have here - I was confused myself and thought I needed to plug it into maybe `ArmResponseFilesProcessorImpl` - but think @hemantasharma1129 said that it will be included in another area of retention that is already ticketed. So will leave it for Hemanta to call out here.

src/main/java/uk/gov/hmcts/darts/arm/client/ArmApiClient.java
richardbull: Yep, I spotted you were doing something with wiremock in darts-api.
I was initially wanting to check against our darts-stub-services on staging to check I got the OK response ok actually, but I found that it was giving me Error 500 more often than not yesterday, so the deployment there doesn't feel too stable for some reason. Therefore I just ended up mocking the clients and checking the parameters.

richardbull: Done in commit 4bb0083d82cade406a78ca36c0948fb6032cd80f

src/main/resources/application-local.yaml
```diff
@@ -26,8 +26,8 @@ darts:
   redis:
     ssl-enabled: false
   storage:
-    arm:
-      token-base-url: ${ARM_TOKEN_BASE_URL:localhost:4551}
+    arm-api:
+      url: ${ARM_URL:http://localhost:4551}
```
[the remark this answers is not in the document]
richardbull: Done in commit 4bb0083d82cade406a78ca36c0948fb6032cd80f

src/main/java/uk/gov/hmcts/darts/arm/config/ArmDataManagementConfiguration.java
richardbull: I've left this as removing it here in our version of spring boot will currently fail all the integration tests e.g.

`Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'armApiServiceImpl' defined in file [ArmApiServiceImpl.class]: Unsatisfied dependency expressed through constructor parameter 0: No qualifying bean of type 'uk.gov.hmcts.darts.arm.config.ArmApiConfigurationProperties' available: expected at least 1 bean which qualifies as autowire candidate.`

PR #1009, opened 2024-02-01
DMP - 2039 api problem responses

JasonPittockCGI:

cd /tmp
### Change description ###

```
[ ] Yes
[ x] No
```


src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioControllerAddAudioMetadataIntTest.java
```diff
@@ -146,7 +146,7 @@ void addAudioMetadataNonExistingCourthouse() throws Exception {
 
         String actualJson = mvcResult.getResponse().getContentAsString();
         String expectedJson = """
-            {"type":"COMMON_100","title":"Provided courthouse does not exist","status":400,"detail":"Courthouse 'TEST' not found."}""";
+            {"type":"100","title":"Provided courthouse does not exist","status":400,"detail":"Courthouse 'TEST' not found."}""";
```
mestebanez, line 149: Just a suggestion but maybe less fragile to use the enumeration here, just incase the code changes in future. Same pattern would be needed for all integration tests.

src/main/resources/openapi/hearings.yaml
```diff
@@ -244,3 +244,15 @@ components:
       enum:
         - "AUTHORISATION_100"
       x-enum-varnames: [ USER_NOT_AUTHORISED_FOR_COURTHOUSE ]
+
+    HearingsErrorCode:
+      type: string
+      enum:
+        - "100"
```
mestebanez, line 251: Would need an op name prefixing them with the api name i.e. something like HEARINGS_100 to ensure system wide uniqueness in the error code

src/main/resources/openapi/retentions.yaml
```diff
@@ -10,7 +10,6 @@ info:
 servers:
   - url: http://localhost:4550/
 
-
```
mestebanez, line 13: ?

src/main/resources/openapi/dailylist.yaml
```diff
@@ -472,3 +472,24 @@ components:
...
+     type: string
+     enum:
+      - "100"
+      - "101"
+      - "102"
+      - "103"
+      - "104"
+     x-enum-varnames: [ FAILED_TO_PROCESS_DAILYLIST, XML_OR_JSON_NEEDS_TO_BE_PROVIDED, XML_EXTRA_PARAMETERS_MISSING, DAILY_LIST_NOT_FOUND, INTERNAL_ERROR ]
+
+    DailyListTitleErrors:
```
mestebanez, line 486: didn't expect titles to be here but kind of works fine :)

src/main/resources/openapi/hearings.yaml
mestebanez: This would apply across all of the yaml files.

chrisbellingham-hmcts: Agree, the intent of the `type` field in the JSON response is to uniquely identify the error across the application.

It's hard to ensure a purely numeric value will be unique, because the enum definitons are spread across multiple openapi specs. Whereas adding a prefix (aligned with the feature name, e.g. HEARINGS) effectively namespaces the numeric values, minimising the chances of clashes with values used in other specs.

src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioControllerAddAudioMetadataIntTest.java
```diff
@@ -146,7 +146,7 @@ void addAudioMetadataNonExistingCourthouse() throws Exception {
 
         String actualJson = mvcResult.getResponse().getContentAsString();
         String expectedJson = """
-            {"type":"COMMON_100","title":"Provided courthouse does not exist","status":400,"detail":"Courthouse 'TEST' not found."}""";
```
chrisbellingham-hmcts, line 149: The intent of the `type` field is really to allow consumers to identify specific errors, and potentially handle them. Are we confident that none of our API consumers (frontend) are reliant on the value of the `type` field for error handling? I think one of the ACs on DMP-2039 should be that none of the responses are impacted, i.e this is a refactoring change rather than a functional change. What's your view @mestebanez?

src/main/java/uk/gov/hmcts/darts/common/exception/AuditApiError.java
```diff
@@ -3,25 +3,27 @@
...
+import uk.gov.hmcts.darts.audit.model.AuditTitleErrors;
 
 @Getter
 @RequiredArgsConstructor
 public enum AuditApiError implements DartsApiError {
 
     FILTERS_WERE_EMPTY(
-        "100",
+        AuditErrorCode.AUDIT_FILTERS_WERE_EMPTY.getValue(),
         HttpStatus.BAD_REQUEST,
```
chrisbellingham-hmcts, line 15: As the intent of the refactoring is to define the error details in openapi such that it's self-documenting, should we also pull the desired HTTP status from there, rather than hardcoding it here?

This is a general comment applicable to all implementations of `DartsApiError`

src/main/resources/openapi/hearings.yaml
JasonPittockCGI: This was always my concern that the consumer would lose info. Happy to go back and refactor to include the prefix in the open api yaml.

src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioControllerAddAudioMetadataIntTest.java
JasonPittockCGI: Yeah. Agree it could be improved. Ideally we should create the expectedJson with a builder and serialize to json. Perhaps we could do this on another housekeeping ticket as we would have to do it consistently across the board?

mestebanez (commented):
Not essential but I wonder wether its worth adding a generic test to assert that:-

1) all of our enum entries that extend DartsApiError return a type string matching the regex:-

_.*

2) that we dont have any duplicate numbers across enumerations. This would guard against regression and enforce that any new errors are correctly defined, else the test fails.

mestebanez (approved):

PR #1012, opened 2024-02-02
Fix to lazy eod

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-1737

### Change description ###

Fix of lazy loaded object

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmResponseFilesProcessorImpl.java
```diff
@@ -100,6 +100,11 @@ public void processResponseFiles() {
         }
     }
 
+    @Transactional
```
chrisbellingham-hmcts, line 103: I was of the understanding that `@Transactional` has no effect on private methods. Are you sure this fixes the issue?

karen-hedges: Well spotted

karen-hedges:
Gone for a different approach

PR #1016, opened 2024-02-02
Dmp 1737.transactional fix

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-1737

### Change description ###

Fix issue with lazily loaded objects

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmResponseFilesProcessorImpl.java
```diff
@@ -111,6 +108,18 @@ private void initialisePreloadedObjects() {
         userAccount = userIdentity.getUserAccount();
     }
 
+    @SuppressWarnings("java:S3655")
+    @Transactional
+    public void processInputUploadFile(Integer eodId) {
```
tonymort, line 113: Transactional methods need to be called externally from the class as it only triggers the transactional aspect when it hits the proxy object

hemantasharma1129 (approved):

PR #1023, opened 2024-02-05
Remove overly restrictive regex constraint from usermanagement openapi

chrisbellingham-hmcts:
Regex for username field of user search is overly-restrictive at this time. A separate excerise will review all applied openapi constraints and tighen/loosen as necessary, with steer from migrated legacy data.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/openapi/usermanagement.yaml
+++ b/src/main/resources/openapi/usermanagement.yaml
@@ -316,7 +316,6 @@ components:
     UserName:
       minLength: 1
       maxLength: 256
-      pattern: ^\S$|^\S[a-zA-Z '-]*\S$
       type: string
     UserEmailAddress:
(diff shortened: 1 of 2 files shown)


davet1985 (approved):

src/main/resources/openapi/usermanagement.yaml
```diff
@@ -316,7 +316,6 @@ components:
     UserName:
```
richardbull, line 316: Does this need to become `UserFullName` to avoid any confusion with the old `user_name` column?

chrisbellingham-hmcts: We have some FE tickets in progress that would be impacted by changing the attribute name, will consider this as an improvement.

PR #1025, opened 2024-02-05
DMP-2166 ATS DOWNLOAD trim start and end of sessions (if applicable)

richardbull:

[DMP-2166](https://tools.hmcts.net/jira/browse/DMP-2166)

### Change description ###

Update processAudioForDownload to trim start and end of sessions (if applicable)

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImplTest.java
```diff
@@ -61,65 +67,113 @@ void setUp() {
...
+                Duration.ofMinutes(9)))
+                .thenReturn(trimmedAudioFileInfo);
+
+        // When
         List<List<AudioFileInfo>> sessions = outboundFileProcessor.processAudioForDownload(
-            mediaEntityToDownloadLocation,
-            TIME_12_00,
-            TIME_12_10
+                mediaEntityToDownloadLocation,
+                TIME_12_02,
```
johnrace0, line 109: just wondering why checkstyle doesn't pick up on all these indentation variations?

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioServiceImpl.java
```diff
@@ -73,23 +74,23 @@ public class AudioServiceImpl implements AudioService {
...
 
-    private static AudioFileInfo createAudioFileInfo(MediaEntity mediaEntity, Path downloadPath) {
-        return new AudioFileInfo(
-            mediaEntity.getStart().toInstant(),
-            mediaEntity.getEnd().toInstant(),
-            mediaEntity.getChannel(),
-            downloadPath,
-            false
-        );
+    private AudioFileInfo createAudioFileInfo(MediaEntity mediaEntity, Path downloadPath) {
```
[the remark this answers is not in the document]
richardbull: It's a private method that is only used by `preview` in this class.
It's only got one instance of the service injected by spring in `AudioController`, so I didn't see the point of it being static.  I assume it was an intellij refactor thing, bit like the constants in the same class that end up being public each time and people forgetting to change them to private.

src/test/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImplTest.java
richardbull: No idea - Pass.  Maybe it's not a standard checkstyle concern.
We have a simple https://github.com/hmcts/darts-api/blob/master/.editorconfig
But I believe we are supposed to be using `GoogleStyle` Code Style - hopefully mine hasn't reverted to something else as I did upgrade my IntelliJ version recently

src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
```diff
@@ -205,33 +217,41 @@ private AudioFileInfo merge(List<AudioFileInfo> audioFileInfos)
...
+                                        OffsetDateTime mediaRequestStartTime,
+                                        OffsetDateTime mediaRequestEndTime)
         throws ExecutionException, InterruptedException, IOException {
         List<AudioFileInfo> processedAudios = new ArrayList<>();
         for (AudioFileInfo audioFileInfo : audioFileInfos) {
-            AudioFileInfo trimmedAudio = trimToPeriod(audioFileInfo, start, end);
-            processedAudios.add(trimmedAudio);
+            final boolean audioStartsBeforeRequest = audioFileInfo.getStartTime().isBefore(mediaRequestStartTime.toInstant());
+            final boolean audioEndsAfterRequest = audioFileInfo.getEndTime().isAfter(mediaRequestEndTime.toInstant());
+            if (audioStartsBeforeRequest || audioEndsAfterRequest) {
```
mario-paniccia, line 246: I might be wrong. But if we are doing this check here, isn't then the below code we've introduced redundant:

```
if (numberOfSessions > 2) {
            finalisedAudioSessions.addAll(groupedAudioSessions.subList(1, numberOfSessions - 1)); // no need to trim mid session
        }
```

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioOperationServiceImpl.java
```diff
@@ -107,14 +106,14 @@ public List<AudioFileInfo> concatenateWithGaps(final String workspaceDir, final
...
-                outputPath,
-                false
+            audioFileInfoList.add(
+                AudioFileInfo.builder()
+                    .startTime(getEarliestStartTime(seperatedAudioFileInfo))
+                    .endTime(getLatestEndTime(seperatedAudioFileInfo))
+                    .channel(channel)
+                    .path(outputPath)
+                    .build()
             );
```
mario-paniccia, line 116: nice idea introducing a Builder 👍

mario-paniccia (approved):
approved, just a minor comment

src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
richardbull: I'd prefer that we steer away from doing any trim operation at all for the mid sessions as that may lead to erroneous files perhaps if it's understood that the audio files should not need to be touched by ffmpeg - they will already contain the ViqHeader for example, so I think they are best left untouched. It might also improve performance if there are a lot of mid session files to not need to process.

PR #1026, opened 2024-02-06
DMP-2156 Preview Audio Unavailable

johnrace0:
https://tools.hmcts.net/jira/browse/DMP-2156

### Change description ###
Preview audio is unavailable if file is only in long term storage ARM

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/darts/audio/component/AudioRequestBeingProcessedFromArchiveQuery.java
```diff
@@ -8,4 +8,5 @@ public interface AudioRequestBeingProcessedFromArchiveQuery {
 
     List<AudioRequestBeingProcessedFromArchiveQueryResult> getResults(Integer mediaRequestId);
 
+    List<AudioRequestBeingProcessedFromArchiveQueryResult> getResultsByMediaIds(List<Integer> mediaIds);
```
richardbull, line 11: I think this should be from `hearingId` in a separate query component?

src/main/java/uk/gov/hmcts/darts/audio/component/impl/AudioRequestBeingProcessedFromArchiveQueryImpl.java
```diff
@@ -73,4 +76,24 @@ public List<AudioRequestBeingProcessedFromArchiveQueryResult> getResults(Integer
         );
     }
 
+    @Override
+    public List<AudioRequestBeingProcessedFromArchiveQueryResult> getResultsByMediaIds(List<Integer> mediaIds) {
+        return jdbcTemplate.query(
+            AUDIO_REQUEST_QUERY +
```
richardbull, line 82: I don't think you can use the same base query for yours as it's possible that no audio requests has happened yet for that hearing-media.
We have the `hearing_id` that can be used for the WHERE clause.
Then I think the FROM can start from `hearing` rather than `media_request` table.


**GET /audio/hearings/{hearing_id}/audios**

I think you probably want a separate `MediaBeingProcessedFromArchiveQuery` then you wouldn't need to worry about the media_request stuff.

Looking at postgres_app_v62_sec_v20_ret_v7 – Data Dictionary
(https://github.com/hmcts/darts-api/blob/dmp-2156-preview-audio-unavailable/src/main/resources/db/reference/modernised_darts_erd.pdf)
```
Table: darts.media
Linked from Table: darts.hearing_media_ae
Links to Table: darts.hearing
```

src/main/java/uk/gov/hmcts/darts/audio/component/AudioRequestBeingProcessedFromArchiveQuery.java
johnrace0: created new query component - AudioBeingProcessedFromArchiveQuery

src/main/java/uk/gov/hmcts/darts/audio/component/impl/AudioRequestBeingProcessedFromArchiveQueryImpl.java
johnrace0: modified the query etc to be based on hearing id - we now have a separate AudioBeingProcessedFromArchiveQuery - note that both this and the original query both return a list of AudioRequestBeingProcessedFromArchiveQueryResult - have left this named as it is rather than complicating things by renaming to something even more ghastly like AudioRequestorAudioBeingProcessedFromArchiveQueryResult

richardbull: Yep, I don't think it should be renamed to something like that.
I introduced it originally to represent the result model for my query around the notification: **audio_request_being_processed_from_archive**

It acts on the **Audio Requests** POST endpoint (media_request context) whereas yours is an **Audio** GET endpoint - for determining if preview should be available (media context)

src/main/java/uk/gov/hmcts/darts/audio/component/AudioBeingProcessedFromArchiveQuery.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.hmcts.darts.audio.component;
+
+import uk.gov.hmcts.darts.audio.model.AudioRequestBeingProcessedFromArchiveQueryResult;
+
+import java.util.List;
+
+public interface AudioBeingProcessedFromArchiveQuery {
+
+    List<AudioRequestBeingProcessedFromArchiveQueryResult> getResults(Integer hearingId);
```
richardbull, line 9: I'd find it a bit less confusing tbh if this result object was something different to match your query component - so `AudioBeingProcessedFromArchiveQueryResult`.  You might even be able to map directly to the generated `AudioMetadata`, so your query could do the channel1 filter and set the result fields in one hit rather than separately enriching the found entities in the `AudioController`?

src/integrationTest/java/uk/gov/hmcts/darts/audio/component/impl/AudioRequestBeingProcessedFromArchiveQueryImplIntTest.java
```diff
@@ -7,6 +7,7 @@
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.transaction.annotation.Transactional;
+import uk.gov.hmcts.darts.audio.component.AudioBeingProcessedFromArchiveQuery;
```
richardbull, line 10: I think I'd prefer to see a separate integration test for your query - however that looks?
`AudioBeingProcessedFromArchiveQueryImplIntTest`?
Do you agree @hemantasharma1129 ?

johnrace0: have split into separate integration test - to prevent any arguments :-D

src/main/java/uk/gov/hmcts/darts/audio/component/AudioBeingProcessedFromArchiveQuery.java
johnrace0: created new class for result - AudioBeingProcessedFromArchiveQueryResult.
left the rest of it alone though because a) it works ok b) individual methods could potentially reused for other purposes so combining would prevent this c) refactoring could have undesired affect of making code harder to understand

src/main/java/uk/gov/hmcts/darts/audio/component/impl/AudioRequestBeingProcessedFromArchiveQueryImpl.java
```diff
@@ -17,14 +17,13 @@
...
 public class AudioRequestBeingProcessedFromArchiveQueryImpl implements AudioRequestBeingProcessedFromArchiveQuery {
-
     private final NamedParameterJdbcTemplate jdbcTemplate;
     private final AudioRequestBeingProcessedFromArchiveQueryResultRowMapper rowMapper;
 
     @Override
     public List<AudioRequestBeingProcessedFromArchiveQueryResult> getResults(Integer mediaRequestId) {
         return jdbcTemplate.query(
-            """
+                """
```
richardbull, line 26: Can you restore the white space in this file - not sure if it makes any difference here, but notice your new Query class has it like I had it originally?

src/integrationTest/java/uk/gov/hmcts/darts/audio/component/impl/AudioBeingProcessedFromArchiveQueryImplIntTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+@Transactional
+@TestInstance(PER_CLASS)
+class AudioBeingProcessedFromArchiveQueryImplIntTest extends IntegrationBase {
+
+    @Autowired
+    private JdbcTemplate jdbcTemplate;
+
+    @Autowired
+    private AudioRequestBeingProcessedFromArchiveQuery audioRequestBeingProcessedFromArchiveQuery;
```
richardbull, line 28: This looks like it can be removed?

src/integrationTest/java/uk/gov/hmcts/darts/audio/component/impl/AudioRequestBeingProcessedFromArchiveQueryImplIntTest.java
```diff
@@ -26,6 +27,9 @@ class AudioRequestBeingProcessedFromArchiveQueryImplIntTest extends IntegrationB
     @Autowired
     private AudioRequestBeingProcessedFromArchiveQuery audioRequestBeingProcessedFromArchiveQuery;
 
+    @Autowired
+    private AudioBeingProcessedFromArchiveQuery audioBeingProcessedFromArchiveQuery;
```
richardbull, line 31: This can be removed also

richardbull (approved):

PR #1028, opened 2024-02-06
Aligning with google style specified via editor config

tonymort:
Aligning with google style specified via editor config

```
[ ] Yes
[x] No
```


.editorconfig
```diff
@@ -1,21 +1,1258 @@
...
-end_of_line = lf
 charset = utf-8
-trim_trailing_whitespace = true
-insert_final_newline = true
-max_line_length = 160
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = false
+max_line_length = 300
```
richardbull, line 7: Why 300 here over what we had previously 160?

.editorconfig
```diff
@@ -1,21 +1,1258 @@
...
-indent_size = 2
-end_of_line = lf
 charset = utf-8
-trim_trailing_whitespace = true
-insert_final_newline = true
-max_line_length = 160
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = false
```
richardbull, line 6: Referring to previous, should be true here?

.editorconfig
```diff
@@ -1,21 +1,1258 @@
-# http://editorconfig.org
-root = true
-
 [*]
```
richardbull, line 1: I think we should keep the option `trim_trailing_whitespace = true`

.editorconfig
```diff
@@ -1,21 +1,1258 @@
...
+ij_java_align_multiline_annotation_parameters = false
+ij_java_align_multiline_array_initializer_expression = false
+ij_java_align_multiline_assignment = false
+ij_java_align_multiline_binary_operation = false
+ij_java_align_multiline_chained_methods = false
+ij_java_align_multiline_deconstruction_list_components = true
+ij_java_align_multiline_extends_list = false
+ij_java_align_multiline_for = false
+ij_java_align_multiline_method_parentheses = false
+ij_java_align_multiline_parameters = false
```
richardbull, line 58: These 2 lines look like they should be true based on original?
```
ij_java_align_multiline_parameters = true
ij_java_align_multiline_parameters_in_calls = true
```

.editorconfig
```diff
@@ -1,21 +1,1258 @@
...
+ij_java_blank_lines_around_method_in_interface = 1
+ij_java_blank_lines_before_class_end = 0
+ij_java_blank_lines_before_imports = 1
+ij_java_blank_lines_before_method_body = 0
+ij_java_blank_lines_before_package = 0
+ij_java_block_brace_style = end_of_line
+ij_java_block_comment_add_space = false
+ij_java_block_comment_at_first_column = true
+ij_java_builder_methods = 
+ij_java_call_parameters_new_line_after_left_paren = false
```
richardbull, line 96: Also these 2 lines?

```
ij_java_call_parameters_new_line_after_left_paren = true
ij_java_call_parameters_right_paren_on_new_line = true
```

.editorconfig
```diff
@@ -1,21 +1,1258 @@
...
+ij_java_call_parameters_right_paren_on_new_line = false
+ij_java_call_parameters_wrap = normal
```
richardbull, line 98: `ij_java_call_parameters_wrap = on_every_item`?

.editorconfig
```diff
@@ -1,21 +1,1258 @@
...
+ij_java_filter_dd_suffix = 
+ij_java_finally_on_new_line = false
+ij_java_for_brace_force = always
+ij_java_for_statement_new_line_after_left_paren = false
+ij_java_for_statement_right_paren_on_new_line = false
+ij_java_for_statement_wrap = normal
+ij_java_generate_final_locals = false
+ij_java_generate_final_parameters = false
+ij_java_if_brace_force = always
+ij_java_imports_layout = *,|,java.**,javax.**,|,$*
```
richardbull, line 162: `ij_java_imports_layout = *,|,java.**,javax.**,|,$*,|`?

.idea/codeStyles/Project.xml
richardbull, line 1: Should these 2 files just be removed and ignored in gitignore? Don't think we need them if we are using `.editorconfig`?
```
.idea/codeStyles/Project.xml            .idea/codeStyles/codeStyleConfig.xml
```

mestebanez (commented):
Should this also feed into our checkstyle rules so we see build failures related to this?

.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -1,5 +1,6 @@
 <component name="ProjectCodeStyleConfiguration">
   <state>
     <option name="USE_PER_PROJECT_SETTINGS" value="true" />
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="GoogleStyle" />
```
mestebanez, line 4: Just out of interest, how does this work. I dont have the google style formatting installed by default?

mestebanez: i assume the pre-req here would be to install https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml

richardbull: Yep, I think that's the point, people probably weren't aware of importing the GoogleStyle scheme into IntelliJ (https://google.github.io/styleguide/intellij-java-google-style.xml)
They weren't in version control - so I think the project xml codeStyles can be removed if Tony has exported the GoogleStyle scheme into the `.editorconfig` and further customised as seen in our current file e.g. to correct the java imports

PR #1040, opened 2024-02-08
DMP-2245 - move validate_only to query param

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-2245

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/retention/controller/RetentionController.java
+++ b/src/main/java/uk/gov/hmcts/darts/retention/controller/RetentionController.java
@@ -43,9 +43,10 @@ public class RetentionController implements RetentionApi {
             securityRoles = {JUDGE, REQUESTER, APPROVER, TRANSCRIBER, TRANSLATION_QA, ADMIN},
             globalAccessSecurityRoles = {JUDGE, ADMIN})
     @Override
-    public ResponseEntity<PostRetentionResponse> retentionsPost(PostRetentionRequest postRetentionRequest) {
+    public ResponseEntity<PostRetentionResponse> retentionsPost(Boolean validateOnly,
+                                                                PostRetentionRequest postRetentionRequest) {
         RetentionsPostRequestValidator.validate(postRetentionRequest);
-        PostRetentionResponse response = retentionPostService.postRetention(postRetentionRequest);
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/hmcts/darts/retention/controller/RetentionController.java
```diff
@@ -43,9 +43,10 @@ public ResponseEntity<List<GetCaseRetentionsResponse>> retentionsGet(Integer cas
             securityRoles = {JUDGE, REQUESTER, APPROVER, TRANSCRIBER, TRANSLATION_QA, ADMIN},
             globalAccessSecurityRoles = {JUDGE, ADMIN})
     @Override
-    public ResponseEntity<PostRetentionResponse> retentionsPost(PostRetentionRequest postRetentionRequest) {
+    public ResponseEntity<PostRetentionResponse> retentionsPost(Boolean validateOnly,
```
mario-paniccia, line 46: minor: shall we add `@RequestParam` ? just to be consistent with what we do in other controllers

src/test/java/uk/gov/hmcts/darts/retention/service/impl/RetentionPostServiceImplTest.java
```diff
@@ -130,7 +130,7 @@ void fail_NoCase() {
 
         var exception = assertThrows(
```
mario-paniccia, line 131: shall we add a test where we verify that if the query param is null it defaults to `false` as per the specification?

mario-paniccia (approved):
just a couple of minor comments if you want

src/main/java/uk/gov/hmcts/darts/retention/controller/RetentionController.java
cmagowan-hmcts: there was actually a comment one time to not do that, as it appears in the RetentionApi that we are implementing (auto generated in there) and so we were asked not to duplicate it in the implementation.

src/test/java/uk/gov/hmcts/darts/retention/service/impl/RetentionPostServiceImplTest.java
cmagowan-hmcts: scenario now added

PR #1046, opened 2024-02-09
Increase max replica count ahead of perf test

jackmaloney:


charts/darts-api/values.yaml
```diff
@@ -94,7 +94,7 @@ function:
   aadIdentityName: darts
   pollingInterval: 60
   minReplicaCount: 0
-  maxReplicaCount: 2
+  maxReplicaCount: 5
```
davet1985, line 97: We should probably do this in flux, otherwise we will get 5 replicas on each cluster in each environment?!

PR #1048, opened 2024-02-09
 DMP-2259 Move @Transactional to single iteration of OutboundAudioDeleter automated task

mario-paniccia:
…d operation iteration to improve job performance

### JIRA link ###
https://tools.hmcts.net/jira/browse/DMP-2259

### Change description ###
Additional changes:
- removed `@Transactional` from integration tests to prevent false negatives (test passing when instead it should have failed due to a bug in the production code). Integration tests would be better not be `@Transactional` to have the test running in a setting that replicates Production
- changed some JPA one-to-one entity associations to be eagerly fetched (default) rather than lazy. Lazy loading is recommended for *-to-many associations, but lazy loading one-to-one complicates working with entities and should be enabled only **after** measuring performance and if justified by a relevant perf improvement. To avoids premature optimisations.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/audio/service/impl/OutboundAudioDeleterProcessorSingleElementImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/service/impl/OutboundAudioDeleterProcessorSingleElementImpl.java
@@ -1,0 +1,75 @@
+package uk.gov.hmcts.darts.audio.service.impl;
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+import uk.gov.hmcts.darts.audio.entity.MediaRequestEntity;
+import uk.gov.hmcts.darts.audio.enums.MediaRequestStatus;
+import uk.gov.hmcts.darts.audio.service.OutboundAudioDeleterProcessorSingleElement;
+import uk.gov.hmcts.darts.common.entity.ObjectRecordStatusEntity;
+import uk.gov.hmcts.darts.common.entity.TransformedMediaEntity;
+import uk.gov.hmcts.darts.common.entity.TransientObjectDirectoryEntity;
+import uk.gov.hmcts.darts.common.entity.UserAccountEntity;
+import uk.gov.hmcts.darts.common.repository.ObjectRecordStatusRepository;
+import uk.gov.hmcts.darts.common.repository.TransformedMediaRepository;
+import uk.gov.hmcts.darts.common.repository.TransientObjectDirectoryRepository;
+
+import java.time.OffsetDateTime;
+import java.util.List;
+
+import static uk.gov.hmcts.darts.common.enums.ObjectRecordStatusEnum.MARKED_FOR_DELETION;
+
+
+@Service
+@RequiredArgsConstructor
+@Slf4j
+public class OutboundAudioDeleterProcessorSingleElementImpl implements OutboundAudioDeleterProcessorSingleElement {
+
+    private final ObjectRecordStatusRepository objectRecordStatusRepository;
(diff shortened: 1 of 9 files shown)


tonymort (approved):

src/main/java/uk/gov/hmcts/darts/common/entity/base/CreatedModifiedBaseEntity.java
```diff
@@ -21,7 +20,7 @@ public class CreatedModifiedBaseEntity extends CreatedBaseEntity {
     @Column(name = "last_modified_ts")
     private OffsetDateTime lastModifiedDateTime;
 
-    @ManyToOne(fetch = FetchType.LAZY)
+    @ManyToOne
```
cmagowan-hmcts, line 23: was there a reason the Lazy loading was removed?

mario-paniccia: yep I've mentioned in the PR description. Happy to have a chat about it if you have concerns Chris

src/main/java/uk/gov/hmcts/darts/audio/service/impl/OutboundAudioDeleterProcessorSingleElementImpl.java
```diff
@@ -0,0 +1,83 @@
...
+        if (areAllTransformedMediasExpired) {
+            mediaRequest.setLastModifiedBy(userAccount);
+            mediaRequest.setStatus(MediaRequestStatus.EXPIRED);
+            mediaRequestRepository.saveAndFlush(mediaRequest);
+        }
+    }
+
+    private void markTransformedMediaAsExpired(UserAccountEntity userAccount, TransformedMediaEntity transformedMedia) {
+
+        transformedMedia.setExpiryTime(OffsetDateTime.now());
```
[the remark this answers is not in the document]
mario-paniccia: I thought that was tests. This is production code. Unless my understanding was wrong

src/main/java/uk/gov/hmcts/darts/common/entity/base/CreatedModifiedBaseEntity.java
cmagowan-hmcts: i think we should be leaving the Lazy loading. this class is used by practically every entity and so will result in extra db calls all over the system for no reason, as nothing needs to know the users name for functionality.
I don't think we should be changing code behaviour just to help make a test a little shorter.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/OutboundAudioDeleterProcessorSingleElementImpl.java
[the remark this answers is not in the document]
cmagowan-hmcts: it was to aid in tests yes, using currentDateTimeHelper in the code allows the test to mock it and mock any dateTime they want, which then allows you to assert that expiryTime was set to that specific dateTime.

src/main/java/uk/gov/hmcts/darts/common/entity/base/CreatedModifiedBaseEntity.java
mario-paniccia: fair enough, I've reverted the change.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/OutboundAudioDeleterProcessorSingleElementImpl.java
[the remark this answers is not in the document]
mario-paniccia: got it. It's a good improvement 👍 I've changed the code to use the currentDateTimeHelper, cheers

PR #1056, opened 2024-02-13
Dmp 2139 arm clean up job asynchronous

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-2139

### Change description ###

Created an automated task to cleanup ARM response files

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/arm/api/ArmDataManagementApi.java
```diff
@@ -19,5 +19,7 @@ public interface ArmDataManagementApi extends BlobContainerDownloadable {
 
     void deleteResponseBlob(String blobName);
 
+    boolean deleteBlobData(String blobPathAndName);
```
richardbull, line 22: I don't think we need another delete operation here - the one above seemed suitable when I was tripping over the deleteResponseBlobs stuff in my ticket - see https://github.com/hmcts/darts-api/blob/DMP-2045-ArmResponseFiles-cleanup/src/main/java/uk/gov/hmcts/darts/arm/api/ArmDataManagementApi.java#L20

I think we should align to what I have in my branch as that already had the blob path stuff already `dropzone/DARTS/response/`

src/main/java/uk/gov/hmcts/darts/task/service/impl/AutomatedTaskServiceImpl.java
```diff
@@ -312,78 +317,85 @@ private void addInboundToUnstructuredTaskRegistrar(ScheduledTaskRegistrar taskRe
...
-            transcriptionsApi
+                automatedTaskRepository,
+                lockProvider,
+                automatedTaskConfigurationProperties,
+                transcriptionsApi
         );
-        closeUnfinishedTranscriptionsAutomatedTask.setLastCronExpression(getAutomatedTaskCronExpression(
-            closeUnfinishedTranscriptionsAutomatedTask));
+        closeUnfinishedTranscriptionsAutomatedTask.setLastCronExpression(getAutomatedTaskCronExpression(closeUnfinishedTranscriptionsAutomatedTask));
         Trigger trigger = createAutomatedTaskTrigger(closeUnfinishedTranscriptionsAutomatedTask);
```
johnrace0, line 338: indentation looks a bit weird here - but we seem to be having issues all over lately with formatting!

src/main/java/uk/gov/hmcts/darts/arm/service/ArmService.java
```diff
@@ -26,4 +26,5 @@ public interface ArmService {
      */
     boolean deleteBlobData(String containerName, String blobPathAndName);
 
+
```
richardbull, line 29: Can we remove this extra white space?

src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmResponseFilesProcessorImpl.java
```diff
@@ -38,7 +38,7 @@ public class ArmResponseFilesProcessorImpl implements ArmResponseFilesProcessor
     @Override
     public void processResponseFiles() {
         initialisePreloadedObjects();
-        // Fetch All records from external_object_directory table with external_location_type as 'ARM' and object_record_status with 'Arm Drop Zone'.
+        // Fetch All records from external Object Directory table with external_location_type as 'ARM' and status with 'ARM dropzone'.
```
richardbull, line 41: external_object_directory.ors_id:13	'Arm Drop Zone'

src/main/java/uk/gov/hmcts/darts/task/runner/impl/CleanupArmResponseFilesAutomatedTask.java
```diff
@@ -0,0 +1,38 @@
...
+import lombok.extern.slf4j.Slf4j;
+import net.javacrumbs.shedlock.core.LockProvider;
+import uk.gov.hmcts.darts.arm.service.CleanupArmResponseFilesService;
+import uk.gov.hmcts.darts.common.repository.AutomatedTaskRepository;
+import uk.gov.hmcts.darts.task.config.AutomatedTaskConfigurationProperties;
+
+import static uk.gov.hmcts.darts.task.runner.AutomatedTaskName.CLEANUP_ARM_RESPONSE_FILES_TASK_NAME;
+
+@Slf4j
+public class CleanupArmResponseFilesAutomatedTask extends AbstractLockableAutomatedTask {
```
johnrace0, line 12: general observation - we have loads of these automated task classes now and they all do the same thing - only difference is the task name and the service class. just wondering if we could do this some other way?

src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmServiceImpl.java
```diff
@@ -137,7 +137,7 @@ public boolean deleteBlobData(String containerName, String blobPathAndName) {
             }
 
         } catch (Exception e) {
-            log.error("Could not delete from container: " + containerName + " blobPathAndName: " + blobPathAndName, e.getMessage(), e);
+            log.error("Could not delete from container {} blobPathAndName {} - {}", containerName, blobPathAndName, e.getMessage(), e);
```
richardbull, line 140: Wondering why we need the exception message and stacktrace, does stacktrace not give everything anyway?

src/main/java/uk/gov/hmcts/darts/arm/util/ArmResponseFilesHelper.java
```diff
@@ -0,0 +1,40 @@
...
+import static java.util.Objects.nonNull;
+import static uk.gov.hmcts.darts.arm.util.ArchiveConstants.ArchiveRecordOperationValues.ARM_FILENAME_SEPARATOR;
+import static uk.gov.hmcts.darts.arm.util.ArchiveConstants.ArchiveResponseFileAttributes.ARM_RESPONSE_FILE_EXTENSION;
+
+@UtilityClass
+@SuppressWarnings({"HideUtilityClassConstructor"})
+public class ArmResponseFilesHelper {
+
+    public static String generateSuffix(String filenameKey) {
+        return ARM_FILENAME_SEPARATOR + filenameKey + ARM_RESPONSE_FILE_EXTENSION;
```
johnrace0, line 15: is it worth using stringbuilder for this one as in getPrefix()?

src/main/java/uk/gov/hmcts/darts/arm/service/impl/CleanupArmResponseFilesServiceImpl.java
```diff
@@ -0,0 +1,230 @@
...
+                } else {
+                    log.warn("ARM response file {} not input upload file for EOD {}", armInputUploadFilename, externalObjectDirectory.getId());
+                }
+            }
+        } else {
+            log.info("Unable to find input upload file with prefix {} for cleanup with EOD {}", prefix, externalObjectDirectory.getId());
+        }
+    }
+
+    private void readInputUploadFile(ExternalObjectDirectoryEntity externalObjectDirectory, String armInputUploadFilename) {
```
johnrace0, line 118: not sure the name of this method is conveying what it is trying to do

src/main/java/uk/gov/hmcts/darts/arm/service/impl/CleanupArmResponseFilesServiceImpl.java
```diff
@@ -0,0 +1,230 @@
...
+            }
+        } catch (IllegalArgumentException e) {
+            // This occurs when the filename is not parsable
+            log.error("Unable to process filename {} for {} - {}", armInputUploadFilename, externalObjectDirectory.getId(), e.getMessage());
+        } catch (Exception e) {
+            log.error("Unable to list responses for file {} for {} - {}", armInputUploadFilename, externalObjectDirectory.getId(), e.getMessage());
+        }
+    }
+
+    private void processResponseBlobs(List<String> responseBlobs, ExternalObjectDirectoryEntity externalObjectDirectory,
```
johnrace0, line 137: maybe deleteResponseBlobs()?

src/main/java/uk/gov/hmcts/darts/arm/service/impl/CleanupArmResponseFilesServiceImpl.java
```diff
@@ -0,0 +1,230 @@
...
+    private ObjectRecordStatusEntity failedArmResponseChecksum;
+    private UserAccountEntity userAccount;
+
+
+    @Override
+    public void cleanupResponseFiles() {
+        initialisePreloadedObjects();
+
+        ExternalLocationTypeEntity armLocation = externalLocationTypeRepository.getReferenceById(ExternalLocationTypeEnum.ARM.getId());
+        List<ObjectRecordStatusEntity> statuses = List.of(storedStatus,
```
richardbull, line 67: I think it would be better if we didn't need to populate these class fields - we don't really need to do the `findById` below as we just want the status id's for the query. So I think using the `getReferenceById` in the same way as armLocation is fine as we're not doing anything with the entity e.g. logging description or whatever was being done in the other service.

It can be simplified further by using `.id in (statusIds)` in the `findByStatusInAndStorageLocationAndResponseCleanedAndLastModifiedDateTimeBefore` query.

src/main/java/uk/gov/hmcts/darts/arm/service/impl/CleanupArmResponseFilesServiceImpl.java
```diff
@@ -0,0 +1,230 @@
...
+        for (String responseFile : responseBlobs) {
+            if (responseFile.endsWith(ArmResponseFilesHelper.generateSuffix(ARM_CREATE_RECORD_FILENAME_KEY))) {
+                createRecordFilename = responseFile;
+            } else if (responseFile.endsWith(ArmResponseFilesHelper.generateSuffix(ARM_UPLOAD_FILE_FILENAME_KEY))) {
+                uploadFilename = responseFile;
+            } else if (responseFile.endsWith(ArmResponseFilesHelper.generateSuffix(ARM_INVALID_LINE_FILENAME_KEY))) {
+                invalidFileFilename = responseFile;
+            }
+        }
+
```
johnrace0, line 152: maybe call armResponseFilesToBeDeleted.clear() here just to be safe - i can't tell whether it is or should be empty at this stage

src/main/java/uk/gov/hmcts/darts/arm/service/impl/CleanupArmResponseFilesServiceImpl.java
```diff
@@ -0,0 +1,230 @@
...
+    private void initialisePreloadedObjects() {
+        storedStatus = objectRecordStatusRepository.findById(STORED.getId()).get();
+        failedArmResponseManifestFileStatus = objectRecordStatusRepository.findById(ARM_RESPONSE_MANIFEST_FAILED.getId()).get();
+        failedArmResponseProcessing = objectRecordStatusRepository.findById(ARM_RESPONSE_PROCESSING_FAILED.getId()).get();
+        failedArmResponseChecksum = objectRecordStatusRepository.findById(ARM_RESPONSE_CHECKSUM_VERIFICATION_FAILED.getId()).get();
+
+        userAccount = userIdentity.getUserAccount();
+    }
+
+    public static String getPrefix(ExternalObjectDirectoryEntity externalObjectDirectory) {
```
richardbull, line 224: Is this the right access modifier?

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
```diff
@@ -134,6 +134,22 @@ List<Integer> findMediaFileIdsIn2StorageLocationsBeforeTime(ObjectRecordStatusEn
...
                                                                 OffsetDateTime lastModifiedBefore);
 
+
+    @Query(
+        """
+            SELECT eod FROM ExternalObjectDirectoryEntity eod
+            WHERE eod.status in :statuses
+            AND eod.externalLocationType = :type
+            AND eod.responseCleaned = :responseCleaned
+            AND eod.lastModifiedDateTime < :lastModifiedBefore
```
richardbull, line 144: I think we should add an order by clause here - so that the earliest will always be at the top of the list?
`ORDER BY eod.lastModifiedDateTime`.
What do you think?

src/main/resources/db/migration/common/V1_273__cleanup_arm_response_files_automated_task.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO darts.automated_task (aut_id,task_name,task_description,cron_expression,cron_editable)
+VALUES (10,'CleanupArmResponseFiles','Cleans up ARM response files','0 0 21 * * *',true);
+
+ALTER SEQUENCE aut_seq RESTART WITH 11;
```
richardbull, line 4: Wondering why we have a mixture of `darts` schema use in the same sql file - should the sequence also have the `darts.` prefix?

src/test/java/uk/gov/hmcts/darts/arm/service/CleanupArmResponseFilesServiceImplTest.java
```diff
@@ -0,0 +1,139 @@
...
+            userIdentity,
+            armDataManagementConfiguration,
+            currentTimeHelper
+        );
+    }
+
+    @Test
+    void cleanupResponseFilesSuccess() {
+
+        when(externalLocationTypeRepository.getReferenceById(3)).thenReturn(externalLocationTypeArm);
```
richardbull, line 93: I'd prefer to see us using the standing data enum here in the tests for better readability like we use them in our service code.

src/test/java/uk/gov/hmcts/darts/arm/service/CleanupArmResponseFilesServiceImplTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+        when(externalObjectDirectoryRepository.findByStatusInAndStorageLocationAndResponseCleanedAndLastModifiedDateTimeBefore(
+            any(),
+            any(),
+            anyBoolean(),
+            any()
+        )).thenReturn(List.of(externalObjectDirectory));
+
+        String inputUploadBlobFilename = "123_456_1_6a374f19a9ce7dc9cc480ea8d4eca0fb_1_iu.rsp";
+        List<String> inputUploadFilenameResponseBlobs = new ArrayList<>();
```
richardbull, line 118: We could use `List.of` in these List instantiations to simplify it also

src/test/java/uk/gov/hmcts/darts/arm/service/CleanupArmResponseFilesServiceImplTest.java
```diff
@@ -0,0 +1,139 @@
...
+        hashcodeResponseBlobs.add(uploadFileFilename);
+        when(armDataManagementApi.listResponseBlobs(hashcode)).thenReturn(hashcodeResponseBlobs);
+
+        when(armDataManagementApi.deleteBlobData(any())).thenReturn(true);
+
+        when(userIdentity.getUserAccount()).thenReturn(testUser);
+
+        cleanupArmResponseFilesService.cleanupResponseFiles();
+
+        verify(externalObjectDirectoryRepository, times(1)).saveAndFlush(externalObjectDirectoryEntityCaptor.capture());
```
richardbull, line 136: No need for `times(1)` in verify - that is the default.
I guess if we didn't mock the externalObjectDirectory we could verify the flag field `responseCleaned`

src/integrationTest/java/uk/gov/hmcts/darts/arm/service/CleanupArmResponseFilesServiceIntTest.java
```diff
@@ -0,0 +1,677 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+import static uk.gov.hmcts.darts.common.enums.ExternalLocationTypeEnum.ARM;
+import static uk.gov.hmcts.darts.common.enums.ObjectRecordStatusEnum.ARM_RESPONSE_CHECKSUM_VERIFICATION_FAILED;
+import static uk.gov.hmcts.darts.common.enums.ObjectRecordStatusEnum.ARM_RESPONSE_MANIFEST_FAILED;
+import static uk.gov.hmcts.darts.common.enums.ObjectRecordStatusEnum.ARM_RESPONSE_PROCESSING_FAILED;
+import static uk.gov.hmcts.darts.common.enums.ObjectRecordStatusEnum.STORED;
+
+@SpringBootTest
+@ActiveProfiles({"intTest", "h2db"})
```
richardbull, line 44: What's the reason for overriding these annotations from `IntegrationBase`?

src/integrationTest/java/uk/gov/hmcts/darts/arm/service/CleanupArmResponseFilesServiceIntTest.java
```diff
@@ -0,0 +1,677 @@
...
+    private ArmDataManagementConfiguration armDataManagementConfiguration;
+    @MockBean
+    private CurrentTimeHelper currentTimeHelper;
+
+    private CleanupArmResponseFilesService cleanupArmResponseFilesService;
+
+    @BeforeEach
+    void setupData() {
+
+        cleanupArmResponseFilesService = new CleanupArmResponseFilesServiceImpl(
```
richardbull, line 69: Pretty sure if you autowired the `CleanupArmResponseFilesService`, looking at the other Autowired and MockBean usage, I'd expect this to happen anyway without being explicit here

src/integrationTest/java/uk/gov/hmcts/darts/arm/service/CleanupArmResponseFilesServiceIntTest.java
```diff
@@ -0,0 +1,677 @@
...
+        );
+
+        MediaEntity savedMedia = dartsDatabase.save(
+            MediaTestData.createMediaWith(
+                hearing.getCourtroom(),
+                OffsetDateTime.parse("2023-09-26T13:00:00Z"),
+                OffsetDateTime.parse("2023-09-26T13:45:00Z"),
+                1
+            ));
+        savedMedia.setChecksum("C3CCA7021CF79B42F245AF350601C284");
```
richardbull, line 97: I thought our checksums seemed to end with == e.g. `T2UrmSYWNZmvvYqBcMnV0g==`.
I think we were expecting them to be based on what we see in the Azure Blob Storage CONTENT-MD5 tag value as a base64 encoded representation of the binary MD5 hash value.
Hopefully we don't have a mismatch here of expected values for ARM checksum verification? What do you think @hemantasharma1129

src/integrationTest/java/uk/gov/hmcts/darts/arm/service/CleanupArmResponseFilesServiceIntTest.java
```diff
@@ -0,0 +1,677 @@
...
+
+        List<String> hashcodeResponseBlobs = new ArrayList<>();
+        String hashcode = "6a374f19a9ce7dc9cc480ea8d4eca0fb";
+        String createRecordFilename = "6a374f19a9ce7dc9cc480ea8d4eca0fb_a17b9015-e6ad-77c5-8d1e-13259aae1895_1_cr.rsp";
+        String uploadFileFilename = "6a374f19a9ce7dc9cc480ea8d4eca0fb_04e6bc3b-952a-79b6-8362-13259aae1895_1_uf.rsp";
+        hashcodeResponseBlobs.add(createRecordFilename);
+        hashcodeResponseBlobs.add(uploadFileFilename);
+        when(armDataManagementApi.listResponseBlobs(hashcode)).thenReturn(hashcodeResponseBlobs);
+
+        when(armDataManagementApi.deleteBlobData(any())).thenReturn(true);
```
richardbull, line 129: I think it would be better to add the verify for the mocking to make it more understandable for the next person what files we're expecting to delete in the test?

src/integrationTest/java/uk/gov/hmcts/darts/arm/service/CleanupArmResponseFilesServiceIntTest.java
```diff
@@ -0,0 +1,677 @@
...
+        Optional<ExternalObjectDirectoryEntity> foundMediaEod = dartsDatabase.getExternalObjectDirectoryRepository().findById(armEod.getId());
+        assertTrue(foundMediaEod.isPresent());
+        ExternalObjectDirectoryEntity foundMedia = foundMediaEod.get();
+        assertTrue(foundMedia.isResponseCleaned());
+    }
+
+    @Test
+    void cleanupResponseFilesSuccessWithFiles_InputUpload_CreateRecord_UploadFileAndStateManifestFailed() {
+
+        HearingEntity hearing = dartsDatabase.createHearing(
```
richardbull, line 148: If all these tests have similar setup, can we move this stuff into the `setupData` method above to make the test smaller to read?

src/main/resources/application.yaml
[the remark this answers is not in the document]
richardbull: I prefer Duration, but given the property is named appropriately i.e. it says days - then it feels fine to me.
Much rather see that than half a change done to say duration in the property name like I've seen done previously :)
Only say duration in the property name, if the value is Duration is a good start.

src/main/java/uk/gov/hmcts/darts/task/runner/impl/CleanupArmResponseFilesAutomatedTask.java
karen-hedges: Maybe best to raise a ticket about this

src/main/java/uk/gov/hmcts/darts/arm/util/ArmResponseFilesHelper.java
karen-hedges: Intellij complains about using string builder for this. Also its just adding 3 strings which I would consider the max for strings being concatenated

src/integrationTest/java/uk/gov/hmcts/darts/arm/service/CleanupArmResponseFilesServiceIntTest.java
karen-hedges: Copy and paste issue. Will correct

src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmServiceImpl.java
karen-hedges: I don't generally use stacktrace unless theres a real need. I was debating whether it should have stacktrace or not.

src/main/java/uk/gov/hmcts/darts/task/runner/impl/CleanupArmResponseFilesAutomatedTask.java
johnrace0: will forward to Hemanta to consider

johnrace0: @hemantasharma1129 - might or might not be worth looking into this

src/integrationTest/java/uk/gov/hmcts/darts/arm/service/CleanupArmResponseFilesServiceIntTest.java
karen-hedges: This was done when checksum was nullable but its now changed so has a value set when created. I have taken this out.

src/main/java/uk/gov/hmcts/darts/arm/service/impl/CleanupArmResponseFilesServiceImpl.java
karen-hedges: It is cleared at line 95 and 204 depending on which loop it is in

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
```diff
@@ -134,6 +134,22 @@ List<Integer> findMediaFileIdsIn2StorageLocationsBeforeTime(ObjectRecordStatusEn
...
+            """
+    )
+    List<ExternalObjectDirectoryEntity> findByStatusInAndStorageLocationAndResponseCleanedAndLastModifiedDateTimeBefore(
```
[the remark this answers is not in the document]
karen-hedges: It is because I was going to try not to use the query but it does convey what fields it is using

src/integrationTest/java/uk/gov/hmcts/darts/task/service/AutomatedTaskServiceTest.java
```diff
@@ -593,9 +595,7 @@ void givenConfiguredTasksUpdateCronAndResetCronForProcessArmResponseFilesAutomat
 
         Optional<AutomatedTaskEntity> originalAutomatedTaskEntity =
             automatedTaskService.getAutomatedTaskEntityByTaskName(automatedTask.getTaskName());
-        log.info("TEST - Original task {} cron expression {}", automatedTask.getTaskName(),
-                 originalAutomatedTaskEntity.get().getCronExpression()
-        );
+        log.info("TEST - Original task {} cron expression {}", automatedTask.getTaskName(), originalAutomatedTaskEntity.get().getCronExpression());
 
         automatedTaskService.updateAutomatedTaskCronExpression(automatedTask.getTaskName(), "*/9 * * * * *");
```
richardbull, line 600: Is this Cron expression correct for "Daily @ 9PM".  Looking at https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/support/CronExpression.html
I expected the third asterisk to have `21`?

richardbull (approved):

hemantasharma1129 (approved):

PR #1059, opened 2024-02-13
DMP-1993 - make use of new LogApi

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-1993


davet1985 (changes requested):
Tests please 🙏

davet1985 (approved):
🐹

PR #1090, opened 2024-02-19
DMP-1139 Added DETS datastore lookup

johnrace0:
https://tools.hmcts.net/jira/browse/DMP-1139

### Change description ###
Preview Audio - Fetch from ODS (Unstructured) or DETS

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
```diff
@@ -298,11 +304,17 @@ private Map<MediaEntity, Path> downloadAndSaveMediaToWorkspace(List<MediaEntity>
...
-        log.debug("Downloading audio blob for {} from unstructured datastore", id);
-        BinaryData binaryData = getUnstructuredAudioBlob(id);
+        BinaryData binaryData = null;
+        try {
+            log.debug("Downloading audio blob for {} from unstructured datastore", id);
+            binaryData = getUnstructuredAudioBlob(id);
+        } catch (BlobStorageException blobStorageException) {
+            log.debug("Downloading audio blob for {} from dets datastore", id);
+            binaryData = getDetsAudioBlob(id);
+        }
```
mario-paniccia, line 317: I wonder whether it would be better to be more specific here, i.e. downloading from DETS only if we get an exception that is the equivalent in Azure of a 404. 
This would prevent a scenario where the file is in unstructured, but download fails for some reason, and we do an unnecessary download from DETS which will fail because the file is not there

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
```diff
@@ -298,11 +304,17 @@ private Map<MediaEntity, Path> downloadAndSaveMediaToWorkspace(List<MediaEntity>
     @Override
     public Path saveMediaToWorkspace(MediaEntity mediaEntity) throws IOException {
         UUID id = getMediaLocation(mediaEntity).orElseThrow(
```
mario-paniccia, line 306: as is we'd be using the UUID of the file in unstructured. Would the file in DETS be guaranteed to have the same UUID? Or we should update `getMediaLocation` to find the unique UUID for the file in DETS?

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
johnrace0: @richardbull - can we get a second opinion on this please

hemantasharma1129: We don't need these checks here, its already been done in Facade.

johnrace0: @hemantasharma1129 facade checks for unstructured -> dets -> arm. are we ok to use it here (since we are not checking arm)?

richardbull: I don't particularly like the `id` variable name that was there before, we chop and change between using `location` and `externalLocation` quite a bit. I prefer externalLocation, only because that corresponds to our table `external_object_directory.external_location`. In Azure Storage it's just a uuid `Name` property.

I'm not sure about the `BlobStorageException` currently as I haven't done much work in that area. I see what @mario-paniccia is saying though. As a tiny compromise, we could possibly add the `blobStorageException` to the debug line?

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
mario-paniccia: what I was wondering if it's possible that the same file has got 2 different UUIDs for DETs and unstructured

johnrace0: possibly - but i think that would be an error with data upload/storage - since the eod only has 1 uuid

src/main/java/uk/gov/hmcts/darts/datamanagement/api/DataManagementApi.java
```diff
@@ -13,6 +13,8 @@ public interface DataManagementApi extends BlobContainerDownloadable {
 
     BinaryData getBlobDataFromUnstructuredContainer(UUID blobId);
 
+    BinaryData getBlobDataFromDetsContainer(UUID blobId);
```
mestebanez, line 16: If this is DETS related wouldnt thew new method be on the uk.gov.hmcts.darts.dets.api.DetsDataManagementApi

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
```diff
@@ -109,6 +110,11 @@ public BinaryData getUnstructuredAudioBlob(UUID location) {
         return dataManagementApi.getBlobDataFromUnstructuredContainer(location);
     }
 
+    @Override
+    public BinaryData getDetsAudioBlob(UUID location) {
+        return dataManagementApi.getBlobDataFromDetsContainer(location);
```
mestebanez, line 115: Call BlobContainerDownloadable.downloadBlobFromContainer to avoid bringing downloads into memory

PR #1105, opened 2024-02-21
DMP-2235 Alter EOD sequence

johnrace0:
https://tools.hmcts.net/jira/browse/DMP-2235

### Change description ###
Update the Sequence of External Object Directory table

```
[ ] Yes
[X] No
```


src/main/resources/db/migration/common/V1_275__alter_eod_sequence.sql
```diff
@@ -0,0 +1 @@
+ALTER SEQUENCE eod_seq RESTART WITH 300000000;
```
karen-hedges, line 1: I'm just wondering if the id is starting so high, is there really enough integers left before hitting the max:
  2147,483,647 max ints
-  300,000,000
= 1847,483,647

hemantasharma1129: 300 mill - after discussing with migration.

1.8 billion - Should be fine.I did kept it as Long in the ticket initially but in refinement others have suggested 1.8 Bill should be fine. Please note migration only has some 80-100 million records which is accumulated in last 7-8 years.

hemantasharma1129 (approved):

richardbull (approved):

karen-hedges (approved):

PR #1111, opened 2024-02-22
DMP-2382 fixed bug in logging media processing failure

mario-paniccia:
https://tools.hmcts.net/jira/browse/DMP-2382

```
[ ] Yes
[ x] No
```


src/integrationTest/java/uk/gov/hmcts/darts/audio/service/AudioTransformationServiceHandleKedaInvocationForMediaRequestsTest.java
```diff
@@ -139,7 +139,6 @@ public void handleKedaInvocationForMediaRequestsShouldSucceedAndUpdateRequestSta
 
     @ParameterizedTest
     @EnumSource(names = {"DOWNLOAD", "PLAYBACK"})
-    @Transactional
```
mario-paniccia, line 142: I've used this integration test to manually verify my fix. 
Interestingly, having this test as `@Transactional` did not allow to see the bug. When `@Transactional` the log is correctly an error log.
Only when removing `@Transactional` the bug showed up revealing that we were logging the wrong log message  in Production.
This reinforces my believe that we should avoid `@Transactional` tests. They cause plenty of false positives

hemantasharma1129 (approved):

davet1985 (approved):
❤️

src/integrationTest/java/uk/gov/hmcts/darts/audio/service/AudioTransformationServiceHandleKedaInvocationForMediaRequestsTest.java
jackmaloney: I was just looking at this exact thing! This was giving a false positive in the code, and looks like a 'failed' state had been set on an object when in fact it was still set to 'processing' meaning the wrong thing is getting logged to the console. We should probably remove transactional from all these tests and then fix the code to set the state correctly.

mario-paniccia: I agree Jack, I would suggest we avoid `@Transactional` on the future tests we write

PR #1122, opened 2024-02-23
DMP-782: Implement streams for first phase of add audio flow to impro…

chrisbellingham-hmcts:
# [DMP-782](https://tools.hmcts.net/jira/browse/DMP-782)

The existing `POST /audios` endpoint accepts a binary payload (audio) which is subsequently uploaded to the Inbound Datastore via the Azure Blob Client.

The existing implementation reads the full binary payload into memory at various stages in this flow, including:
* When computing the checksum
* When passing the `BinaryData` instance to the blob client's `upload` method.

This PR reworks that flow to favour streams, such that:
* The checksum is computed as the stream is consumed, using `DigestInputStream`
* The input stream is passed directly to the blob client

The blobClient will break the payload down into multiple pieces, and will initiate an API (upload) call to Azure for each piece. BlobClient will consume into memory the entire size of that piece, so we also configure BlobClient to limit the size of those pieces. Additional configuration parameters are exposed as Spring properties such that they can be tuned during performance testing. Current values are not expected to be final.

## Testing

Behaviour between `master` branch and `dmp-782-streams` branch was compared **locally**.

An audio file - `noise_large.mp2` - was generated with filesize of ~320MB, close to the 350MB max filesize supported by DARTS. Requests were made to `POST /audios` as follows to observe behaviour for such large files:

```shell
curl -v --request POST \
  --url http://localhost:4550/audios \
  --header 'Authorization: Bearer <token>' \
  --header 'Content-Type: multipart/form-data' \
  --form 'file=@/Users/chrisbellingham/Music/noise_large.mp2;type=audio/mpeg' \
  --form 'metadata={"started_at":"2023-11-23T10:36:05.362Z","ended_at":"2023-11-23T10:36:05.362Z","channel":1,"total_channels":4,"format":"mp2","filename":"filename_goes_here","courthouse":"SWANSEA","courtroom":1,"media_file":"media_file_goes_here","file_size":100,"checksum":null,"cases":[1]};type=application/json'
```

JVM was configured to set fixed values for Max and Initial heap (`-Xmx1024m -Xms1024m`) to give a stable basis for comparison.

### Existing `master` branch

Request fails with:
```
Caused by: java.lang.OutOfMemoryError: Java heap space
```

### `dmp-782-streams` branch

Request succeeds after ~3 mins. File is present within the Inbound container in Azure, and an `external_object_directory` record is created with the expected checksum value.

To test how the memory usage scales with concurrency, two requests were made ~1 minute apart:

add_audio_profile

Garbage collection was manually triggered aggressively during upload to obtain an accurate view of actual usage (blue line). From delta between requests, the memory cost of a single request is ~70MB for a 320MB request payload.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/util/FileContentChecksum.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/util/FileContentChecksum.java
@@ -2,23 +2,30 @@ package uk.gov.hmcts.darts.common.util;
 
 import org.springframework.stereotype.Component;
 
+import java.security.DigestInputStream;
+
 import static org.apache.commons.codec.binary.Base64.encodeBase64;
 import static org.apache.commons.codec.digest.DigestUtils.md5;
 
+/**
+ * Calculates the file checksum equivalent to the Azure Blob Storage CONTENT-MD5 tag value as a base64 encoded
+ * representation of the binary MD5 hash value. This could be done in Bash e.g. openssl md5 -binary "Test
+ * Document.doc" | base64
+ */
 @Component
 @SuppressWarnings({"java:S4790"})
 public class FileContentChecksum {
 
-    /**
-     * Calculates the file checksum equivalent to the Azure Blob Storage CONTENT-MD5 tag value as a base64 encoded
-     * representation of the binary MD5 hash value. This could be done in Bash e.g. openssl md5 -binary "Test
-     * Document.doc" | base64
-     *
-     * @param bytes the file content
-     * @return the calculated checksum
-     */
     public String calculate(byte[] bytes) {
-        return new String(encodeBase64(md5(bytes)));
+        return encodeToString(md5(bytes));
+    }
+
+    public String calculate(DigestInputStream digestInputStream) {
+        return encodeToString(digestInputStream.getMessageDigest().digest());
+    }
+
+    private String encodeToString(byte[] bytes) {
+        return new String(encodeBase64(bytes));
     }
 
 }
--- a/src/main/java/uk/gov/hmcts/darts/common/util/FileContentChecksum.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/util/FileContentChecksum.java
@@ -13,9 +13,13 @@ import static org.apache.commons.codec.digest.DigestUtils.md5;
  * Document.doc" | base64
  */
 @Component
-@SuppressWarnings({"java:S4790"})
+@SuppressWarnings({"java:S4790", "checkstyle:SummaryJavadoc"})
 public class FileContentChecksum {
 
+    /**
+     * @deprecated This implementation is not memory-efficient with large files, use calculate(DigestInputStream digestInputStream) instead.
+     */
+    @Deprecated
     public String calculate(byte[] bytes) {
         return encodeToString(md5(bytes));
     }
--- a/src/main/java/uk/gov/hmcts/darts/datamanagement/api/impl/DataManagementApiImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/datamanagement/api/impl/DataManagementApiImpl.java
@@ -13,6 +13,7 @@ import uk.gov.hmcts.darts.datamanagement.api.DataManagementApi;
 import uk.gov.hmcts.darts.datamanagement.config.DataManagementConfiguration;
 import uk.gov.hmcts.darts.datamanagement.service.DataManagementService;
 
+import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
@@ -20,6 +21,7 @@ import java.util.UUID;
 
 @Service
 @RequiredArgsConstructor
+@SuppressWarnings("checkstyle:SummaryJavadoc")
 public class DataManagementApiImpl implements DataManagementApi {
 
     private final DataManagementService dataManagementService;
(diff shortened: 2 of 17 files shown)


src/main/java/uk/gov/hmcts/darts/datamanagement/api/impl/DataManagementApiImpl.java
```diff
@@ -40,11 +42,6 @@ public BinaryData getBlobDataFromInboundContainer(UUID blobId) {
         return dataManagementService.getBlobData(getInboundContainerName(), blobId);
     }
 
-    @Override
-    public UUID saveBlobDataToOutboundContainer(BinaryData binaryData) {
-        return dataManagementService.saveBlobData(getOutboundContainerName(), binaryData);
-    }
-
```
chrisbellingham-hmcts, line 47: Author note: unused, so removed here.

cmagowan-hmcts (approved):

hemantasharma1129 (approved):

mestebanez:
Adding this for traceability. I keep losing the link :). See [file-size-threshold](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/web/servlet/MultipartProperties.html#getFileSizeThreshold())

https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/web/servlet/MultipartProperties.html

src/main/java/uk/gov/hmcts/darts/common/util/FileContentChecksum.java
```diff
@@ -2,23 +2,30 @@
...
 
-    /**
-     * Calculates the file checksum equivalent to the Azure Blob Storage CONTENT-MD5 tag value as a base64 encoded
-     * representation of the binary MD5 hash value. This could be done in Bash e.g. openssl md5 -binary "Test
-     * Document.doc" | base64
-     *
-     * @param bytes the file content
-     * @return the calculated checksum
-     */
     public String calculate(byte[] bytes) {
```
mestebanez, line 19: I guess we should eliminate this eventually?

src/main/java/uk/gov/hmcts/darts/datamanagement/api/impl/DataManagementApiImpl.java
```diff
@@ -79,6 +76,15 @@ public void deleteBlobDataFromUnstructuredContainer(UUID blobId) throws AzureDel
         dataManagementService.deleteBlobData(getUnstructuredContainerName(), blobId);
     }
 
+    @Override
+    public UUID saveBlobDataToInboundContainer(InputStream inputStream) {
```
mestebanez, line 80: This makes sense. Maybe there is scope for some kind of general upload interface much like we did for the download ability

src/main/java/uk/gov/hmcts/darts/datamanagement/service/impl/DataManagementServiceImpl.java
```diff
@@ -66,6 +74,22 @@ public UUID saveBlobData(String containerName, BinaryData binaryData) {
...
+    @Override
+    public UUID saveBlobData(String containerName, InputStream inputStream) {
+        BlobServiceClient serviceClient = blobServiceFactory.getBlobServiceClient(dataManagementConfiguration.getBlobStorageAccountConnectionString());
+        BlobContainerClient containerClient = blobServiceFactory.getBlobContainerClient(containerName, serviceClient);
+
+        var uniqueBlobId = UUID.randomUUID();
+
+        var client = blobServiceFactory.getBlobClient(containerClient, uniqueBlobId);
+
+        var uploadOptions = new BlobParallelUploadOptions(inputStream);
```
mestebanez, line 86: Not you're concern here but when I was looking at the download operation I noticed that we were simply letting the Azure Runtime exception propagate in the case of a download failure. Should we throw some kind of darts api media exception in these cases? The reason I mention it is that I didnt know if some upstream logic wanted to handle the failure in some way i.e. retry.

src/main/java/uk/gov/hmcts/darts/common/util/FileContentChecksum.java
chrisbellingham-hmcts: Yeah good shout, I'll add `@Deprecated` to this one.

src/main/java/uk/gov/hmcts/darts/datamanagement/api/impl/DataManagementApiImpl.java
chrisbellingham-hmcts: Agree, I'll work on some tickets in this area and will make sure you're involved.

src/main/java/uk/gov/hmcts/darts/datamanagement/service/impl/DataManagementServiceImpl.java
chrisbellingham-hmcts: Yeah I think for any interaction with Azure, as a minumum we should retry at `darts-api` level. Not sure if BlobClient has that functionality built-in, or if we need to implement it ourselves. In general I think we need to better understand the failure modes around Azure. I know some initial failure mode work was done here: https://tools.hmcts.net/confluence/display/DMP/DARTS+Potential+Failure+Points.

PR #1125, opened 2024-02-26
DMP-2280_Update_GET_courthouses_to_include_region_id

mmmalik1:
[
](https://tools.hmcts.net/jira/browse/DMP-2280)

### Change description ###
Update the existing `GET /courthouses` endpoint to include a `region_id` 

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/common/controller/CourthouseApiTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/common/controller/CourthouseApiTest.java
@@ -16,6 +16,8 @@ import uk.gov.hmcts.darts.common.entity.CourthouseEntity;
 import uk.gov.hmcts.darts.common.entity.RegionEntity;
 import uk.gov.hmcts.darts.common.entity.SecurityGroupEntity;
 import uk.gov.hmcts.darts.common.entity.UserAccountEntity;
+import uk.gov.hmcts.darts.common.repository.CourthouseRepository;
+import uk.gov.hmcts.darts.common.repository.RegionRepository;
 import uk.gov.hmcts.darts.courthouse.model.ExtendedCourthouse;
 import uk.gov.hmcts.darts.testutils.IntegrationBase;
 import uk.gov.hmcts.darts.testutils.stubs.AdminUserStub;
@@ -64,6 +66,11 @@ class CourthouseApiTest extends IntegrationBase {
     private static final OffsetDateTime ORIGINAL_LAST_LOGIN_TIME = OffsetDateTime.of(2023, 10, 27, 22, 0, 0, 0, ZoneOffset.UTC);
     private static final OffsetDateTime ORIGINAL_LAST_MODIFIED_DATE_TIME = OffsetDateTime.of(2023, 10, 27, 22, 0, 0, 0, ZoneOffset.UTC);
     private static final OffsetDateTime ORIGINAL_CREATED_DATE_TIME = OffsetDateTime.of(2023, 10, 27, 22, 0, 0, 0, ZoneOffset.UTC);
+    public static final String SWANSEA_CROWN_COURT = "SWANSEA CROWN COURT";
+    public static final String LEEDS_COURT = "LEEDS";
+    public static final String MANCHESTER_COURT = "MANCHESTER";
+    public static final String WALES_REGION = "Wales";
+    public static final String NORTH_WEST_REGION = "North West";
 
     @Autowired
     private MockMvc mockMvc;
@@ -77,8 +84,14 @@ class CourthouseApiTest extends IntegrationBase {
     @MockBean
     private UserIdentity mockUserIdentity;
 
+    @Autowired
+    RegionRepository regionRepository;
+
+    @Autowired
+    CourthouseRepository courthouseRepository;
+
     @Test
-    void courthousesGet() throws Exception {
+    void adminCourthousesGet() throws Exception {
         UserAccountEntity user = adminUserStub.givenUserIsAuthorised(mockUserIdentity);
         createEnabledUserAccountEntity(user);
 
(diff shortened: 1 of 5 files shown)


src/integrationTest/java/uk/gov/hmcts/darts/common/controller/CourthouseApiTest.java
```diff
@@ -144,6 +153,39 @@ void courthousesWithRegionAndSecurityGroupsGet() throws Exception {
 
     }
 
+    @Test
+    @Transactional
+    void courthousesGetWithRegionId() throws Exception {
+        CourthouseEntity ch1 = dartsDatabase.createCourthouseUnlessExists("HAVERFORDWEST");
+
+        RegionEntity region1 = new RegionEntity();
+        region1.setRegionName("Reading");
```
chrisbellingham-hmcts, line 162: Not that it really matters, but suggest using a region name more aligned with the region names we'll actually use. See https://tools.hmcts.net/jira/browse/DMP-2367 for some reference data.

src/integrationTest/java/uk/gov/hmcts/darts/common/controller/CourthouseApiTest.java
```diff
@@ -144,6 +153,39 @@ void courthousesWithRegionAndSecurityGroupsGet() throws Exception {
...
+        regionRepository.save(region1);
+        ch1.setRegion(region1);
+
+        dartsDatabase.createCourthouseUnlessExists("SWANSEA");
+
+        CourthouseEntity ch3 = dartsDatabase.createCourthouseUnlessExists("MANCHESTER");
+        RegionEntity region3 = new RegionEntity();
+        region3.setRegionName("KNOWLE");
```
chrisbellingham-hmcts, line 170: Not that it really matters, but suggest using a region name more aligned with the region names we'll actually use. See https://tools.hmcts.net/jira/browse/DMP-2367 for some reference data.

src/integrationTest/java/uk/gov/hmcts/darts/common/controller/CourthouseApiTest.java
```diff
@@ -144,6 +153,39 @@ void courthousesWithRegionAndSecurityGroupsGet() throws Exception {
...
+        regionRepository.save(region3);
+        ch3.setRegion(region3);
+
+        MockHttpServletRequestBuilder requestBuilder = get("/courthouses")
+            .contentType(MediaType.APPLICATION_JSON_VALUE);
+        MvcResult response = mockMvc.perform(requestBuilder).andExpect(status().isOk())
+            .andExpect(jsonPath("$[0].courthouse_name", is("HAVERFORDWEST")))
+            .andExpect(jsonPath("$[0].region_id", is(notNullValue())))
```
chrisbellingham-hmcts, line 178: The test would be even more valuable if we assert that the id is the exact value that we expect, rather than simply not null. You should be able to get the id from the `RegionEntity` returned from L163.

src/integrationTest/java/uk/gov/hmcts/darts/common/controller/CourthouseApiTest.java
```diff
@@ -144,6 +153,39 @@ void courthousesWithRegionAndSecurityGroupsGet() throws Exception {
...
+            .andExpect(jsonPath("$[1].courthouse_name", is("SWANSEA")))
+            .andExpect(jsonPath("[1].region_id").doesNotExist())
+            .andExpect(jsonPath("$[2].courthouse_name", is("MANCHESTER")))
+            .andExpect(jsonPath("$[2].region_id", is(notNullValue())))
```
chrisbellingham-hmcts, line 182: As comment above.

src/integrationTest/java/uk/gov/hmcts/darts/common/controller/CourthouseApiTest.java
```diff
@@ -144,6 +153,39 @@ void courthousesWithRegionAndSecurityGroupsGet() throws Exception {
 
     }
 
+    @Test
+    @Transactional
+    void courthousesGetWithRegionId() throws Exception {
```
chrisbellingham-hmcts, line 158: Suggest renaming to `courthousesGet`, and renaming the existing `courthousesGet` on L85 to `adminCourthousesGet`.

chrisbellingham-hmcts, line 158: How does this test differ in purpose from the existing test `courthousesGetAll`?

src/main/java/uk/gov/hmcts/darts/courthouse/controller/CourthousesController.java
```diff
@@ -79,8 +79,8 @@ public ResponseEntity<Void> courthousesCourthouseIdPut(
     public ResponseEntity<List<ExtendedCourthouse>> courthousesGet(
 
     ) {
-        List<CourthouseEntity> courtHouseEntities = courthouseService.getAllCourthouses();
-        List<ExtendedCourthouse> responseEntities = mapper.mapFromListEntityToListExtendedCourthouse(courtHouseEntities);
+        List<CourthouseEntity> courthouseEntities = courthouseService.getAllCourthouses();
+        List<ExtendedCourthouse> responseEntities = courthouseService.getAllExtendedCourthouses(courthouseEntities);
```
chrisbellingham-hmcts, line 83: The method `getAllExtendedCourthouses` is named as a getter, but it has an arg. This is an unusual combination so I think this will cause some confusion to anyone casually reading this.

Suggest following a similar pattern to the other methods in this class and change it so something like `mapFromEntitiesToExtendedCourthouses`, which is a better description of what the method is doing.

src/main/java/uk/gov/hmcts/darts/courthouse/service/CourthouseServiceImpl.java
```diff
@@ -80,6 +81,26 @@ public AdminCourthouse getAdminCourtHouseById(Integer id) {
         return adminCourthouse;
     }
 
+    @Override
+    public List<ExtendedCourthouse> getAllExtendedCourthouses(List<CourthouseEntity> courthouseEntities) {
```
chrisbellingham-hmcts, line 85: I think this method is more complex than it needs to be, and I see why - by using the `mapFromListEntityToListExtendedCourthouse()` method you end up with a List of ExtendedCourthouses, where each Courthouse has no association with a Region. So you then need to reconcile that in an awkard way on lines L89-99.

To me, a simpler solution would be to use the `mapFromEntityToExtendedCourthouse()` method, which would allow you to do all the mapping needed for each courthouse before moving onto the next courthouse. This is basically the same pattern as you've already used in `UserManagementServiceImpl.getUsers()`. E.g:

```java
    @Override
    public List<ExtendedCourthouse> mapFromEntitiesToExtendedCourthouses(List<CourthouseEntity> courthouseEntities) {
        List<ExtendedCourthouse> extendedCourthouses = new ArrayList<>();

        courthouseEntities
            .forEach(courthouseEntity -> {
                ExtendedCourthouse extendedCourthouse = mapper.mapFromEntityToExtendedCourthouse(courthouseEntity);
                extendedCourthouse.setRegionId(courthouseEntity.getRegion() == null ? null : courthouseEntity.getRegion().getId());
                extendedCourthouses.add(extendedCourthouse);
            });

        return extendedCourthouses;
    }
```

src/integrationTest/java/uk/gov/hmcts/darts/common/controller/CourthouseApiTest.java
mmmalik1: Updated

src/integrationTest/java/uk/gov/hmcts/darts/common/controller/CourthouseApiTest.java
mmmalik1: Updated

src/main/java/uk/gov/hmcts/darts/courthouse/controller/CourthousesController.java
mmmalik1: updated

src/main/java/uk/gov/hmcts/darts/courthouse/service/CourthouseServiceImpl.java
mmmalik1: updated

src/integrationTest/java/uk/gov/hmcts/darts/common/controller/CourthouseApiTest.java
mmmalik1: updated

src/integrationTest/java/uk/gov/hmcts/darts/common/controller/CourthouseApiTest.java
mmmalik1: updated

src/integrationTest/java/uk/gov/hmcts/darts/common/controller/CourthouseApiTest.java
mmmalik1: updated

mmmalik1: courthouseGetAll assertsTrue by checking a court exists in a list and providing the content body only. It sets up the created and last modified date and time but does not compare the actual values that it sets.
courthouseGet sets up regions for courthouses and validates if the region_id exists and the value. This can be further extended if required.

chrisbellingham-hmcts (approved):

PR #1129, opened 2024-02-27
Dmp 2281 admin get regions

garethmorriscgi:
https://tools.hmcts.net/jira/browse/DMP-2281 

### Change description ###
Added the GET all regions endpoint to the courthouse feature

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/courthouse/controller/CourthousesController.java
+++ b/src/main/java/uk/gov/hmcts/darts/courthouse/controller/CourthousesController.java
@@ -37,7 +37,7 @@ public class CourthousesController implements CourthousesApi {
 
     private final CourthouseService courthouseService;
 
-    private final CourthouseToCourthouseEntityMapper mapper;
+    private final CourthouseToCourthouseEntityMapper courthouseMapper;
 
     private final AdminRegionToRegionEntityMapper regionMapper;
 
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/hmcts/darts/courthouse/controller/CourthousesController.java
```diff
@@ -36,6 +39,8 @@ public class CourthousesController implements CourthousesApi {
 
     private final CourthouseToCourthouseEntityMapper mapper;
```
chrisbellingham-hmcts, line 40: Since we now have two mappers, it's worth renaming this variable to `courthouseMapper` for clarity.

src/main/resources/openapi/courthouse.yaml
```diff
@@ -40,7 +40,31 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AdminRegionList'
+        '404':
+          description: Not Found
+          content:
+            application/json+problem:
+              schema:
+                $ref: './problem.yaml'
```
chrisbellingham-hmcts, line 60: We don't have any ACs around 404, is this needed?

chrisbellingham-hmcts (approved):

src/main/java/uk/gov/hmcts/darts/courthouse/controller/CourthousesController.java
garethmorriscgi: Agreed - I had already done this for the unit test. Updated controller as suggested.

src/main/resources/openapi/courthouse.yaml
garethmorriscgi: Removed this from api and therefore the try/catch/throw 404 from the controller

PR #1131, opened 2024-02-27
dmp-1072 annotation upload audit

JasonPittockCGI:
https://tools.hmcts.net/jira/browse/DMP-1072

DMP-1072 Introduce annotation upload audit event

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/darts/annotation/service/impl/AnnotationUploadServiceImpl.java
```diff
@@ -57,6 +64,8 @@ public Integer upload(MultipartFile multipartFile, Annotation annotation) {
                 inboundExternalObjectDirectory,
                 unstructuredExternalObjectDirectory,
                 annotation.getHearingId());
+            final var hearing = hearingRepository.findById(annotation.getHearingId());
+            hearing.ifPresent(hearingEntity -> auditApi.recordAudit(IMPORT_ANNOTATION, userIdentity.getUserAccount(), hearingEntity.getCourtCase()));
```
tonymort, line 68: Can recordAudit throw an exception? If so it might be a better pattern move it after the try catch block.

JasonPittockCGI: Hi Tony. It's included in the try so the audit is only recorded if the annotation upload is successful. If it is outside the catch block, the upload could throw an exception, yet the audit would still be recorded. Please advise if this is the correct behaviour.

tonymort (approved):

PR #1137, opened 2024-02-27
DMP-2003: Added logging when get cases requested by dar pc

tonymort:
Added logging when get cases requested by dar pc

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
@@ -66,6 +66,7 @@ public class CaseServiceImpl implements CaseService {
     @Override
     @Transactional
     public List<ScheduledCase> getHearings(GetCasesRequest request) {
+        logApi.casesRequestedByDarPc(request);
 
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
```diff
@@ -71,8 +73,12 @@ public List<ScheduledCase> getHearings(GetCasesRequest request) {
             request.getDate()
         );
         createCourtroomIfMissing(hearings, request);
-        return casesMapper.mapToScheduledCases(hearings);
 
+        List<ScheduledCase> scheduledCases = casesMapper.mapToScheduledCases(hearings);
+
+        logApi.casesRequestedByDarPc(request);
```
davet1985, line 79: Maybe log at the start, in case an exception happens during the mapping?

davet1985 (commented):
One question

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
tonymort: Yeah, was thinking about this wrong

davet1985 (approved):

PR #1138, opened 2024-02-27
Dmp 1910 download annotation facade

johnrace0:
https://tools.hmcts.net/jira/browse/DMP-1910

### Change description ###
Download Annotation - Fetch from ODS(Unstructured) or DETS or ARM

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/darts/annotation/service/impl/AnnotationDownloadServiceImpl.java
```diff
@@ -44,14 +48,29 @@ public AnnotationResponseDto downloadAnnotationDoc(Integer annotationId, Integer
...
-        final ExternalObjectDirectoryEntity externalObjectDirectoryEntity = eodDir.get(0);
+        final ExternalObjectDirectoryEntity latestExternalObjectDirectoryEntity = eodDir.get(0);
+
+        var downloadableExternalObjectDirectories = DownloadableExternalObjectDirectories
+            .getFileBasedDownload(eodDir);
+
+        dataManagementFacade.getDataFromUnstructuredArmAndDetsBlobs(downloadableExternalObjectDirectories);
+        var downloadResponseMetaData = downloadableExternalObjectDirectories.getResponse();
 
-        final InputStreamResource blobStream = annotationDataManagement.download(externalObjectDirectoryEntity);
```
davet1985, line 49: Does this make `AnnotationDataManagement.download` redundant? Would the call out to the facade not be better in there?

johnrace0: yes - it can go in there - I don't think this was here when i wrote my original changes

src/main/java/uk/gov/hmcts/darts/annotation/service/impl/AnnotationDownloadServiceImpl.java
```diff
@@ -44,14 +43,15 @@ public AnnotationResponseDto downloadAnnotationDoc(Integer annotationId, Integer
             throw new DartsApiException(INVALID_ANNOTATIONID_OR_ANNOTATION_DOCUMENTID);
         }
 
-        final ExternalObjectDirectoryEntity externalObjectDirectoryEntity = eodDir.get(0);
+        final ExternalObjectDirectoryEntity latestExternalObjectDirectoryEntity = eodDir.get(0);
 
-        final InputStreamResource blobStream = annotationDataManagement.download(externalObjectDirectoryEntity);
+        final InputStreamResource blobStream = annotationDataManagement.download(latestExternalObjectDirectoryEntity);
```
richardbull, line 48: If you compare with Transcription download - you'll notice that we actually send all the external_object_directory 's to the Facade to do the "heavy lifting". This was the design that @hemantasharma1129 wanted as I first had something similar when I started a similar integration in [DMP-656](https://tools.hmcts.net/jira/browse/DMP-656)

src/main/java/uk/gov/hmcts/darts/annotation/service/impl/AnnotationDownloadServiceImpl.java
```diff
@@ -44,14 +43,15 @@ public AnnotationResponseDto downloadAnnotationDoc(Integer annotationId, Integer
...
 
         return AnnotationResponseDto.builder()
             .resource(blobStream)
-            .fileName(externalObjectDirectoryEntity.getAnnotationDocumentEntity().getFileName())
-            .externalLocation(externalObjectDirectoryEntity.getExternalLocation())
+            .fileName(latestExternalObjectDirectoryEntity.getAnnotationDocumentEntity().getFileName())
+            .externalLocation(latestExternalObjectDirectoryEntity.getExternalLocation())
```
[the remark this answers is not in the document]
johnrace0: is that the same for the filename as well?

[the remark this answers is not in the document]
richardbull: Just the external_location header. We still want the filename as that should be used in the attachment download, so we get the same named file
e.g. ` content-disposition: attachment; filename="Test Document.docx"`

src/main/java/uk/gov/hmcts/darts/annotation/service/impl/AnnotationDataManagement.java
```diff
@@ -62,7 +62,7 @@ public InputStreamResource download(ExternalObjectDirectoryEntity externalObject
             return new InputStreamResource(downloadResponseMetaData.getInputStream());
         } catch (IOException e) {
             log.error("Failed to download annotation document {}",
```
richardbull, line 64: Please can you check this error message - we don't know what EOD was used.
So I think it makes sense for it just to be the `annotation_document` ID here instead?

src/main/java/uk/gov/hmcts/darts/annotation/service/impl/AnnotationDownloadServiceImpl.java
```diff
@@ -45,12 +45,11 @@ public AnnotationResponseDto downloadAnnotationDoc(Integer annotationId, Integer
 
         final ExternalObjectDirectoryEntity latestExternalObjectDirectoryEntity = eodDir.get(0);
 
-        final InputStreamResource blobStream = annotationDataManagement.download(latestExternalObjectDirectoryEntity);
+        final InputStreamResource blobStream = annotationDataManagement.download(eodDir);
 
         return AnnotationResponseDto.builder()
             .resource(blobStream)
             .fileName(latestExternalObjectDirectoryEntity.getAnnotationDocumentEntity().getFileName())
```
richardbull, line 52: I know we should have the annotation_document from one of the external_object_directory links, but it feels better to me to use a `findById` here with the `annotationDocumentId` param.
So then we wouldn't need `latestExternalObjectDirectoryEntity` on line 46?

I'll let you decide if you agree on doing that here :)

src/main/java/uk/gov/hmcts/darts/annotation/service/impl/AnnotationDataManagement.java
johnrace0: changed to annotation_doc ID

src/main/java/uk/gov/hmcts/darts/annotation/service/impl/AnnotationDownloadServiceImpl.java
johnrace0: yeah - the EODs are returned in descending order so the first one is the latest. the database supports multiple EOD with same annotation id. I've decided to leave as is :-)

richardbull (approved):

PR #1141, opened 2024-02-28
dmp 1919 change to single query

chrisdavecm:
https://tools.hmcts.net/jira/browse/DMP-1919

### Change description ###
Change the query for processing

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -134,6 +134,7 @@ darts:
         arm-minimum:
           weeks: 30
     pending-retention-days: 7
+    inbound-to-unstructured-limit: 100
   daily-list:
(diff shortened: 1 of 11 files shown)


src/main/resources/application.yaml
```diff
@@ -134,6 +134,7 @@ darts:
         arm-minimum:
           weeks: 30
     pending-retention-days: 7
+    inbound-to-unstructured-limit: 100
```
davet1985, line 137: Doesn't look like this is used at the moment?

chrisdavecm: It is now, had to solve a conflict to get it working

PR #1145, opened 2024-02-28
Using standard error patterns for courthouse not found

davet1985:

https://tools.hmcts.net/jira/browse/DMP-2282

### Change description ###

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/courthouse/controller/CourthousesController.java
+++ b/src/main/java/uk/gov/hmcts/darts/courthouse/controller/CourthousesController.java
@@ -7,13 +7,14 @@ import jakarta.persistence.EntityNotFoundException;
 import lombok.RequiredArgsConstructor;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
-import org.springframework.orm.jpa.JpaObjectRetrievalFailureException;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestBody;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/darts/courthouse/controller/CourthousesController.java
```diff
@@ -61,7 +62,7 @@ public ResponseEntity<AdminCourthouse> adminCourthousesCourthouseIdGet(
 
             return new ResponseEntity<>(adminCourthouse, HttpStatus.OK);
         } catch (EntityNotFoundException | JpaObjectRetrievalFailureException | IllegalStateException e) {
-            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+            throw new DartsApiException(CourthouseApiError.COURTHOUSE_NOT_FOUND);
```
jackmaloney, line 65: Consider adding a unit test to cover these exceptions, and adding a log.error message before each one with the 'e' variable passed in so the stack trace can be seen, otherwise this information will be lost. We saw similar issue in model court testing last week (elsewhere in code) and was a nightmare to work out what had actually caused the issue. https://tools.hmcts.net/jira/browse/DMP-2433 has been raised already to address but if we're changing this code now, makes sense to add it.

davet1985: I've checked and amended the exception handling here and added a couple of unit tests.

jackmaloney (approved):

PR #1157, opened 2024-03-01
DMP-2403 Update security schema: Super Admin

richardbull:

[DMP-2403](https://tools.hmcts.net/jira/browse/DMP-2403)
[DMP-1072](https://tools.hmcts.net/jira/browse/DMP-1072)

### Change description ###

* Update security schema: Super Admin
* Update standing data / flyway_schema_history: version 1.279 update audit activity

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/retention/controller/RetentionController.java
+++ b/src/main/java/uk/gov/hmcts/darts/retention/controller/RetentionController.java
@@ -18,10 +18,10 @@ import uk.gov.hmcts.darts.retentions.model.PostRetentionResponse;
 import java.util.List;
 
 import static uk.gov.hmcts.darts.authorisation.constants.AuthorisationConstants.SECURITY_SCHEMES_BEARER_AUTH;
-import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.ADMIN;
 import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.APPROVER;
 import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.JUDGE;
 import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.REQUESTER;
+import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.SUPER_ADMIN;
 import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.TRANSCRIBER;
 import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.TRANSLATION_QA;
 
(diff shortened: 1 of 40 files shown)


chrisbellingham-hmcts (commented):
Hey Richard. Should this PR also include checking that `SUPER_ADMIN` is authorised to use the endpoints corresponding to the matrix within https://tools.hmcts.net/jira/browse/DMP-2403?

richardbull:
> Hey Richard. Should this PR also include checking that `SUPER_ADMIN` is authorised to use the endpoints corresponding to the matrix within https://tools.hmcts.net/jira/browse/DMP-2403?

@chrisbellingham-hmcts Not sure, I added the `security_role_permission_ae WHERE rol_id = 11;` where I tried to interpret that matrix, and the existing ADMIN security_role was renamed to be SUPER_ADMIN.
I think there was talk in refinement about the BA's identifying the endpoints for the Authorisation if they are missing, so there may be some more work / checking required in that area.  Not sure where we need to draw the line though to get this ticket done?

src/main/java/uk/gov/hmcts/darts/retention/controller/RetentionController.java
```diff
@@ -33,20 +33,22 @@ public class RetentionController implements RetentionApi {
...
     @Override
     public ResponseEntity<List<GetCaseRetentionsResponse>> retentionsGet(Integer caseId) {
         return new ResponseEntity<>(retentionService.getCaseRetentions(caseId), HttpStatus.OK);
     }
 
     @SecurityRequirement(name = SECURITY_SCHEMES_BEARER_AUTH)
     @Authorisation(contextId = ContextIdEnum.CASE_ID, bodyAuthorisation = true,
-            securityRoles = {JUDGE, REQUESTER, APPROVER, TRANSCRIBER, TRANSLATION_QA, ADMIN},
-            globalAccessSecurityRoles = {JUDGE, ADMIN})
+        securityRoles = {JUDGE, REQUESTER, APPROVER, TRANSCRIBER},
```
johnrace0, line 44: is SUPER_ADMIN not relevant here?

hemantasharma1129:
I can see only the renaming ROLE changes are made but the controller layer Authorisation should use the new SUPER_ADMIN as well. Please make those changes as well.

src/main/java/uk/gov/hmcts/darts/retention/controller/RetentionController.java
richardbull: My understanding of the controller authorisation is that the `securityRoles` list the roles that it is necessary to check against a `contextId` e.g. case_id param

According to our Authorisation business logic (DARTS All user permissions.xlsx), "Super Admin" Can access all courthouses. So we wouldn't expect the security_group to have any association with courthouse.
That's what the meaning behind `global_access	true` on the security_group is about.
So the global role is present in the other list that Karen added to the annotation called `globalAccessSecurityRoles`.

`SELECT grp_id, cth_id FROM security_group_courthouse_ae WHERE grp_id = 1;`

security_group
--------------
grp_id	1
rol_id	11
group_name	SUPER_ADMIN
global_access	true

richardbull:
> @richardbull I can see only renaming the ROLE changes are made but the controller layer Authorisation should use the new SUPER_ADMIN role as well. Please make those changes.

Sure, I'll have a look and try to interpret further.

PR #1160, opened 2024-03-01
Removing default dar pc notification url

tonymort:
Removing default dar pc notification url.  The url will now come only from the node_register table.

```
[ ] Yes
[x] No
```


davet1985 (commented):
Do we not need that for the local env?

davet1985:
> Do we not need that for the local env?

davet1985 (approved):

chrisdavecm (approved):

PR #1164, opened 2024-03-01
DMP-2283 modified admin/security-groups to handle option filter crite…

garethmorriscgi:
…ria (role ids, courthouse id)

https://tools.hmcts.net/jira/browse/DMP-2283 

### Change description ###
Update the existing `GET /admin/security-groups` endpoint to accept optional query (filter) parameters role_ids and courthouse_id

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/integrationTest/resources/tests/usermanagement/SecurityGroupControllerGetTest/securityGroupsGetEndpointAllReturned.json
+++ b/src/integrationTest/resources/tests/usermanagement/SecurityGroupControllerGetTest/securityGroupsGetEndpointAllReturned.json
@@ -1,0 +1,13 @@
+[
+  {"id":-17,"security_role_id":10,"global_access":true,"display_state":true,"courthouse_ids":[],"name":"Mid Tier Group"},
+  {"id":-16,"security_role_id":9,"global_access":true,"display_state":true,"courthouse_ids":[],"name":"Dar Pc Group"},
+  {"id":-15,"security_role_id":8,"global_access":true,"display_state":true,"courthouse_ids":[],"name":"Cpp Group"},
+  {"id":-14,"security_role_id":7,"global_access":true,"display_state":true,"courthouse_ids":[],"name":"Xhibit Group"},
+  {"id":-6,"security_role_id":6,"global_access":true,"display_state":true,"courthouse_ids":[],"name":"Test RCJ Appeals"},
+  {"id":-5,"security_role_id":5,"global_access":false,"display_state":true,"courthouse_ids":[],"name":"Test Language Shop"},
+  {"id":-4,"security_role_id":4,"global_access":false,"display_state":true,"courthouse_ids":[],"name":"Test Transcriber"},
+  {"id":-3,"security_role_id":3,"global_access":false,"display_state":true,"courthouse_ids":[],"name":"Test Judge"},
+  {"id":-2,"security_role_id":2,"global_access":false,"display_state":true,"courthouse_ids":[],"name":"Test Requestor"},
+  {"id":-1,"security_role_id":1,"global_access":false,"display_state":true,"courthouse_ids":[],"name":"Test Approver"},
+  {"id":1,"security_role_id":11,"global_access":true,"display_state":true,"courthouse_ids":[],"name":"ADMIN"}
+]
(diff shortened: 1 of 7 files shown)


src/main/resources/openapi/usermanagement.yaml
```diff
@@ -356,6 +364,11 @@ components:
     RoleDisplayState:
       type: boolean
 
+    RoleIds:
+      type: array
+      items:
+        type: integer
```
johnrace0, line 370: or '#/components/schemas/RoleId' ?

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/SecurityGroupServiceImpl.java
```diff
@@ -47,12 +48,39 @@ public SecurityGroupWithIdAndRole createSecurityGroup(SecurityGroup securityGrou
         return securityGroupWithIdAndRole;
     }
 
-    public List<SecurityGroupWithIdAndRole> getSecurityGroups() {
+    public List<SecurityGroupWithIdAndRole> getSecurityGroups(List<Integer> roleIds, Integer courthouseId) {
         List<SecurityGroupEntity> securityGroupEntities = securityGroupRepository.findAll();
```
richardbull, line 52: I'm wondering if we should be trying to do a custom finder here on the `SecurityGroupRepository`?
Probably one to confirm with @hemantasharma1129 / @chrisbellingham-hmcts

chrisbellingham-hmcts: The approach we've taken elsewhere is to construct the DB query dynamically, based on the provided filter params (e.g `UserSearchQueryImpl.java`). I think this approach is a good one for large datasets, albeit at the cost of some complexity.

However since the `security_group` table will be relatievly small (within the order of hundreds of records), the findAll approach is probably good enough. There's a benefit to readability in doing the filtering after-the-fact with the java streams api. Let's revisit if performance testing tells a different story though.

src/main/resources/openapi/usermanagement.yaml
chrisbellingham-hmcts: An alternative approach is to define an array type in-line within the `parameters.schema` directive, and `$ref` the `RoleId` type:

```yaml
  /admin/security-groups:
    get:
      summary: Get security groups
      parameters:
        - in: query
          name: role_ids
          description: Filter by list of role ids
          style: form
          explode: false
          schema:
            type: array
            items:
              $ref: '#/components/schemas/RoleId'
```

The general benefit is we don't clutter the `components.schema` namespace with lots of `FooList` types, and we get to reuse the existing `RoleId` definition.

chrisbellingham-hmcts (approved):

src/integrationTest/resources/tests/usermanagement/SecurityGroupControllerGetTest/securityGroupsGetEndpointAllReturned.json
```diff
@@ -0,0 +1,13 @@
...
+  {"id":-15,"security_role_id":8,"global_access":true,"display_state":true,"courthouse_ids":[],"name":"Cpp Group"},
+  {"id":-14,"security_role_id":7,"global_access":true,"display_state":true,"courthouse_ids":[],"name":"Xhibit Group"},
+  {"id":-6,"security_role_id":6,"global_access":true,"display_state":true,"courthouse_ids":[],"name":"Test RCJ Appeals"},
+  {"id":-5,"security_role_id":5,"global_access":false,"display_state":true,"courthouse_ids":[],"name":"Test Language Shop"},
+  {"id":-4,"security_role_id":4,"global_access":false,"display_state":true,"courthouse_ids":[],"name":"Test Transcriber"},
+  {"id":-3,"security_role_id":3,"global_access":false,"display_state":true,"courthouse_ids":[],"name":"Test Judge"},
+  {"id":-2,"security_role_id":2,"global_access":false,"display_state":true,"courthouse_ids":[],"name":"Test Requestor"},
+  {"id":-1,"security_role_id":1,"global_access":false,"display_state":true,"courthouse_ids":[],"name":"Test Approver"},
+  {"id":1,"security_role_id":11,"global_access":true,"display_state":true,"courthouse_ids":[],"name":"ADMIN"}
```
richardbull, line 12: Can you check this group please - I'd expect it to be `SECURITY_ADMIN` now after [DMP-2403](https://tools.hmcts.net/jira/browse/DMP-2403)

richardbull (approved):

PR #1170, opened 2024-03-04
DMP-2367 Add Region data

JasonPittockCGI:
DMP-2367 Add region data

[### JIRA link (if applicable) ###](https://tools.hmcts.net/jira/browse/DMP-2367)

### Change description ###

```
[ ] Yes
[ x] No
```


src/main/resources/db/migration/common/V1_280__add_regions.sql
```diff
@@ -0,0 +1,9 @@
+INSERT INTO region (reg_id, region_name) VALUES(1, 'London');
```
mario-paniccia, line 1: this file's version needs to be bumped up now, we have files on master with versions past this one

src/main/resources/db/migration/common/V1_280__add_regions.sql
```diff
@@ -0,0 +1,9 @@
...
+INSERT INTO region (reg_id, region_name) VALUES(2, 'South East');
+INSERT INTO region (reg_id, region_name) VALUES(3, 'South West');
+INSERT INTO region (reg_id, region_name) VALUES(4, 'Midlands');
+INSERT INTO region (reg_id, region_name) VALUES(5, 'North East');
+INSERT INTO region (reg_id, region_name) VALUES(6, 'North West');
+INSERT INTO region (reg_id, region_name) VALUES(7, 'Wales');
```
mario-paniccia, line 7: on staging DB we already have the following:
-11  Wales
-10  London
I wonder whether we want to do some cleanup there

src/main/resources/db/migration/common/V1_281__courthouse_region.sql
```diff
@@ -0,0 +1,5 @@
+INSERT INTO courthouse_region_ae (cth_id, reg_id) VALUES (2,5);
+INSERT INTO courthouse_region_ae (cth_id, reg_id) VALUES (3,6);
+INSERT INTO courthouse_region_ae (cth_id, reg_id) VALUES (1,2);
+
+ALTER SEQUENCE aut_seq RESTART WITH 4;
```
richardbull, line 5: Think this can be removed as it appears to be `automated_task` sequence

src/main/resources/db/migration/common/V1_281__courthouse_region.sql
```diff
@@ -0,0 +1,5 @@
+INSERT INTO courthouse_region_ae (cth_id, reg_id) VALUES (2,5);
+INSERT INTO courthouse_region_ae (cth_id, reg_id) VALUES (3,6);
+INSERT INTO courthouse_region_ae (cth_id, reg_id) VALUES (1,2);
```
richardbull, line 3: I think we should check this region value with  @naivinhta - I would expect Swansea to be South West, so `reg_id=3`?

src/main/resources/db/migration/common/V1_280__add_regions.sql
```diff
@@ -0,0 +1,9 @@
...
+
+ALTER SEQUENCE aut_seq RESTART WITH 8;
```
richardbull, line 9: This should be `reg_seq` - although it's odd that we have a sequence for a standing data table.
But hey ho - may as well bump it whilst it's here

src/main/resources/db/migration/common/V1_280__add_regions.sql
JasonPittockCGI: I've bumped them up again.

src/main/resources/db/migration/common/V1_280__add_regions.sql
JasonPittockCGI: Hia Richard. Believe their are issues with the sequences not being set between integration tests when the databases are cleared.

src/main/resources/db/migration/common/V1_281__courthouse_region.sql
JasonPittockCGI: Checking ...

src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/DartsDatabaseStub.java
```diff
@@ -205,6 +207,7 @@ public void clearDatabaseInThisOrder() {
         // securityGroupRepository.deleteAll(securityGroupBin);
         // securityGroupBin.clear();
         courthouseRepository.deleteAll();
+        regionRepository.deleteAll();
```
richardbull, line 210: I don't think we should delete standing data: region
What do you think here @tonymort ?

tonymort: Yeah, that is the general pattern, theres example of where we do this however which is wrong and should be fixed. Going forward however though I agree lets not introduce more this.  Means we should create/delete new regions in integration tests and only rely on the standing data.

JasonPittockCGI: I could see the courthouses were deleted and took it as similar standing data. I can remove and push again.

JasonPittockCGI: Removed, but had knock on issues. So I added again. We are probably better doing this consistently and removing the clearing of all standing data together on a seperate ticket.

src/main/resources/db/migration/common/V1_280__add_regions.sql
JasonPittockCGI: Hey Mario. Yes. Good point, the data in staging could do with a cleanup, I assume this would be on a separate ticket and would require a fair amount of collaboration with the testers.

src/main/resources/db/migration/common/V1_281__courthouse_region.sql
JasonPittockCGI: Nai updated ticket and the sql scripts have been changed to reflect this.

src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/DartsDatabaseStub.java
richardbull: I'm assuming @tonymort means to look at your integration test - and you would basically expect to see the 7 regions returned in the GET request (since we wouldn't be deleting them).
They are standing data in the sense that we have purposefully loaded them via our script and wouldn't expect them to be added by a service (endpoint).

richardbull (approved):

tonymort (approved):

mario-paniccia (approved):

PR #1195, opened 2024-03-11
Dmp 1661 integration tests ats download

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-1661

### Change description ###

Investigate integration and or functional tests to cover ffmpeg commands

```
[ ] Yes
[x ] No
```


src/functionalTest/java/uk/gov/hmcts/darts/audio/AudioTransformationServiceFunctionalTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+        String audioRequest = """
+            {
+              "hearing_id": <<hearingId>>,
+              "requestor": 0,
+              "start_time": "2024-03-11T10:15:00.000Z",
+              "end_time": "2024-03-11T10:20:00.000Z",
+              "request_type": "DOWNLOAD"
+            }
+            """;
```
chrisbellingham-hmcts, line 63: I understand that one of the challenges with the functional test approach is that when it comes to run the test against staging (which happens upon merging into master), creating a media request may result in KEDA picking the request up and processing it before the test has a chance to.

Could a solution to that be:
- Create a new endpoint in `TestSupportController` that allows the func test to create a media request with a non-`OPEN` `status`. E.g. `PROCESSING`.
- Invoke `/functional-tests/handleKedaInvocationForMediaRequests/` with this requestId.

I think this would avoid the issue, as KEDA will only detect `OPEN` requests. The `processAudioRequest()` method doesn't care what the status is, it will overwite it to `PROCESSING` anyway.

A bit hacky, but might be an option to explore?

karen-hedges: This was the reason that this was not continued past this point as there was a potential race condition. Again its all got a bit hacky and then to actually do some verification of handleKedaInvocationForMediaRequests would require to either add a new functional test endpoint that uses the integration database tools to test the db is as expected or use current endpoints that may not give you all of what you want to verify.

karen-hedges:
This contains an experiment to see if functional tests for ATS can be done but now that integration tests are possible, this is redundant

PR #1219, opened 2024-03-15
Dmp 2204 auto task close old cases

chrisdavecm:
https://tools.hmcts.net/jira/browse/DMP-2204

### Change description ###
New automated task to close old cases

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/cases/service/impl/CloseOldCasesProcessorImpl.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private final EventRepository eventRepository;
+
+    private final MediaRepository mediaRepository;
+
+
+    @Transactional
+    @Override
+    public void closeCases() {
+        List<CourtCaseEntity> courtCaseEntityList = caseRepository.findOpenCasesToClose(OffsetDateTime.now().minusYears(6));
```
jackmaloney, line 39: Can we make the 6 years configurable? That may change in future.

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CloseOldCasesProcessorImpl.java
```diff
@@ -0,0 +1,89 @@
...
+    private void closeCase(CourtCaseEntity courtCase) {
+        List<EventEntity> eventList = new ArrayList<>();
+        for (HearingEntity hearingEntity: courtCase.getHearings()) {
+            eventList.addAll(hearingEntity.getEventList());
+        }
+        if (eventList.size() > 1) {
+            eventList.sort(Comparator.comparing(EventEntity::getCreatedDateTime).reversed());
+            //find latest closed event
+            Optional<EventEntity> closedEvent =
+                eventList.stream().filter(eventEntity -> eventEntity.getEventType().getEventName().equals("Case closed")).findFirst();
```
jackmaloney, line 53: Hardcoding this to "case closed" doesn't seem like the best thing to do. Firstly, this is an XHIBIT even type, and so they could change the name of that event at any time. Secondly, CPP has their own case closed event called "Archive case", and this logic wouldn't work for that scenario. Happy to talk this through and try and come up with a solution that is future proof if you need.

jackmaloney (approved):

PR #1225, opened 2024-03-18
DMP-2339: Flushing so that we get the latest last modified timestamp

tonymort:
Flushing so that we get the latest last modified timestamp

```
[ ] Yes
[x] No
```


davet1985 (commented):
🚽

davet1985 (approved):

PR #1230, opened 2024-03-19
dmp 2528 bug fix for inbound to unstructured

chrisdavecm:
https://tools.hmcts.net/jira/browse/DMP-2528

### Change description ###
bug fix for zero byte file

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/datamanagement/service/impl/InboundToUnstructuredProcessorSingleElementImpl.java
```diff
@@ -226,6 +227,9 @@ private void performValidation(
         if (fileSize > maxFileSize) {
             unstructured.setStatus(getStatus(FAILURE_FILE_SIZE_CHECK_FAILED));
         }
+        if (0 == fileSize) {
+            unstructured.setStatus(getStatus(FAILURE_EMPTY_FILE));
+        }
```
chrisbellingham-hmcts, line 232: This is more of a question on the existing structure of the `performValidation()` method. The way the conditionals are structured essentially makes each condition fall-through. So if a file is of the wrong format, and has an unexpected size then `unstructured.status` is being overwritten, and all we're only left with the last status value set.

I find that a bit counterintuitive - is that an intentional design choice, or should the method return on the first validation failure?

Or if we're happy to only set the last failure reason, should we at least log any proceeding failure reasons?

chrisbellingham-hmcts (approved):

src/main/resources/db/migration/common/V1_287__failure_empty_file.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO object_record_status (ors_id, ors_description) VALUES (20, 'Empty File Failed');
```
richardbull, line 1: FYI I think there's an expectation from @barry-jones50410 / DBA team for us to maintain a record of standing data records in `resources/db/reference/modernised_darts_standing_data.sql`?

PR #1231, opened 2024-03-19
DMP-2030 fix retention policy table

JasonPittockCGI:
DMP-2030 Implement GET /admin/retention-policy-types

fix retention policy table

[### JIRA link (if applicable) ###](https://tools.hmcts.net/jira/browse/DMP-2530)

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/resources/db/migration/common/V1_286__update_external_location.sql
+++ b/src/main/resources/db/migration/common/V1_286__update_external_location.sql
@@ -1,0 +1,2 @@
+-- replace policy_end_ts with null
+UPDATE darts.retention_policy_type SET policy_end_ts = null Where rpt_id = 1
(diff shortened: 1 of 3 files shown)


src/main/resources/db/migration/common/V1_286__update_external_location.sql
```diff
@@ -0,0 +1,2 @@
+-- replace policy_end_ts with null
+UPDATE darts.retention_policy_type SET policy_end_ts = null Where rpt_id = 1
```
richardbull, line 2: I think you just need to revert what you did earlier in V1_260__retention_policies.sql

See commit on my branch - 
https://github.com/hmcts/darts-api/commits/DMP-2517-adminGetSecurityGroup/src/main/resources/db/migration/common/V1_260__retention_policies.sql

Don't think you need to add another script if the previous didn't get applied to any environments because of the version mismatch as the `policy_end_ts` is already null?

JasonPittockCGI: Hia Richard. Yeah Agree. Been on with Chris B. Will push again.

PR #1235, opened 2024-03-20
DMP-2374 only return authorised courthouses in get courthouses endpoint

mmmalik1:
https://tools.hmcts.net/jira/browse/DMP-2374

### Change description ###
The endpoint to get all courthouses GET /courthouses currently just returns all courthouses, this should be changed to only returns the courthouses available to the current user based on their permissions.

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/darts/courthouse/service/CourthouseServiceImpl.java
```diff
@@ -115,7 +121,48 @@ public List<ExtendedCourthouse> mapFromEntitiesToExtendedCourthouses(List<Courth
 
     @Override
     public List<CourthouseEntity> getAllCourthouses() {
-        return courthouseRepository.findAll();
+        List<Integer> userCourthouseIds = new ArrayList<>();
+
+        UserAccountEntity userAccount = authorisationApi.getCurrentUser();
```
cmagowan-hmcts, line 126: shouls this not just use the `authorisationApi.getListOfCourthouseIdsUserHasAccessTo()` method?
then make a call to `courthouseRepository` with those courthouseIds?

richardbull: Although looking at the current finder `UserRolesCourthousesRepository#findAllCourthouseIdsByUserAccount` 
It feels a bit strange to me - I think we could have a return type of List<CourthouseEntity> with query having a `SELECT DISTINCT urc.courthouse`
Since if you look at the `UserRolesCourthousesEntity#courthouse` it will be currently eagerly fetching the `CourthouseEntity` anyway?

cmagowan-hmcts: when `UserRolesCourthousesRepository#findAllCourthouseIdsByUserAccount` was first made, i made it for the AdvancedSearch, and i only needed the courthouseIds as it was being used in the search, so there was no need to bring back the full courthouse object as that would mean further calls to the database. 
AdvancedSearch has now been enhanced to use a different view, so this method is no longer used by it, so i guess can be modified as required.

mmmalik1: Code and tests updated

src/main/java/uk/gov/hmcts/darts/courthouse/service/CourthouseServiceImpl.java
```diff
@@ -115,7 +122,39 @@ public List<ExtendedCourthouse> mapFromEntitiesToExtendedCourthouses(List<Courth
...
+
+    }
+
+    private List<CourthouseEntity> getCourthouseEntities() {
+        List<CourthouseEntity> userCourthouseEntities = new ArrayList<>();
+        List<Integer> courthouseIdsUserHasAccessTo = authorisationApi.getListOfCourthouseIdsUserHasAccessTo();
+
+
+        for (Integer courthouseId : courthouseIdsUserHasAccessTo) {
+            Optional<CourthouseEntity> courthouseEntity = courthouseRepository.findById(courthouseId);
```
cmagowan-hmcts, line 154: as mentioned earlier, we should be creating a method in courthouseRepository called findByIdIn

src/main/java/uk/gov/hmcts/darts/courthouse/service/CourthouseServiceImpl.java
cmagowan-hmcts: none of the code from lines 126 to 142 is needed.

mmmalik1: In the case where Admin or Judge is provided this coded is required.

mmmalik1: code updated as per suggestion

cmagowan-hmcts (approved):

src/main/java/uk/gov/hmcts/darts/common/repository/CourthouseRepository.java
```diff
@@ -34,4 +34,6 @@ AND securityRole.id IN (:roleIds)
 
     boolean existsByDisplayNameIgnoreCaseAndIdNot(String name, Integer id);
 
+    List<CourthouseEntity> findByIdIn(List<Integer> courthouseIds);
```
richardbull, line 37: Don't think we need this if we already have `findAllById` from ListCrudRepository?

But as per earlier comment, the existing view entity could either be simplified to just be plain old Integer columns - no ManyToOne joins needed if you are going to use a findAllById here or changing the query select and return type 
`uk.gov.hmcts.darts.common.repository.UserRolesCourthousesRepository#findAllCourthouseIdsByUserAccount`

PR #1244, opened 2024-03-22
DMP-2537 PATCH /admin/security-groups/{security_group_id}

chrisdavecm:
https://tools.hmcts.net/jira/browse/DMP-2537

### Change description ###
Patch security group

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/SecurityGroupServiceImpl.java
```diff
@@ -82,6 +96,87 @@ public List<SecurityGroupWithIdAndRole> getSecurityGroups(List<Integer> roleIds,
...
+        } else {
+            //throw a 404 not found
+            throw new DartsApiException(
+                UserManagementError.SECURITY_GROUP_NOT_FOUND,
+                String.format("Security group id %d not found", securityGroupId));
+        }
+
+    }
+
+    protected void updateSecurityGroupEntity(SecurityGroupPatch securityGroupPatch, SecurityGroupEntity securityGroupEntity) {
```
richardbull, line 119: Can this be made package-private for use in the `SecurityGroupServiceImplTest`?

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/SecurityGroupServiceImpl.java
```diff
@@ -82,6 +96,87 @@ public List<SecurityGroupWithIdAndRole> getSecurityGroups(List<Integer> roleIds,
...
+        validate(securityGroupPatch, securityGroupEntity);
+
+        String name = securityGroupPatch.getName();
+        String displayName = securityGroupPatch.getDisplayName();
+
+        if (name != null) {
```
richardbull, line 125: Wondering for these updates `name / displayName / description` - does it make sense to allow blank here?
I guess an admin user could force the update each time just adding blank spaces to name if they wanted to save the record.
I think we should use `StringUtils.isNotBlank` - what do you think @chrisbellingham-hmcts ?

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/SecurityGroupServiceImpl.java
```diff
@@ -82,6 +96,87 @@ public List<SecurityGroupWithIdAndRole> getSecurityGroups(List<Integer> roleIds,
...
+                Optional<CourthouseEntity> courthouseEntity = courthouseRepository.findById(courthouseId);
+                if (courthouseEntity.isPresent()) {
+                    courthouseEntities.add(courthouseEntity.get());
+                } else {
+                    throw new DartsApiException(
+                        UserManagementError.COURTHOUSE_NOT_FOUND,
+                        String.format("Courthouse id %d not found", courthouseId));
+                }
+            }
+            securityGroupEntity.setCourthouseEntities(courthouseEntities);
```
richardbull, line 148: Looking at the ManyToMany association here, we seem to have another join table definition `CourthouseEntity#securityGroups`

I think it should just be `@ManyToMany(mappedBy = SecurityGroupEntity_.COURTHOUSE_ENTITIES)`

Since the association is bi-directional, I think we are supposed to manage the other side of the relationship too, so that the entity manager knows about the opposite way back to security_group from courthouse?

So wondering if we might get unexpected results somewhere else, depending on how the courthouse service / management is used perhaps - something to look out for during testing

src/integrationTest/java/uk/gov/hmcts/darts/usermanagement/controller/PatchSecurityGroupIntTest.java
```diff
@@ -0,0 +1,346 @@
...
+            .andExpect(jsonPath("$.global_access").value(false))
+            .andExpect(jsonPath("$.security_role_id").isNumber())
+            .andExpect(jsonPath("$.courthouse_ids[0]").value(courthouseEntity1.getId()))
+            .andExpect(jsonPath("$.courthouse_ids[1]").value(courthouseEntity2.getId()))
+            .andExpect(jsonPath("$.user_ids").isEmpty())
+            .andReturn();
+    }
+
+    @Test
+    void patchSecurityGroupShouldSucceedWhenProvidedWithUserIds() throws Exception {
```
chrisbellingham-hmcts, line 146: When I try the PATCH containing a user_id when running locally the change doesn't appear to "stick". I.e I'd expect a record to be created in `security_group_user_account_ae`, but it doesn't appear to be, despite the ids appearing in the repsonse. Maybe I'm missing it, but can you check that scenario when running the full app locally against a db? 

Also would be worth checking the db state is as expected here in the integration test.

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/SecurityGroupServiceImpl.java
```diff
@@ -4,30 +4,44 @@
...
 
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Optional;
+import java.util.Set;
 
 import static uk.gov.hmcts.darts.usermanagement.exception.UserManagementError.SECURITY_GROUP_NOT_FOUND;
 
+
```
richardbull, line 35: Remove these 2 lines?

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/SecurityGroupServiceImpl.java
```diff
@@ -82,6 +96,87 @@ public List<SecurityGroupWithIdAndRole> getSecurityGroups(List<Integer> roleIds,
...
+    @Transactional
+    @Override
+    public SecurityGroupWithIdAndRoleAndUsers modifySecurityGroup(Integer securityGroupId, SecurityGroupPatch securityGroupPatch) {
+
+        Optional<SecurityGroupEntity> securityGroupEntityOptional = securityGroupRepository.findById(securityGroupId);
+
+        if (securityGroupEntityOptional.isPresent()) {
+            SecurityGroupEntity securityGroupEntity = securityGroupEntityOptional.get();
+            updateSecurityGroupEntity(securityGroupPatch, securityGroupEntity);
+            securityGroupRepository.saveAndFlush(securityGroupEntity);
```
richardbull, line 108: I think we should assign the variable here after the saveAndFlush operation?
i.e. `securityGroupEntity = `

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/SecurityGroupServiceImpl.java
```diff
@@ -82,6 +96,87 @@ public List<SecurityGroupWithIdAndRole> getSecurityGroups(List<Integer> roleIds,
...
+                }
+            }
+            securityGroupEntity.setUsers(userAccountEntities);
+        }
+    }
+
+    private void validate(SecurityGroupPatch securityGroupPatch, SecurityGroupEntity securityGroupEntity) {
+        Integer id = securityGroupEntity.getId();
+
+        securityGroupRepository.findByGroupNameIgnoreCaseAndIdNot(securityGroupPatch.getName(), id)
```
chrisbellingham-hmcts, line 170: Could you add a check to also disallow duplicate display_name please?

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/SecurityGroupServiceImpl.java
chrisbellingham-hmcts: Agree on non blank for name and displayName (although this should already be enforced at the openapi layer, but belt and braces?).

Description is optional, so blank is an allowable value as sending a blank string here this is the only way to "clear" an existing description.

src/main/resources/openapi/usermanagement.yaml
```diff
@@ -168,6 +168,36 @@ paths:
...
+        - SecurityGroup
+      operationId: modifySecurityGroup
+      parameters:
+        - $ref: '#/components/parameters/SecurityGroupId'
+      requestBody:
+        $ref: '#/components/requestBodies/SecurityGroupPatchRequest'
+      responses:
+        '200':
+          $ref: '#/components/responses/SecurityGroupPatchSuccess'
+        '403':
```
richardbull, line 183: I'd expect to see the other client errors listed here also - e.g. 404 you have `UserManagementError.SECURITY_GROUP_NOT_FOUND` and `UserManagementError.COURTHOUSE_NOT_FOUND` and
`UserManagementError.USER_NOT_FOUND`?

There is also 400, 401 errors?

Also 409? `DUPLICATE_SECURITY_GROUP_NAME_NOT_PERMITTED`

If you can check them please

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/SecurityGroupServiceImpl.java
chrisbellingham-hmcts: To that end @chrisdavecm , can you please relax the constraint for the `SecurityGroupDescription` schema type in `usermanagement.yaml` to have ` minLength: 0`, rather than `1`?

src/main/resources/openapi/usermanagement.yaml
```diff
@@ -197,6 +201,26 @@ paths:
...
                 status: 403
+        '404':
+          description: Not Found Error
+          content:
+            application/json+problem:
+              schema:
+                $ref: './problem.yaml'
+              example:
+                type: "USER_MANAGEMENT_104"
+                title: "The provided security group does not exist"
```
richardbull, line 212: Should the corresponding USER_MANAGEMENT_104 title here be `The provided courthouse does not exist`?

richardbull (approved):
Approved with minor comment on yaml - if you're happy to check / address?

src/functionalTest/java/uk/gov/hmcts/darts/usermanagement/UserManagementFunctionalTest.java
```diff
@@ -224,4 +228,94 @@ void shouldGetSecurityGroups() throws JsonProcessingException {
                                                                                               new TypeReference<List<SecurityGroupWithIdAndRole>>(){});
         assertFalse(securityGroupWithIdAndRoles.isEmpty());
     }
+
+    @Test
+    void shouldPatchSecurityGroups() throws JsonProcessingException {
```
richardbull, line 233: Wondering, should this test be put under the `SecurityGroupFunctionalTest` where you've been previously?
Maybe the test above can be delete just as it also appears in the other place?

src/functionalTest/java/uk/gov/hmcts/darts/usermanagement/UserManagementFunctionalTest.java
```diff
@@ -31,6 +34,13 @@ class UserManagementFunctionalTest extends FunctionalTest {
     private static final String EMAIL_ADDRESS = "Email-Address";
     private static final String COURTHOUSE_ID = "courthouse_id";
     private static final String ADMIN_USERS = "/admin/users";
+    static final ObjectMapper MAPPER = new ObjectMapper();
```
richardbull, line 37: We should probably consider adding the `ObjectMapperConfig` class to the `FunctionalTest` SpringBootTest to allow us to autowire the already configured ObjectMapper we have.

richardbull (approved):

PR #1250, opened 2024-03-25
postgres_app_v65_sec_v20_ret_v8

barry-jones50410:
app v65, sec v20, ret v8, & associated erd and drop schema scripts.

### Change description ###

```
[ ] Yes
[ ] No
```


src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -210,14 +210,35 @@
...
+--    add enhanced_auditing to audit table
+--v65 add table audit_heritage
 
 
 -- List of Table Aliases
 -- annotation                  ANN
 -- annotation_document         ADO
 -- audit                       AUD
 -- audit_activity              AUA
+-- audit_heritage              AUH
```
richardbull, line 241: Do we need to have a corresponding sequences for these 2 new tables: `auh_seq` and `ohr_seq`?
If so, then also drop statements in the other file

src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -1287,8 +1397,8 @@ CREATE TABLE transient_object_directory
 CREATE TABLE user_account
 (usr_id                      INTEGER                       NOT NULL
 ,dm_user_s_object_id         CHARACTER VARYING(16)
-,user_name                   CHARACTER VARYING             NOT NULL
-,user_full_name              CHARACTER VARYING             NOT NULL
+,user_name                   CHARACTER VARYING            
+,user_full_name              CHARACTER VARYING
```
richardbull, line 1401: Still finding it tricky to understand the identifying attributes on this table, if more columns are nullable.
Wondering what is important for the user state login information - are we still expecting user_email_address in modernised?

barry-jones50410:
Trusted 3rd Party

Audit_heritage is a store of legacy only audit records, so no inserts intended to arrive from modernised application , no sequence added

Object_hidden_reason is to support modernised application, so sequence has been added

The corresponding drop has been added to v23 of the drop schema

src/main/resources/db/reference/modernised_darts_schema.sql
barry-jones50410: I believe the intention here is to relax the not-null constraints to aid migration, but with the intention of tightening up again upon completion.
For example IMO user_name is the principal internal user_name and email is a natural key for user_account , yet both are nullable, to accommodate any intermediate unknowns.

src/main/resources/db/reference/modernised_darts_schema.sql
```diff
@@ -210,14 +210,35 @@
...
+--v63 add PK constraints to case_transcription_ae & hearing_transcription_ae
+--    add courthouse_code to daily_list
+--    amend transcription_commment.tra_id to be not null
+--v64 add retain_until_ts to annotation_document, case_document, media, transcription_document
+--    add last_modified_by/ts to annotation_document, case_document, transcription_document
+--    add is_deleted to court_case, media, transcription
+--    add update_retention to external_object_directory
+--    add is_retention_updated to court_case
+--    amend user_full_name and user_name on user_account to nullable
+--    amend checksum on annotation_document, case_document, transcription_document to nullable
```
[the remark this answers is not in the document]
barry-jones50410: given external_object_directory does not represent a single instance of any external object, having a single checksum here seems redundant.

src/main/resources/db/reference/modernised_darts_schema.sql
barry-jones50410: audit history is purely a copy of the legacy audit, hence no sequence to match auh_seq is envisaged.
ohr_seq has been added and the corresponding drop will be found in v23 of the drop script.

PR #1254, opened 2024-03-25
DMP-2470 Move data to unstructured

johnrace0:
https://tools.hmcts.net/jira/browse/DMP-2470

### Change description ###
Move data to Unstructured(ODS) from ARM

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/darts/audio/helper/UnstructuredDataHelper.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+        return returnVal;
+    }
+
+    private UUID saveToUnstructuredDataStore(ExternalObjectDirectoryEntity eodEntity, BinaryData binaryData) {
+        UUID uuid = null;
+        try {
+            Map<String, String> metadata = new HashMap<>();
+            metadata.put(MEDIA_REQUEST_ID, String.valueOf(eodEntity.getMedia().getId()));
+            BlobClient blobClient = dataManagementApi.saveBlobDataToContainer(binaryData, DatastoreContainerType.UNSTRUCTURED, metadata);
```
mestebanez, line 69: @chrisbellingham-hmcts I forget. Did you put a pattern in to avoid use of BinaryData when uploading. Should we try to avoid its use?

src/main/java/uk/gov/hmcts/darts/audio/helper/UnstructuredDataHelper.java
```diff
@@ -0,0 +1,146 @@
...
+            removeFromDatabase(eodEntityToDelete);
+        }
+        return returnVal;
+    }
+
+    private UUID saveToUnstructuredDataStore(ExternalObjectDirectoryEntity eodEntity, BinaryData binaryData) {
+        UUID uuid = null;
+        try {
+            Map<String, String> metadata = new HashMap<>();
+            metadata.put(MEDIA_REQUEST_ID, String.valueOf(eodEntity.getMedia().getId()));
```
richardbull, line 68: I'd expect the metadata here to be `media_id`?
We really don't want to be confusing it with `media_request_id`...

src/main/java/uk/gov/hmcts/darts/audio/helper/UnstructuredDataHelper.java
```diff
@@ -0,0 +1,146 @@
...
+            BlobClient blobClient = dataManagementApi.saveBlobDataToContainer(binaryData, DatastoreContainerType.UNSTRUCTURED, metadata);
+            uuid = UUID.fromString(blobClient.getBlobName());
+            log.debug(
+                "Completed upload to unstructured data store for mediaRequestId {}. Successfully uploaded with blobId: {}",
```
richardbull, line 72: should be `mediaId`

src/main/java/uk/gov/hmcts/darts/audio/helper/UnstructuredDataHelper.java
```diff
@@ -0,0 +1,146 @@
...
+                eodEntity.getMedia().getId(),
+                uuid
+            );
+        } catch (Exception e) {
+            log.error(
+                "Upload to unstructured data store failed for mediaRequestId {}.",
```
richardbull, line 78: should be `mediaId`, also add the `e` param to the end of the error log

src/main/java/uk/gov/hmcts/darts/audio/helper/UnstructuredDataHelper.java
```diff
@@ -0,0 +1,146 @@
...
+        if (annotationDocumentEntity != null) {
+            eodUnstructured.setAnnotationDocumentEntity(annotationDocumentEntity);
+        }
+        CaseDocumentEntity caseDocumentEntity = eod.getCaseDocument();
+        if (caseDocumentEntity != null) {
+            eodUnstructured.setCaseDocument(caseDocumentEntity);
+        }
+        eodUnstructured.setExternalLocation(uuid);
+        OffsetDateTime now = OffsetDateTime.now();
+        eodUnstructured.setCreatedDateTime(now);
```
richardbull, line 107: The ExternalObjectDirectoryEntity extends CreatedModifiedBaseEntity which has annotations on these 2 fields to stamp the time, so they are a bit redundant?

src/main/java/uk/gov/hmcts/darts/audio/helper/UnstructuredDataHelper.java
```diff
@@ -0,0 +1,146 @@
...
+        ExternalLocationTypeEntity unstructuredType = externalLocationTypeRepository.getReferenceById(UNSTRUCTURED.getId());
+        eodUnstructured.setStatus(storedStatus);
+        eodUnstructured.setExternalLocationType(unstructuredType);
+        eodUnstructured.setVerificationAttempts(1);
+        var systemUser = userAccountRepository.getReferenceById(SystemUsersEnum.DEFAULT.getId());
+        eodUnstructured.setCreatedBy(systemUser);
+        eodUnstructured.setLastModifiedBy(systemUser);
+        externalObjectDirectoryRepository.save(eodUnstructured);
+        log.debug(
+            "Created new record in EOD mediaRequestId {}. External location: {}",
```
richardbull, line 120: `mediaId`

src/main/java/uk/gov/hmcts/darts/audio/helper/UnstructuredDataHelper.java
```diff
@@ -0,0 +1,146 @@
...
+        } catch (Exception e) {
+            log.error(
+                "Upload to unstructured data store failed for mediaRequestId {}.",
+                eodEntity.getMedia().getId()
+            );
+        }
+        return uuid;
+    }
+
+    @Transactional
```
richardbull, line 85: I think these 2 private methods will need to be called from a public method with `@Transactional` to handle the composite save operations. The `@Transactional` can be removed from the private methods as they are redundant.

src/main/java/uk/gov/hmcts/darts/audio/helper/UnstructuredDataHelper.java
```diff
@@ -0,0 +1,146 @@
...
+        );
+    }
+
+    @Transactional
+    private void removeFromDatabase(ExternalObjectDirectoryEntity eodEntityToDelete) {
+        if (eodEntityToDelete != null) {
+            externalObjectDirectoryRepository.delete(eodEntityToDelete);
+            log.debug(
+                "Deleted old unstructured EOD mediaRequestId {}. External location: {}",
+                eodEntityToDelete.getMedia().getId(),
```
richardbull, line 132: Think you'll need to have another look at the logs and the params throughout.
Also what if the eod links to objects other than media, like transcription_document, then the current logging will NPE?

src/main/java/uk/gov/hmcts/darts/audio/helper/UnstructuredDataHelper.java
```diff
@@ -0,0 +1,146 @@
...
+    private final ExternalLocationTypeRepository externalLocationTypeRepository;
+    private final UserAccountRepository userAccountRepository;
+
+    private static final List<CompletableFuture> jobsList = new ArrayList<>();
+
+    public static List<CompletableFuture> getJobsList() {
+        return jobsList;
+    }
+
+    public boolean createUnstructured(ExternalObjectDirectoryEntity eodEntityToDelete, ExternalObjectDirectoryEntity eodEntity, BinaryData binaryData) {
```
chrisbellingham-hmcts, line 52: Can this method be renamed to better describe what it's doing?

src/main/java/uk/gov/hmcts/darts/audio/helper/UnstructuredDataHelper.java
chrisbellingham-hmcts: Yep, the downside of the approach here is the underlying `public BlobClient saveBlobData(String containerName, BinaryData binaryData, Map<String, String> metadata)` method in `DataManagementApiImpl` will read all data into memory.

Ideally we want to use the `public UUID saveBlobData(String containerName, InputStream inputStream)` method in `DataManagementApiImpl`, which expects an inputstream.

src/main/java/uk/gov/hmcts/darts/common/datamanagement/api/DataManagementFacadeImpl.java
```diff
@@ -155,6 +167,22 @@ private DownloadResponseMetaData getDataFromStorage(List<ExternalObjectDirectory
...
+    ) throws IOException {
+        if (downloadResponseMetaData.getInputStream() != null) {
+            BinaryData binaryData = BinaryData.fromStream(downloadResponseMetaData.getInputStream());
+            CompletableFuture<Void> createUnstructuredJob = CompletableFuture.runAsync(() -> {
+                unstructuredDataHelper.createUnstructured(
+                    eodEntityToDelete,
+                    downloadResponseMetaData.getEodEntity(),
+                    binaryData);
+            });
+            UnstructuredDataHelper.addToJobsList(createUnstructuredJob);
```
chrisbellingham-hmcts, line 182: Calling a static method here is a little contrary to the general IoC approach we're taking elsewhere. Can we instead autowire the `unstructuredDataHelper` component, and call it's `addToJobsList()` method in a non-static way?

src/main/java/uk/gov/hmcts/darts/audio/helper/UnstructuredDataHelper.java
```diff
@@ -0,0 +1,146 @@
...
+
+    public static void addToJobsList(CompletableFuture<Void> saveToUnstructuredFuture) {
+        jobsList.add(saveToUnstructuredFuture);
+    }
+
+    public static void waitForAllJobsToFinish() {
+        jobsList.forEach(CompletableFuture::join);
+        jobsList.clear();
+    }
+}
```
chrisbellingham-hmcts, line 146: Is it necessary for these to be static?

src/main/java/uk/gov/hmcts/darts/audio/helper/UnstructuredDataHelper.java
johnrace0: removed

src/main/java/uk/gov/hmcts/darts/audio/helper/UnstructuredDataHelper.java
johnrace0: @chrisbellingham-hmcts / @mestebanez - having a bit of fun with this one :-/
the job thats doing the blob saving is running in CompletableFuture. unfortunately when I switch to use the input stream, the temp input stream has been deleted before the async job gets run.
can't see any way round this apart from switching back to using BinaryData - open to suggestions :-D

src/main/java/uk/gov/hmcts/darts/common/datamanagement/api/DataManagementFacadeImpl.java
johnrace0: removed statics

src/main/java/uk/gov/hmcts/darts/audio/helper/UnstructuredDataHelper.java
johnrace0: removed statics

src/main/java/uk/gov/hmcts/darts/audio/helper/UnstructuredDataHelper.java
chrisbellingham-hmcts: Maybe we put in a pin in this one, and look to address this via an improvement ticket. Practically, memory efficiency is less critical in ATS MODE, as we'll have a single pod dedicated to servicing the task. That being said I think it still needs solved to allow us to cater for particularly large files.

@hemantasharma1129, what's your view on this one, shall this be merged as-is or do we need this to suport large files (files whose size exceeds the memory allocation of a pod) immediately?

hemantasharma1129: I agree with @chrisbellingham-hmcts . We are suppose to use streams only as loading a file in memory is something we are no longer doing.

johnrace0: sorted - as per discussion

PR #1260, opened 2024-03-26
Dmp 2244 Unstructured To Arm Batch Processor

mario-paniccia:
https://tools.hmcts.net/jira/browse/DMP-2244

Note:
I've tried to implement the new batch processor `UnstructuredToArmBatchProcessorImpl` to be as similar as possible in coding style and naming conventions to `UnstructuredToArmProcessorImpl` and to reuse as much code as possible from `UnstructuredToArmProcessorImpl`


src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmBatchProcessorImpl.java
```diff
@@ -0,0 +1,316 @@
...
+        if (armClient.equalsIgnoreCase("darts")) {
+            return EodHelper.unstructuredLocation();
+        } else if (armClient.equalsIgnoreCase("dets")) {
+            return EodHelper.detsLocation();
+        } else {
+            throw new DartsException(String.format("Invalid arm client '%s'", armClient));
+        }
+    }
+
+    private List<ExternalObjectDirectoryEntity> getArmExternalObjectDirectoryEntities(int batchSize) {
```
karen-hedges, line 148: So this method will prioritise new EOD records that have not been attempted to be pushed to ARM over those that have already been attempted?

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmBatchProcessorImpl.java
```diff
@@ -0,0 +1,316 @@
...
+
+        log.info("Finished running ARM Batch Push processing at: {}", OffsetDateTime.now());
+    }
+
+    private ExternalLocationTypeEntity getEodSourceLocation() {
+        var armClient = armDataManagementConfiguration.getArmClient();
+        if (armClient.equalsIgnoreCase("darts")) {
+            return EodHelper.unstructuredLocation();
+        } else if (armClient.equalsIgnoreCase("dets")) {
+            return EodHelper.detsLocation();
```
karen-hedges, line 142: This job is to move data from unstructured to ARM, so if the client is darts or dets they will both move to ARM? Also is there a possibility of a conflict between the 2?

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmBatchProcessorImpl.java
mario-paniccia: this is a valid point. But the job eventually will push them all anyway, right. If we believe it's important prioritising both types of eods in each run, I will make the change. Might not be an easy one though

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmBatchProcessorImpl.java
mario-paniccia: correct. The ticket requirements state that based on the value of the armClient config, the job needs to get the records either from unstructured or from dets. And push them to arm.
With regards to conflicts, I'll let @hemantasharma1129 reply. But I don't think

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmBatchProcessorImpl.java
karen-hedges: I think you misunderstood what I mean't. If you swap them around so you get batch from the existing records first and then get the remainder of the new records if there is a remainder. @hemantasharma1129 what are your thoughts on this?

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmBatchProcessorImpl.java
```diff
@@ -0,0 +1,316 @@
...
+     * Contains info related to the processing of a batch item.
+     */
+    @Data
+    static class BatchItem {
+
+        private ExternalObjectDirectoryEntity sourceEod;
+        private ExternalObjectDirectoryEntity armEod;
+        private String previousManifestFile;
+        private ObjectRecordStatusEntity previousStatus;
+        private Boolean rawFilePushSuccessful;
```
karen-hedges, line 280: Should this have a default value of false or be a primitive boolean?

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmBatchProcessorImpl.java
```diff
@@ -0,0 +1,316 @@
...
+            this.previousManifestFile = armEod.getManifestFile();
+            this.previousStatus = armEod.getStatus();
+        }
+
+        public void undoManifestFileChange() {
+            this.armEod.setManifestFile(this.previousManifestFile);
+        }
+
+        public boolean isRawFilePushNotNeededOrSuccessfulWhenNeeded() {
+            return rawFilePushSuccessful == null || rawFilePushSuccessful;
```
karen-hedges, line 294: Predominately people now use nonNull instead for null checks but for this Boolean you could use Boolean.TRUE.equals(rawFilePushSuccessful)

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmBatchProcessorImpl.java
mario-paniccia: That was on purpose. I needed a way to distinguish when the push is not needed so not attempted. And when instead it was attempted and its outcome. Rather than using an Optional<Boolean> I check when that field it's `null` to know that the push was not needed

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmBatchProcessorImpl.java
mario-paniccia: that won't work in my case, because when `rawFilePushSuccessful` I need to get a `true`. Because I need to distinguish when the raw push is not attempted because not needed

src/main/resources/application.yaml
```diff
@@ -199,6 +199,10 @@ darts:
 
       response-cleanup-buffer-days: 0
 
+      batch-size: ${ARM_BATCH_SIZE:10}
```
hemantasharma1129, line 202: Dont think we need env variables. We can hardcode the properties here.

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmBatchProcessorImpl.java
mario-paniccia: ok I see, you're suggesting to just swap them. Ok sure happy to do that. Let's hear from Hemanta first

src/main/resources/application.yaml
mario-paniccia: I've left the env var as agreed in a separate conversation. It has been agreed that we'll have a follow up ticket to complete the setup of the env vars everywhere else is needed, since this was out of scope for this ticket

hemantasharma1129 (approved):

PR #1262, opened 2024-03-26
DMP-2322 change RetrieveCoreObjectService

JasonPittockCGI:
https://tools.hmcts.net/jira/browse/DMP-2322

DMP-2322 change RetrieveCoreObjectService to include createdBy and LastModifiedBy

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/common/service/impl/RetrieveCoreObjectServiceImpl.java
```diff
@@ -69,17 +73,22 @@ public HearingEntity retrieveOrCreateHearing(String courthouseName, String court
...
         HearingEntity hearing = new HearingEntity();
+
         hearing.setCourtCase(courtCase);
         hearing.setCourtroom(courtroom);
         hearing.setHearingDate(hearingDate);
         hearing.setNew(true);
         hearing.setHearingIsActual(false);
         hearing.setCreatedBy(userAccount);
         hearing.setLastModifiedBy(userAccount);
+        hearing.setCreatedDateTime(OffsetDateTime.now());
```
chrisbellingham-hmcts, line 87: The disadvantage of calling the static `OffsetDateTime.now()` method is that we have no way to mock the behaviour of this, which makes it hard to test as this value will be different every time the test runs.

So, if we do need to set the timestamp manually via the setter, it's better to use `currentTimeHelper.currentOffsetDateTime()` to obtain the timestamp, such that `currentTimeHelper` can be mocked in tests with a fixed value.

This comment applies to all uses of `OffsetDateTime.now()` in this PR.

src/main/java/uk/gov/hmcts/darts/common/service/impl/RetrieveCoreObjectServiceImpl.java
```diff
@@ -43,6 +45,7 @@ public class RetrieveCoreObjectServiceImpl implements RetrieveCoreObjectService
     private final DefendantRepository defendantRepository;
     private final ProsecutorRepository prosecutorRepository;
     private final AuthorisationApi authorisationApi;
+    private final CurrentTimeHelper currentTimeHelper;
```
chrisbellingham-hmcts, line 48: This is unused (but I think it should be used, see other comments on use of `CurrentDateTime.now()`

src/main/java/uk/gov/hmcts/darts/common/service/impl/RetrieveCoreObjectServiceImpl.java
```diff
@@ -69,17 +73,22 @@ public HearingEntity retrieveOrCreateHearing(String courthouseName, String court
...
+
         hearing.setCourtCase(courtCase);
         hearing.setCourtroom(courtroom);
         hearing.setHearingDate(hearingDate);
         hearing.setNew(true);
         hearing.setHearingIsActual(false);
         hearing.setCreatedBy(userAccount);
         hearing.setLastModifiedBy(userAccount);
+        hearing.setCreatedDateTime(OffsetDateTime.now());
+        hearing.setLastModifiedDateTime(OffsetDateTime.now());
```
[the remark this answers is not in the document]
JasonPittockCGI: Thanks Chris. If we can drop them, that would be cleaner. Looks like they were set unnecessarily in the code before I centralised it in RetrieveCoreObjectService. I'll address and push again.

chrisbellingham-hmcts (approved):

PR #1268, opened 2024-04-01
DMP-2582 - increase advanced search performance

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-2582

- results limited to 501
- allowed the requestor to send in the courthouseId or courtroomId (in case we want to enhance the screen to allow it to send these in directly, due to a dropdown)
- i have created a request complexity calculator that takes a few things into account to decide whether enough criteria has been given, e.g. a judge name of 2 characters or less is not considered complex enough, and so would need to be combined with other items to be considered, please see code here https://github.com/hmcts/darts-api/pull/1268/files#diff-196111a004439aa49c440cc62372f141475407729c7a0abbc62a77d17a451e2aR33
- i have also changed the query to get the courthouseId or courtroomId up front, to get around the issue with the postgres optimiser not working correctly.
- i've removed the needless join onto the userAccount table.

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/repository/CourthouseRepository.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/repository/CourthouseRepository.java
@@ -28,6 +28,14 @@ public interface CourthouseRepository extends JpaRepository<CourthouseEntity, In
         """)
     List<CourthouseEntity> findAuthorisedCourthousesForEmailAddressOrGuid(String emailAddress, Set<Integer> roleIds, String guid);
 
+    @Query("""
+        SELECT DISTINCT ch.id
+        FROM CourthouseEntity ch
+        WHERE upper(ch.courthouseName) like upper(%:name%)
+        or upper(ch.displayName) like upper(%:name%)
+        """)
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/hmcts/darts/common/repository/CourthouseRepository.java
```diff
@@ -28,6 +28,14 @@ AND securityRole.id IN (:roleIds)
         """)
     List<CourthouseEntity> findAuthorisedCourthousesForEmailAddressOrGuid(String emailAddress, Set<Integer> roleIds, String guid);
 
+    @Query("""
+        SELECT DISTINCT ch.id
```
davet1985, line 32: Do we really need DISTINCT here?

davet1985 (approved):

src/main/java/uk/gov/hmcts/darts/common/repository/CourthouseRepository.java
cmagowan-hmcts: removed

PR #1278, opened 2024-04-03
DMP-2532: Implement POST /admin/retention-policy-types

chrisbellingham-hmcts:
## [DMP-2532](https://tools.hmcts.net/jira/browse/DMP-2532)

Implement new endpoint `POST /admin/retention-policy-types`.

This change also includes refactoring to:
- move all functionality related to managing RetentionPolicyTypes into a dedicated service class
- leverage mapstruct for entity/model mapping
- tidy up openapi schema

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/retention/exception/RetentionApiError.java
```diff
@@ -47,6 +47,34 @@ public enum RetentionApiError implements DartsApiError {
         RetentionErrorCode.RETENTION_POLICY_TYPE_ID_NOT_FOUND.getValue(),
         HttpStatus.NOT_FOUND,
         RetentionTitleErrors.RETENTION_POLICY_TYPE_ID_NOT_FOUND.toString()
+    ), NON_UNIQUE_POLICY_NAME(
```
johnrace0, line 50: general comment - we have a lot of this horrible looking boilerplate code - just wondered (not in scope for this ticket) if we could/should do this a better way?

src/test/java/uk/gov/hmcts/darts/retention/validation/PolicyDurationValidatorTest.java
```diff
@@ -0,0 +1,34 @@
...
+        assertDoesNotThrow(() -> policyDurationValidator.validate(durationString));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"00Y00M00D", "0Y0M0D"})
+    void validateShouldThrowExceptionForInvalidValues(String durationString) {
+        var exception = assertThrows(DartsApiException.class, () -> policyDurationValidator.validate(durationString));
+        assertEquals("Duration too short", exception.getMessage());
+    }
+
```
johnrace0, line 33: is it worth having a test for completely dodgy input?

chrisbellingham-hmcts: This is actually the second layer of `duration` validation - the first layer is via openapi which requires the `duration` value to match regex `^\d{1,2}Y\d{1,2}M\d{1,2}D$`, which should exclude dodgy input.

`PolicyDurationValidator` therefore assumes input is per that format, so the test cases are limited to reflect that.

However I have added a comment in the JavaDoc for `PolicyDurationValidator` which states:
```
...
It is assumed the format of the input string matches this pattern through prior validation checks (e.g. at openapi layer).
...
```

src/main/java/uk/gov/hmcts/darts/retention/service/RetentionPolicyService.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.hmcts.darts.retention.service;
+
+import uk.gov.hmcts.darts.retentions.model.AdminPostRetentionRequest;
+import uk.gov.hmcts.darts.retentions.model.RetentionPolicy;
+
+import java.util.List;
+
+public interface RetentionPolicyService {
+
+    List<RetentionPolicy> getRetentionPolicyTypes();
```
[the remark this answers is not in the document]
chrisbellingham-hmcts: Good spot, have renamed `RetentionPolicy` to `RetentionPolicyType` in the openapi spec, and have blown that change through to this class. I have also updated the service name from `RetentionPolicyService` to `RetentionPolicyTypeService` for consistency.

src/main/java/uk/gov/hmcts/darts/retention/exception/RetentionApiError.java
chrisbellingham-hmcts: Yeah it's tricky. This is actually an improvement on how it used to be - we used to hardcode the error codes and the titles directly into the enum here. We now define that info in the openapi (with the aim of self-documenting the error details), and use this enum to tie everything together.

Definitely open to suggestions on how to make this better though, if you have any ideas!

johnrace0 (approved):

PR #1282, opened 2024-04-03
DMP-2249 Setting created/lastModified by columns on daily list

davet1985:

https://tools.hmcts.net/jira/browse/DMP-2249

### Change description ###

- covering various scenarios when creating/updating a daily list
- removing a duplicate parse/save of the daily list JSON

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListServiceImpl.java
@@ -1,6 +1,5 @@
 package uk.gov.hmcts.darts.dailylist.service.impl;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import jakarta.transaction.Transactional;
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListServiceImpl.java
```diff
@@ -59,13 +60,16 @@ public PostDailyListResponse saveDailyListToDatabase(DailyListPostRequest postRe
             //update the record
             savedDailyListEntity = existingRecordOpt.get();
             dailyListMapper.updateDailyListEntity(postRequest, listingCourthouse, savedDailyListEntity);
+            savedDailyListEntity.setLastModifiedBy(userIdentity.getUserAccount());
```
cmagowan-hmcts, line 63: every time we call `userIdentity.getUserAccount()` there are all sorts of lookups on the database, so if we need to call it, probably best to just call it once and keep the result.

davet1985: fixed

cmagowan-hmcts (approved):

PR #1291, opened 2024-04-05
DMP-2717 bug: timestamp on hearing and courtcases for daily list

JasonPittockCGI:
https://tools.hmcts.net/jira/browse/DMP-2717

DMP-2717 bug: timestamp on hearing and courtcases for daily list

```
[ ] Yes
[ x] No
```


davet1985 (commented):
Is it easy to add a simple test for this? It would help to prevent someone from removing these in the future.

JasonPittockCGI:
> Is it easy to add a simple test for this? It would help to prevent someone from removing these in the future.

Will have a look.

davet1985 (approved):

PR #1319, opened 2024-04-11
DMP-2410 Fetching Courtroom details via Hearing table

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-2410

### Change description ###

Added checking to get via hearing if it exists

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/arm/mapper/impl/MediaArchiveRecordMapperImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/arm/mapper/impl/MediaArchiveRecordMapperImpl.java
@@ -257,7 +257,9 @@ public class MediaArchiveRecordMapperImpl implements MediaArchiveRecordMapper {
 
     private static String getCourtroom(MediaEntity media) {
         String courtroom = null;
-        if (nonNull(media.getCourtroom())) {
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/darts/arm/mapper/impl/MediaArchiveRecordMapperImpl.java
```diff
@@ -257,7 +257,9 @@ private static String getCourthouse(MediaEntity media) {
 
     private static String getCourtroom(MediaEntity media) {
         String courtroom = null;
-        if (nonNull(media.getCourtroom())) {
+        if (CollectionUtils.isNotEmpty(media.getHearingList()) && nonNull(media.getHearingList().get(0).getCourtroom())) {
+            courtroom = media.getHearingList().get(0).getCourtroom().getName();
+        } else if (nonNull(media.getCourtroom())) {
```
johnrace0, line 262: Q - ticket says that media courtroom is nullable so should this drop in to here?

karen-hedges: What it should do it try the hearing first as it should really always have a hearing but in case there isn't it does the null check for courtroom before attempting to get the courtroom name from the courtroom

johnrace0 (approved):

PR #1325, opened 2024-04-12
Dmp 2772 - manage duplicate audio upload

mestebanez:

https://tools.hmcts.net/jira/browse/DMP-2772

### Change description ###

Add ability to version files when uploading case files

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioServiceImpl.java
```diff
@@ -123,46 +129,157 @@ public BinaryData encode(Integer mediaId) {
...
+            MediaEntity saveEntity = entity;
+
+            // if the media already exists in the database then create a new media file
+            if (entity.getId() != null) {
+                saveEntity = mapper.mapToMedia(addAudioMetadataRequest);
+
+                saveEntity.setChronicleId(entity.getChronicleId());
+                saveEntity.setAntecedentId(entity.getId().toString());
+
+                log.info("Uploading version of duplicate filename {} under version number {}", entity.getMediaFile(), entity.getLegacyVersionLabel());
```
jackmaloney, line 202: I thought this LegacyVersionLabel was only used for migrated cases?

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioServiceImpl.java
```diff
@@ -123,46 +129,157 @@ public BinaryData encode(Integer mediaId) {
...
+            } else {
+                log.info("New file uploaded {} with version number {}", entity.getMediaFile(), entity.getLegacyVersionLabel());
+
+                saveEntity = mediaRepository.save(saveEntity);
```
jackmaloney, line 206: Is this save needed if it saves again on line 213?

mestebanez: for new media file uploads we set the chronicle id to the same id of the new media record. We first save to get the id and then update the chronicle id with that same id. This only happens on a new media record

jackmaloney (approved):

PR #1332, opened 2024-04-15
DMP-2750 Fixing swagger schema for /userstate endpoint

davet1985:

https://tools.hmcts.net/jira/browse/DMP-2750

### Change description ###

- renaming the `Role` class to `UserStateRole`
- this was incorrectly using another model class named Role which is created by OpenAPI

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/authorisation/model/UserStateRole.java
+++ b/src/main/java/uk/gov/hmcts/darts/authorisation/model/UserStateRole.java
@@ -10,8 +10,7 @@ import java.util.Set;
 
 @Builder
 @Value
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/hmcts/darts/authorisation/model/UserStateRole.java
```diff
@@ -11,7 +11,7 @@
 @Builder
 @Value
 @SuppressWarnings({"PMD.ShortClassName"})
```
chrisbellingham-hmcts, line 13: Maybe this suppression can be removed now?

chrisbellingham-hmcts (approved):

PR #1346, opened 2024-04-17
Strip spaces in time marking note

jackmaloney:


src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListUpdater.java
```diff
@@ -162,7 +162,7 @@ private LocalTime getTimeFromTimeMarkingNote(final String timeMarkingNote) throw
                 rawTime = timeMarkingNote;
             }
 
-            return LocalTime.parse(rawTime, new DateTimeFormatterBuilder()
+            return LocalTime.parse(rawTime.strip(), new DateTimeFormatterBuilder()
```
cmagowan-hmcts, line 165: it might be better to put the change slightly higher in the code, around line 121 maybe?
image

PR #1361, opened 2024-04-22
Dmp 2824

mario-paniccia:


src/integrationTest/java/uk/gov/hmcts/darts/testutils/data/MediaTestData.java
```diff
@@ -71,7 +66,7 @@ public static MediaEntity createMediaWith(CourtroomEntity courtroomEntity, Offse
     }
 
     private String getChecksum() {
-        return new String(encodeBase64(md5(getBinaryData().toBytes())));
+        return TestUtils.encodeToString(md5(MEDIA_TEST_DATA_BINARY_DATA));
```
karen-hedges, line 69: You can call the FileContentChecksum instead so there aren't multiple versions of doing the same thing

src/main/java/uk/gov/hmcts/darts/common/util/FileContentChecksum.java
```diff
@@ -29,10 +34,20 @@ public String calculate(byte[] bytes) {
...
     public String calculate(DigestInputStream digestInputStream) throws IOException {
         return encodeToString(digestInputStream.getMessageDigest().digest());
     }
 
+    @SneakyThrows
+    public String calculateFromFile(Path filePath) {
+        ByteSource byteSource = com.google.common.io.Files.asByteSource(filePath.toFile());
+        HashCode hc = byteSource.hash(Hashing.md5());
+        return encodeToString(hc.asBytes());
+    }
```
karen-hedges, line 49: You could follow the same code that generates the hashcode as is in the method FileContentChecksumTest.calculateFromInputStreamUsingAudioFileFromAzure for the file without the need to use another library

src/test/java/uk/gov/hmcts/darts/datamanagement/service/InboundToUnstructuredProcessorSingleElementImplTest.java
```diff
@@ -117,21 +119,19 @@ void setUp() {
...
 
-        BinaryData binaryData = BinaryData.fromString(TEST_BINARY_DATA);
-        String calculatedChecksum = new String(encodeBase64(md5(binaryData.toBytes())));
-
         when(externalLocationTypeRepository.getReferenceById(2)).thenReturn(externalLocationTypeUnstructured);
         when(externalObjectDirectoryEntityInbound.getMedia()).thenReturn(mediaEntity);
 
         when(mediaEntity.getMediaFormat()).thenReturn(MP2);
-        when(mediaEntity.getChecksum()).thenReturn(calculatedChecksum);
+        when(mediaEntity.getChecksum()).thenReturn("checksum");
```
karen-hedges, line 134: Checksum values should be validated with actual values not test data

src/main/java/uk/gov/hmcts/darts/common/util/FileContentChecksum.java
hemantasharma1129: I agree. We dont have to add another method to calculate the MD5 checksum unless its not possible to do with existing method.

src/main/java/uk/gov/hmcts/darts/datamanagement/service/impl/DataManagementServiceImpl.java
```diff
@@ -63,6 +66,26 @@ public BinaryData getBlobData(String containerName, UUID blobId) {
...
+    public Path downloadBlobToFile(String containerName, UUID blobId, String workspace) {
+        BlobServiceClient serviceClient = blobServiceFactory.getBlobServiceClient(dataManagementConfiguration.getBlobStorageAccountConnectionString());
+        BlobContainerClient containerClient = blobServiceFactory.getBlobContainerClient(containerName, serviceClient);
+        BlobClient blobClient = blobServiceFactory.getBlobClient(containerClient, blobId);
+        if (!blobClient.exists()) {
+            log.error("Blob {} does not exist in {} container", blobId, containerName);
+        }
+
+        Path workspacePath = Path.of(workspace);
+        Path targetFile = workspacePath.resolve(UUID.randomUUID() + ".tmp");
```
hemantasharma1129, line 80: Can this change the current checksum of the file? If so, this will cause issues in further validation.

src/integrationTest/java/uk/gov/hmcts/darts/testutils/data/MediaTestData.java
mario-paniccia: that method is `private`, I would have to make it `public` (not even `protected` will work) just for purpose of testing. Is that an acceptable tradeoff?

src/main/java/uk/gov/hmcts/darts/datamanagement/service/impl/DataManagementServiceImpl.java
mario-paniccia: the checksum does not depend on the file name, only on the content so I don't think. Verified locally with Mac Os `md5` command line tool

src/main/java/uk/gov/hmcts/darts/common/util/FileContentChecksum.java
mario-paniccia: My initial version did that, but then I disliked that solution because using a `DigestInputStream` adds so much boilerplate. Furthermore, with the existing solution does not follow proper encapsulation best practice. It's the client that needs to create the `DigestInputStream` rather than `FileContentChecksum`

PR #1363, opened 2024-04-22
DMP-1868 - Add the request and file upload thresholds from the vault

mestebanez:

https://tools.hmcts.net/jira/browse/DMP-1868

### Change description ###

Added support for add audio request and file size thresholds to be derived from the vault

```
[ ] Yes
[x] No
```


jackmaloney (approved):

mestebanez:
> Quick question on this one - how come we're using the vault for these values and not Flux?

No idea. This is consistent with the way we were asked to do it for gateway.

PR #1372, opened 2024-04-24
DMP-2944 ARM - Single Mode - Replicate changes made to batch mode

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-2944

### Change description ###

Added logging of ARM response files CR, UF and IL contents. 
Changed checksum verification to be case insensitive. 
Changed ARM push and pull to single modeinstead of batch mode.

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
```diff
@@ -222,6 +229,71 @@ private void processUploadFile(String armInputUploadFilename, List<String> respo
         }
     }
 
+    private void getAndLogCreateRecordFile(ExternalObjectDirectoryEntity externalObjectDirectory, String createRecordFilename) {
+        if (nonNull(createRecordFilename)) {
+            Path jsonPath = null;
+            try {
+                CreateRecordFilenameProcessor createRecordFilenameProcessor = new CreateRecordFilenameProcessor(createRecordFilename);
+                BinaryData createRecordFileBinary = armDataManagementApi.getBlobData(createRecordFilenameProcessor.getCreateRecordFilenameAndPath());
```
chrisbellingham-hmcts, line 237: Is this approach safe for large files (if large files are possible/relevant to this use case?)

src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
```diff
@@ -222,6 +229,71 @@ private void processUploadFile(String armInputUploadFilename, List<String> respo
...
+                        createRecordFilenameProcessor.getCreateRecordFilename(),
+                        armDataManagementConfiguration.getTempBlobWorkspace(),
+                        appendUuidToWorkspace
+                    );
+
+                    if (nonNull(jsonPath) && jsonPath.toFile().exists()) {
+                        logResponseFileContents(jsonPath);
+                    }
+                }
+            } catch (Exception e) {
```
chrisbellingham-hmcts, line 251: It looks like only `IOException` is thrown in the try, so we can we catch that rather than catching the top-level `Exception` type?

src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
```diff
@@ -222,6 +229,71 @@ private void processUploadFile(String armInputUploadFilename, List<String> respo
...
+                        uploadFileFilenameProcessor.getUploadFileFilename(),
+                        armDataManagementConfiguration.getTempBlobWorkspace(),
+                        appendUuidToWorkspace
+                    );
+
+                    if (nonNull(jsonPath) && jsonPath.toFile().exists()) {
+                        logResponseFileContents(jsonPath);
+                    }
+                }
+            } catch (Exception e) {
```
chrisbellingham-hmcts, line 278: Can we catch a more specific exception?

src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
```diff
@@ -222,6 +229,71 @@ private void processUploadFile(String armInputUploadFilename, List<String> respo
...
+        }
+    }
+
+    private void logResponseFileContents(Path jsonPath) {
+        try {
+            String contents = FileUtils.readFileToString(jsonPath.toFile(), UTF_8);
+            log.info("Contents of ARM response file {} - \n{}",
+                     jsonPath.toFile().getAbsoluteFile(),
+                     contents);
+        } catch (Exception e) {
```
chrisbellingham-hmcts, line 292: Can we catch a more specific exception?

src/main/resources/application.yaml
```diff
@@ -205,7 +205,7 @@ darts:
       response-cleanup-buffer-days: 0
       event-date-adjustment-years: 100
 
-      batch-size: ${ARM_BATCH_SIZE:10}
+      batch-size: ${ARM_BATCH_SIZE:0}
```
chrisbellingham-hmcts, line 208: Intentional?

src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
```diff
@@ -99,8 +102,7 @@ private void initialisePreloadedObjects() {
     }
 
     private void processInputUploadFile(ExternalObjectDirectoryEntity externalObjectDirectory) {
-        // IU - Input Upload - This is the manifest file which gets renamed by ARM.
-        // EODID_MEDID_ATTEMPTS_6a374f19a9ce7dc9cc480ea8d4eca0fb_1_iu.rsp
+        // IU Input Upload File - EODID_MEDID_ATTEMPTS_6a374f19a9ce7dc9cc480ea8d4eca0fb_1_iu.rsp
```
chrisbellingham-hmcts, line 105: Can any context be added to this comment, as it looks a bit cryptic to the uninformed reader. What is the signifance of the `EODID_MEDID_ATTEMPTS_6a374f19a9ce7dc9cc480ea8d4eca0fb_1_iu.rsp` part?

src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
karen-hedges: these are response files which are small json files so size wont be an issue. There is going to be a ticket to change all the ARM stuff from BinaryData but its not been played yet.

src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
karen-hedges: The filename starts with external object directory id  then the object id such as media id then the transfer attempts then some hash code, then a status then the file type so in this case its IU

karen-hedges: Its not the manifest file as was originally thought

src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
karen-hedges: We don't really care what exceptions are thrown as it stops the whole process. All we are doing is logging the contents of the response file so that if the file gets deleted we have some traceability of what was in it. It helps with identifying any issues more than anything else.

src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
karen-hedges: Again as mentioned above not really relevant as all we want to do is log the contents so specific exceptions are necessary.

src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
karen-hedges: As above

src/main/resources/application.yaml
karen-hedges: yes thats to switch from batch mode to single mode. There is a ticket in the next sprint to remove this into the automated tasks table.

src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
chrisbellingham-hmcts: Ok, can a description to that effect be added to the actual code comment? I think that would make the intent of the comment a bit clearer.

chrisbellingham-hmcts (approved):

PR #1388, opened 2024-04-26
DMP-2735: Transcription search

tonymort:
https://tools.hmcts.net/jira/browse/DMP-2735

Admin search transcriptions.  

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/repository/TranscriptionRepository.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/repository/TranscriptionRepository.java
@@ -6,7 +6,9 @@ import org.springframework.stereotype.Repository;
 import uk.gov.hmcts.darts.common.entity.TranscriptionEntity;
 import uk.gov.hmcts.darts.common.entity.TranscriptionStatusEntity;
 import uk.gov.hmcts.darts.common.entity.TranscriptionTypeEntity;
+import uk.gov.hmcts.darts.transcriptions.model.TranscriptionSearchResult;
 
+import java.time.LocalDate;
 import java.time.OffsetDateTime;
 import java.util.List;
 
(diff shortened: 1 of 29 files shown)


src/main/java/uk/gov/hmcts/darts/common/repository/TranscriptionRepository.java
```diff
@@ -75,4 +77,74 @@ List<TranscriptionEntity> findByHearingIdTypeStartAndEndAndIsManualAndNotStatus(
...
+                t.createdDateTime,
+                ts.id,
+                t.isManualTranscription)
+            FROM TranscriptionEntity t
+            JOIN t.hearings h
+            JOIN h.courtCase cc
+            JOIN h.courtroom cr
+            JOIN cr.courthouse cth
+            JOIN t.transcriptionStatus ts
+            JOIN t.createdBy ua
```
cmagowan-hmcts, line 96: how worried are we regarding performance here?

e.g. if the requestedBy is not provided in the search, then there would be no point in joining the UserAccount table in the query. Perhaps the performance hit is negligible?

tonymort: The performance seems quite good against 5000 records in my local postgres  about 1 second.  So in the absence any performance requirements I think that seems ok.

src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/AdminTranscriptionSearchServiceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+        transcriptionSearchResponse.setCaseNumber(transcriptionSearchResult.caseNumber());
+        transcriptionSearchResponse.setCourthouseId(transcriptionSearchResult.courthouseId());
+        transcriptionSearchResponse.setHearingDate(transcriptionSearchResult.hearingDate());
+        transcriptionSearchResponse.setRequestedAt(transcriptionSearchResult.requestedAt());
+        transcriptionSearchResponse.setIsManualTranscription(transcriptionSearchResult.isManualTranscription());
+        transcriptionSearchResponse.setTranscriptionStatusId(transcriptionSearchResult.transcriptionStatusId());
+        return transcriptionSearchResponse;
+    }
+
+    private static boolean transcriptionIdIsNotOwnedBy(Integer transcriptionId, List<Integer> transcriptionsForOwner) {
```
cmagowan-hmcts, line 62: This method looks like it might create confusing results.
It might be better to flip the logic?
i.e.
```
private static boolean transcriptionIdIsOwnedBy(Integer transcriptionId, List<Integer> transcriptionsForOwner) {
        return transcriptionId != null
            && !isEmpty(transcriptionsForOwner)
            && transcriptionsForOwner.contains(transcriptionId);
    }
```

Then when you call it use !transcriptionIdIsOwnedBy.

you would therefore no longer need the separate check on line 28 `if (transcriptionsForOwner.isEmpty()) {`

tonymort: Yeah I'll go with that

davet1985 (approved):

PR #1397, opened 2024-04-30
DMP-2812: Returning automated task by id

tonymort:
https://tools.hmcts.net/jira/browse/DMP-2812

Returning automated task by id

```
[ ] Yes
[x] No
```


src/main/resources/db/migration/common/V1_296__add_users_and_timestamps_to_automated_tasks.sql
davet1985, line 1: You'll need to bump the number as 296 already exists

src/main/resources/db/migration/common/V1_296__add_users_and_timestamps_to_automated_tasks.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE darts.automated_task
+SET created_ts = now(), last_modified_ts = now(), created_by = 0, last_modified_by = 0
+WHERE true;
```
davet1985, line 3: You can remove the `WHERE` clause, it's not needed.

davet1985 (changes requested):
Couple of minor comments

src/main/resources/db/migration/common/V1_296__add_users_and_timestamps_to_automated_tasks.sql
tonymort: I got errors when running without a where clause, this was directly against postgres so assumed it would be the case when part of a migration.  Unless it was a limitation imposed by datagrip

davet1985: I tried running one on pgadmin and it worked so possibly.

davet1985 (approved):

PR #1422, opened 2024-05-08
DMP-2868 ARM Push and Retention Date calculation - link transcription…

mario-paniccia:
https://tools.hmcts.net/jira/browse/DMP-2868

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/entity/TranscriptionEntity.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/entity/TranscriptionEntity.java
@@ -155,4 +155,16 @@ public class TranscriptionEntity extends CreatedModifiedBaseEntity {
         }
         return hearings.get(0);
     }
+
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/hmcts/darts/common/entity/TranscriptionEntity.java
```diff
@@ -155,4 +155,16 @@ public HearingEntity getHearing() {
         }
         return hearings.get(0);
     }
+
+    public List<CourtCaseEntity> associatedCourtCases() {
```
johnrace0, line 159: getter method should probably be getAssociatedCourtCases() ?

johnrace0, line 159: may be worth a comment to explain what the association is

johnrace0 (approved):

hemantasharma1129 (approved):

PR #1431, opened 2024-05-10
DMP-2941 Update Security Group endpoints to hide system users

johnrace0:
https://tools.hmcts.net/jira/browse/DMP-2941

### Change description ###
Update Security Group endpoints to hide system users

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/SecurityGroupServiceImpl.java
```diff
@@ -177,7 +179,38 @@ void updateSecurityGroupEntity(SecurityGroupPatch securityGroupPatch, SecurityGr
     }
 
     private void patchSecurityGroupUsers(SecurityGroupPatch securityGroupPatch, SecurityGroupEntity securityGroupEntity) {
```
chrisbellingham-hmcts, line 181: A subtle variation of AC3 is a PATCH request with:
```json
{
   "user_ids": []
}
```
This is used when we wish to remove all (non-system) users from the group.

I think with the implementation here, it would also remove system users. Could you check that please?

P.S I have added AC4 onto the ticket to cover this scenario.

src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/UserAccountStub.java
```diff
@@ -28,7 +29,7 @@ public class UserAccountStub {
 
     private static final int SYSTEM_USER_ID = 0;
     public static final String INTEGRATION_TEST_USER_EMAIL = "integrationtest.user@example.com";
-    private static final String SEPARATE_TEST_USER_EMAIL = "separateintegrationtest.user@example.com";
+    public static final String SEPARATE_TEST_USER_EMAIL = "separateintegrationtest.user@example.com";
```
chrisbellingham-hmcts, line 32: Can this be `private`?

johnrace0: it's now used in SecurityGroupControllerIntTest.java in the same way as INTEGRATION_TEST_USER_EMAIL so needs to be public

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/SecurityGroupServiceImpl.java
johnrace0: good spot. updated to comply with AC4

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/SecurityGroupServiceImpl.java
```diff
@@ -177,7 +179,38 @@ void updateSecurityGroupEntity(SecurityGroupPatch securityGroupPatch, SecurityGr
...
+
+            // filter the incoming list to remove system users
+            List<UserAccountEntity> patchUsers = userAccountRepository
+                .findByIdIn(securityGroupPatch.getUserIds());
+
+            if (userIds.size() > 0 && patchUsers.isEmpty()) {
+                throw new DartsApiException(
+                    UserManagementError.USER_NOT_FOUND,
+                    String.format("No User accounts found for patch user IDs %s", securityGroupPatch.getUserIds()));
+            }
```
chrisbellingham-hmcts, line 201: Could you help me understand the purpose of this one?

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/SecurityGroupServiceImpl.java
```diff
@@ -177,7 +179,38 @@ void updateSecurityGroupEntity(SecurityGroupPatch securityGroupPatch, SecurityGr
...
+                .map(UserAccountEntity::getId)
+                .toList();
+
+            // join the 2 lists - distinct
+            userIds = Stream.concat(systemUserIds.stream(), patchNonSystemUserIds.stream())
+                .distinct()
+                .collect(Collectors.toList());
+        }
+
         if (userIds != null) {
```
chrisbellingham-hmcts, line 214: The condtional on L214 duplicates the conditonal on L184, so can we just have a single conditional block for all this code?

Or it may be even cleaner to just return early if userIds is null, and avoid a layer of nesting altogether. E.g:

```java
    private void patchSecurityGroupUsers(SecurityGroupPatch securityGroupPatch, SecurityGroupEntity securityGroupEntity) {

        List<Integer> userIds = securityGroupPatch.getUserIds();
        if (userIds == null) {
            return;
        }

       // All the meat of the existing method goes here.
```

johnrace0: yeh. i think the confusion is in the re-use of userIds later on for the combined list. I've changed to return if userIds is initially null - and used a different name (combinedUserIds) for the combined list

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/SecurityGroupServiceImpl.java
johnrace0: this is to check that the list of user ids passed in to the patch contains valid users. i added a comment in the code

chrisbellingham-hmcts (approved):

PR #1443, opened 2024-05-14
DMP-3098 Edit event mappings

jackmaloney:
https://tools.hmcts.net/jira/browse/DMP-3098

### Change description ###
Allow an admin user to update an event mapping using the 'is_revision' query parameter on the POST event-mappings endpoint and set previous versions to inactive.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventControllerPostEventMappingTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventControllerPostEventMappingTest.java
@@ -13,6 +13,7 @@ import uk.gov.hmcts.darts.common.enums.SecurityRoleEnum;
 import uk.gov.hmcts.darts.testutils.GivenBuilder;
 import uk.gov.hmcts.darts.testutils.IntegrationBase;
 
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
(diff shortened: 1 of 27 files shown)


src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventControllerPostEventMappingTest.java
```diff
@@ -184,6 +212,20 @@ void eventMappingsPostEndpointShouldReturn409ErrorWhenDuplicateEventMapping() th
             .andReturn();
     }
 
+    @Test
+    void eventMappingsPostEndpointShouldReturn409ErrorWhenEventMappingDoesNotExistForIsRevisionTrue() throws Exception {
+        given.anAuthenticatedUserWithGlobalAccessAndRole(SUPER_ADMIN);
+
+        MockHttpServletRequestBuilder requestBuilder = post(EVENT_MAPPINGS_ENDPOINT + "?is_revision=true")
```
chrisbellingham-hmcts, line 219: Suggest using `.queryParam()` per previous comment.

src/main/java/uk/gov/hmcts/darts/common/repository/EventHandlerRepository.java
```diff
@@ -22,14 +22,12 @@ public interface EventHandlerRepository extends JpaRepository<EventHandlerEntity
...
-            AND eh.subType = :subType
-            AND eh.active = true
-        )
+        SELECT eh FROM EventHandlerEntity eh
+        WHERE eh.type = :type
+        AND eh.subType = :subType
+        AND eh.active = true
         """)
-    boolean doesActiveMappingForTypeAndSubtypeExist(String type, String subType);
+    List<EventHandlerEntity> findActiveMappingsForTypeAndSubtypeExist(String type, String subType);
```
chrisbellingham-hmcts, line 30: Would it be better to drop the `Exist` part of the method name: i.e `findActiveMappingsForTypeAndSubtype`?

src/main/resources/openapi/event.yaml
chrisbellingham-hmcts, line 1: On L228ish, we can specify `default: false` for `is_revision`. I.e if the param is not provided in the request, it will be set to `false`. This means we don't need to worry about null handling for this value in the service layer.

```yaml
        - in: query
          name: is_revision
          schema:
            type: boolean
            default: false
          description: "parameter to indicate if the request is an update to an existing mapping"
```

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventMappingServiceImpl.java
```diff
@@ -94,4 +119,8 @@ private EventMapping mapToEventMapping(EventHandlerEntity eventEntity) {
 
         return mapping;
     }
+
+    private boolean isUpdateToExistingMappingRequest(Boolean isRevision) {
+        return isRevision != null && isRevision;
+    }
```
chrisbellingham-hmcts, line 126: I think we can eliminate this method if the comment on `src/main/resources/openapi/event.yaml` can be incorporated.

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventMappingServiceImpl.java
```diff
@@ -51,12 +60,28 @@ public EventMapping postEventMapping(EventMapping eventMapping, Boolean isRevisi
             }
             var createdEventHandler = eventHandlerRepository.saveAndFlush(eventHandlerEntity);
 
+            if (isUpdateToExistingMappingRequest(isRevision)) {
+                updatePreviousVersionsToInactive(activeMappings);
+            }
```
chrisbellingham-hmcts, line 65: When trying to revise an existing type/subtype (with the application standing up locally), I get:

```json
{
	"title": "Internal Server Error",
	"status": 500,
	"detail": "could not execute statement [ERROR: duplicate key value violates unique constraint \"event_handler_event_type_event_event_sub_type_unq\"\n  Detail: Key (event_type, event_sub_type)=(101010101010, 10) already exists.] [insert into darts.event_handler (active,created_by,created_ts,event_name,handler,is_reporting_restriction,event_sub_type,event_type,evh_id) values (?,?,?,?,?,?,?,?,?)]; SQL [insert into darts.event_handler (active,created_by,created_ts,event_name,handler,is_reporting_restriction,event_sub_type,event_type,evh_id) values (?,?,?,?,?,?,?,?,?)]; constraint [event_handler_event_type_event_event_sub_type_unq]"
}
```

Can you reproduce that?

The following index constraint exists on the `event_handler` table:

```sql
    create unique index event_handler_event_type_event_event_sub_type_unq
    on event_handler (event_type, event_sub_type)
    where active;
```
So I wonder if we first need to set the prior revisions to inactive before saving the new entity:

```suggestion
            if (isUpdateToExistingMappingRequest(isRevision)) {
                updatePreviousVersionsToInactive(activeMappings);
            }
            
            var createdEventHandler = eventHandlerRepository.saveAndFlush(eventHandlerEntity);
```

 It's interesting this isn't showing up as an issue in the integration test.

chrisbellingham-hmcts (approved):

PR #1444, opened 2024-05-14
Dmp 2994 add translationqa role

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-2994

### Change description ###

Added translation qa role check to the case search view

```
[ ] Yes
[ x] No
```


src/integrationTest/java/uk/gov/hmcts/darts/cases/service/CaseServiceAdvancedSearchUseInterpreterTest.java
```diff
@@ -271,7 +271,7 @@ void testSearchCasesWithPermissionsGlobalTrueOnCourthouselessUserGroupUseInterpr
 
         // then
         var caseNumbers = resultList.stream().map(AdvancedSearchResult::getCaseNumber).toList();
-        assertThat(caseNumbers).containsExactlyInAnyOrder("Case1", "Case2", "Case4", "Case6");
+        assertThat(caseNumbers).containsExactlyInAnyOrder("Case1", "Case2", "Case3", "Case4", "Case5", "Case6");
```
mario-paniccia, line 274: I'm puzzled by this result. 
If we have set in line 263 use interpreter `true`, how is it possible that the user can access Swansea `Case1` that does not have an interpreter?

mario-paniccia: this might be a no longer valid test case now, since any user with translation QA + interpreter flag should not have any other role. Should we keep it or remove it?
Same consideration for some other test cases in this class.

src/integrationTest/java/uk/gov/hmcts/darts/cases/service/CaseServiceAdvancedSearchUseInterpreterTest.java
```diff
@@ -255,12 +255,12 @@ void testSearchCasesWithPermissionsGlobalTrueOnCourthouselessUserGroupUseInterpr
         // given
         var securityGroup = SecurityGroupTestData.buildGroupForRole(APPROVER);
         securityGroup.setGlobalAccess(true);
-        securityGroup.setUseInterpreter(true);
+        securityGroup.setUseInterpreter(false);
         assignSecurityGroupToUser(user, securityGroup);
 
         var securityGroupSwansea = SecurityGroupTestData.buildGroupForRoleAndCourthouse(TRANSLATION_QA, swanseaCourthouse);
         securityGroupSwansea.setGlobalAccess(false);
```
mario-paniccia, line 262: have we reviewed all the test cases in this class even if they still pass?
Perhaps they still pass but with the new change they should fail now? i.e. we could have false positives

src/integrationTest/java/uk/gov/hmcts/darts/cases/service/CaseServiceAdvancedSearchUseInterpreterTest.java
hemantasharma1129: agree, a translation QA role cant have any other role.

src/integrationTest/java/uk/gov/hmcts/darts/cases/service/CaseServiceAdvancedSearchUseInterpreterTest.java
hemantasharma1129: We can get rid of any test case which we have written with multiple role use case.

hemantasharma1129 (approved):

PR #1446, opened 2024-05-14
DMP-2952 - Fix to remove the problem detail signifiying to the client that the  …

mestebanez:

https://tools.hmcts.net/jira/browse/DMP-2952

### Change description ###

Prevent leaking that the user is a system user in the problem detail

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/usermanagement/service/validation/UserTypeValidator.java
+++ b/src/main/java/uk/gov/hmcts/darts/usermanagement/service/validation/UserTypeValidator.java
@@ -22,8 +22,7 @@ public class UserTypeValidator implements Validator<Integer> {
 
         if (accountEntity.isPresent() && accountEntity.get().getIsSystemUser()) {
             throw new DartsApiException(
-                UserManagementError.USER_NOT_FOUND,
-                String.format("User id %d is a system user", userId));
+                UserManagementError.USER_NOT_FOUND);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/darts/usermanagement/service/validation/UserTypeValidator.java
```diff
@@ -22,8 +22,7 @@ public void validate(Integer userId) {
 
         if (accountEntity.isPresent() && accountEntity.get().getIsSystemUser()) {
             throw new DartsApiException(
-                UserManagementError.USER_NOT_FOUND,
-                String.format("User id %d is a system user", userId));
+                UserManagementError.USER_NOT_FOUND);
```
mario-paniccia, line 25: I'd suggest we return the same error message that the existing USER_NOT_FOUND error uses, so the two are indistinguishable, i.e. [this](https://github.com/hmcts/darts-api/blob/8cd57fb5fe5d04bbcd6e06edfb97a6436abab953/src/main/java/uk/gov/hmcts/darts/usermanagement/service/validation/UserAccountExistsValidator.java#L21) message

mestebanez: @chrisbellingham-hmcts ?

chrisbellingham-hmcts: Agree, motivation for this change to keep the behaviour consistent so let's include the `String.format("User id %d not found", userId));` element.

mario-paniccia (approved):

PR #1452, opened 2024-05-15
DMP-2817 Get event mappings endpoint

jackmaloney:
https://tools.hmcts.net/jira/browse/DMP-2817

### Change description ###
Implementation of the GET /admin/event-mappings endpoint

```
[ ] Yes
[x ] No
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ A local jacoco coverage report can be generated using the following command:-
 The report will be available under ./build/jacocoHtml/index.html. The report incorporates both unit test
 and integration test coverage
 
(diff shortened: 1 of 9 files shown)


chrisbellingham-hmcts (commented):
Perhaps consider a functional test, simply asserting a 200 response and a non-zero length array in the response.

(I didn't suggest a functional test for the `POST`s because data management is trickier and can add a bit of complexity, but for a `GET` that has pre-populated data it's an easier win).

chrisbellingham-hmcts (approved):

PR #1454, opened 2024-05-15
DMP-3002 - Proposal for preventing duplicate daily list records

mestebanez:

https://tools.hmcts.net/jira/browse/DMP-3002

### Change description ###

A proposal in order to prevent duplicate judges, defenders and defendants. The solution assumes that the name is the basis of duplicate matching. 

I have proven that the json change surrounding  the integration tests DOES cause a failure with the original code in place and now passes with this change in place.

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/darts/dailylist/mapper/CitizenNameMapper.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private static final int FORENAME_INDEX = 0;
+
+    private static final int SURNAME_INDEX = 1;
+
+    public CitizenName getCitizenName(String name) {
+        String[] citizenName = name.split(NAME_DELIMITER);
+        CitizenName retCitizenName = new CitizenName();
+
+        if (citizenName.length == 2) {
```
chrisbellingham-hmcts, line 23: Is `> 2` ever possible?

src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListUpdater.java
```diff
@@ -235,7 +247,38 @@ private void addJudges(Sitting sitting, HearingEntity hearing) {
...
+    private boolean isDefenders(CourtCaseEntity courtCase, PersonalDetails defenders) {
+        boolean existingDefendant = false;
+        for (DefenceEntity defenceEntity : courtCase.getDefenceList()) {
+            if (citizenNameComparator.compare(defenders.getName(), citizenNameMapper.getCitizenName(defenceEntity.getName())) == 0) {
+                existingDefendant = true;
+            }
+        }
+
+        return existingDefendant;
+    }
```
chrisbellingham-hmcts, line 259: Maybe this could be a bit more readable, and the repetition of the `== 0` comparison refactored out? 

I think including "existing" in the method name would help readability of the calling code too: `isExistingDefender()`.

E.g. something like:

```java
    private boolean isExistingDefender(CourtCaseEntity courtCase, PersonalDetails defenders) {
        return courtCase.getDefenceList().stream()
            .anyMatch(defenceEntity -> isEqual(defenders.getName(), citizenNameMapper.getCitizenName(defenceEntity.getName())));
    }

    private boolean isEqual(CitizenName name1, CitizenName name2) {
        return citizenNameComparator.compare(name1, name2) == 0;
    }
```

And the `isEqual()` method could be used in the other similar methods. Just a thought.

src/main/java/uk/gov/hmcts/darts/dailylist/mapper/CitizenNameMapper.java
```diff
@@ -0,0 +1,36 @@
...
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.darts.dailylist.model.CitizenName;
+
+@Component
+@RequiredArgsConstructor
+@Slf4j
+public class CitizenNameMapper {
```
chrisbellingham-hmcts, line 11: Consider adding a unit test for this class.

src/main/java/uk/gov/hmcts/darts/dailylist/mapper/CitizenNameMapper.java
mestebanez: nope the creation of the name is also controlled by this class which takes the forename and surname and adds a space between them.

mestebanez: yep I looked at this again and the add case api does support adding  names of a non descript length (although the api should probably use citizen name for consistency. I have changed the code so it now consider surnames that consist of multiple values. This will avoid false positive matching when processing the daily list.

chrisbellingham-hmcts (approved):

PR #1461, opened 2024-05-16
DMP-2818 Minor tweak to empty handler logic

jackmaloney:


src/main/java/uk/gov/hmcts/darts/event/mapper/EventHandlerMapper.java
```diff
@@ -29,7 +29,7 @@ public EventHandlerEntity mapFromEventMappingAndMakeActive(EventMapping eventMap
     }
 
     private String getHandlerMappingOrNullHandler(String handler) {
-        return handler != null ? handler : DartsEventNullHandler.class.getSimpleName();
+        return handler != null && !handler.equalsIgnoreCase("") ? handler : DartsEventNullHandler.class.getSimpleName();
     }
```
mario-paniccia, line 33: how about we simplify using Guava `Strings.isNullOrEmpty()`?

mario-paniccia (approved):
approved with one suggestion if you want

src/main/java/uk/gov/hmcts/darts/event/mapper/EventHandlerMapper.java
chrisbellingham-hmcts: I was going to make the same suggestion as @mario-paniccia , but would suggest `StringUtils.isNotBlank()` from `apache.commons` which covers null, empty and blank cases.

src/test/java/uk/gov/hmcts/darts/event/mapper/EventHandlerMapperTest.java
```diff
@@ -43,11 +46,13 @@ void mapsEventMapperToEventHandlerEntityCorrectly() {
             .hasFieldOrPropertyWithValue("createdBy", USER_ACCOUNT_ENTITY);
     }
 
-    @Test
-    void mapsEventMapperToEventHandlerEntityCorrectlyWhenHandlerIsEmpty() {
+    @ParameterizedTest
+    @NullSource
+    @ValueSource(strings = {""})
```
chrisbellingham-hmcts, line 51: Perhaps also test for blank string e.g.:
```suggestion
    @ValueSource(strings = {"", " "})
```

PR #1469, opened 2024-05-19
DMP-3143 Remove batch size from application.yml and use DB for ARM automated tasks

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-3143

### Change description ###

Removed batch size from application.yml and changed the ARM automated tasks to get the batch size from the DB

```
[ ] Yes
[x ] No
```


cmagowan-hmcts (approved):

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmProcessorImpl.java
```diff
@@ -55,7 +55,7 @@ public UnstructuredToArmProcessorImpl(ExternalObjectDirectoryRepository external
                                           FileOperationService fileOperationService,
                                           ArchiveRecordService archiveRecordService) {
         super(objectRecordStatusRepository, userIdentity, externalObjectDirectoryRepository, externalLocationTypeRepository, dataManagementApi,
-              armDataManagementApi, fileOperationService);
+              armDataManagementApi, fileOperationService, 0);
```
hemantasharma1129, line 58: We should not hardcode it in the code even if its single mode. Have it passed in the constructor itself.

PR #1473, opened 2024-05-20
DMP-2767: Auditing admin courthouse operations

tonymort:
https://tools.hmcts.net/jira/browse/DMP-2767

Auditing admin courthouse operations

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/courthouse/service/CourthouseServiceImpl.java
```diff
@@ -285,11 +297,66 @@ public AdminCourthouse updateCourthouse(Integer courthouseId, CourthousePatch co
...
+
+    private static boolean securityGroupsUpdated(CourthousePatch courthousePatch, CourthouseEntity prePatchedEntity) {
+        if (courthousePatch.getSecurityGroupIds() == null) {
+            return false;
+        }
+        var prePatchedGroups = prePatchedEntity.getSecurityGroups().stream()
+            .map(SecurityGroupEntity::getId)
+            .toList();
+        return !prePatchedGroups.equals(courthousePatch.getSecurityGroupIds());
+    }
```
mario-paniccia, line 361: I might be missing something but I wonder whether we need to write this custom audit logic. 
The power of Envers is that it provides automatic auditing by just adding annotations. I would have thought that we could cover the basic admin court case operations requirements we have by using only annotations. 
And then use the Envers API to figure out what type of changes where made to the entity when the need arise to access those changes.

tonymort: This isn't related to envers, we have custom auditing also.

src/main/resources/application.yaml
```diff
@@ -86,7 +86,7 @@ spring:
     properties:
       hibernate:
         default_schema: ${spring.datasource.schema}
-        dialect: org.hibernate.dialect.PostgreSQLDialect
+#        dialect: org.hibernate.dialect.PostgreSQLDialect
```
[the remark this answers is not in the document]
tonymort: Will uncomment, according to documentation we shouldn't need this however, hibernate 6+ figures it out from the JDBC url.  We know we want PostgreSQLDialect so won't do any harm

src/main/java/uk/gov/hmcts/darts/courthouse/service/CourthouseServiceImpl.java
jackmaloney: I believe the requirement is to record that something happened in the 'audit' table and then if they want more detail about what actually changed, they can query the envers table.

mario-paniccia (approved):

PR #1490, opened 2024-05-23
DMP-3103 - Handle a duplicate automatic transcription event gracefully.

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-3103

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/event/service/handler/SentencingRemarksAndRetentionPolicyHandler.java
+++ b/src/main/java/uk/gov/hmcts/darts/event/service/handler/SentencingRemarksAndRetentionPolicyHandler.java
@@ -6,6 +6,7 @@ import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 import uk.gov.hmcts.darts.authorisation.api.AuthorisationApi;
 import uk.gov.hmcts.darts.common.entity.EventHandlerEntity;
+import uk.gov.hmcts.darts.common.exception.DartsApiException;
 import uk.gov.hmcts.darts.common.repository.CaseRepository;
 import uk.gov.hmcts.darts.common.repository.EventRepository;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/darts/event/service/handler/SentencingRemarksAndRetentionPolicyHandler.java
```diff
@@ -57,12 +57,22 @@ public void handle(final DartsEvent dartsEvent, EventHandlerEntity eventHandler)
...
+            var updateTranscription = new UpdateTranscriptionRequest();
+            updateTranscription.setTranscriptionStatusId(APPROVED.getId());
+            updateTranscription.setWorkflowComment("Transcription Automatically approved");
+            transcriptionsApi.updateTranscription(transcriptionResponse.getTranscriptionId(), updateTranscription);
 
-        var updateTranscription = new UpdateTranscriptionRequest();
-        updateTranscription.setTranscriptionStatusId(APPROVED.getId());
-        updateTranscription.setWorkflowComment("Transcription Automatically approved");
-        transcriptionsApi.updateTranscription(transcriptionResponse.getTranscriptionId(), updateTranscription);
+        } catch (DartsApiException dartsException) {
```
mario-paniccia, line 67: I wonder whether it would be better to change the `TranscriptionService.saveTranscriptionRequest` itself to just "do nothing" in case of duplicates, if that is not the new business rule. 
The `DuplicateRequestDetector` rather than throwing an exception (which will be always ignored now), can return a boolean used by the service to skip the save. 
This would probably lead to less code, and less duplication (having to catch/ignore/re-throw TranscriptionApiError.DUPLICATE_TRANSCRIPTION everywhere we use the service). It also avoids potential bugs in the future in case a new client of the service forgets to catch/ignore/re-throw

mario-paniccia (approved):
approved with one comment if you like

PR #1499, opened 2024-05-27
Dmp 3209.daily list request body

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-3209

DailyLists were too big to be transmitted via header, so changing to request body.

### DIFF
--- a/src/main/resources/openapi/dailylist.yaml
+++ b/src/main/resources/openapi/dailylist.yaml
@@ -12,6 +12,65 @@ servers:
 
 
 paths:
+  /dailylists/v2:
+    post:
+      tags:
+        - DailyLists
+      summary: Send Daily List
+      description: XHIBIT/CPP send daily case lists to the DAR PC via DARTS. These daily case lists inform the DAR PC which cases are being heard that day within the courthouse for all of its courtrooms.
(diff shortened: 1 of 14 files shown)


src/main/resources/openapi/dailylist.yaml
```diff
@@ -12,6 +12,65 @@ servers:
...
+              schema:
+                $ref: '#/components/schemas/PostDailyListResponse'
+
+        '500':
+          description: Internal Server Error
+          content:
+            application/json+problem:
+              schema:
+                $ref: './problem.yaml'
+
```
jackmaloney, line 73: Would the below version be removed from another PR, once Gateway updated?

davet1985 (approved):

src/main/resources/openapi/dailylist.yaml
cmagowan-hmcts: yeah, although we may want to revert the name of the new one back to being the original url, so thats up for debate. 
HMCTS way seems to be very against versioning:-
https://hmcts.github.io/restful-api-standards/#115

but i'm not sure their suggestion is compatible with swagger, and it looks overly complicated

PR #1504, opened 2024-05-29
Dmp 3104 transcription documents search

mestebanez:

https://tools.hmcts.net/jira/browse/DMP-3104

### Change description ###

Add a new endpoint that allows clients to search for transcription documents according to the associated requirements

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/common/repository/TranscriptionDocumentRepository.java
```diff
@@ -1,9 +1,57 @@
...
+
+import java.time.LocalDate;
+import java.time.OffsetDateTime;
+import java.util.List;
 
 @Repository
 public interface TranscriptionDocumentRepository extends JpaRepository<TranscriptionDocumentEntity, Integer> {
-}
+
+    @Query("""
```
mestebanez, line 16: @chrisbellingham-hmcts  a few things. The sub string searches on owner, requested by and courthouse display name has been made case insensitive

We left join here in order for the search to proceed under most eventualities. This does means that the hearing and case information may return null depending on the database data. Is this ok or do we a hard join?

mestebanez: Also if there is no courthouse connected directly to the transcription as well as the transcription having no hearings then the courthouse would be returned as null also

chrisbellingham-hmcts: A `transcription` will generally be related to one or more `hearing` via `hearing_transcription_ae`. This will always be the scenario for modernised.

However for some migrated legacy data however, there may be no `transcription` -> `hearing` link. In this scenario there would be a link between the `transcription` and one or many `court_case`, via `case_transcription_ae`.

So I think in either scenario the `courthouse` should be resolvable via `court_case`, either via the `transcription` -> `hearing` > `court_case` route or the `transcription` > `court_case` route.

@hemantasharma1129 is that the correct assessment based on our recent discussions?

src/main/java/uk/gov/hmcts/darts/transcriptions/mapper/TranscriptionResponseMapper.java
```diff
@@ -289,4 +294,51 @@ public GetTranscriptionDetailAdminResponse mapTransactionEntityToTransactionDeta
...
+
+        if (transcriptionDocumentResponse.caseId() != null) {
+            SearchTranscriptionDocumentResponseCase caseResponse = new SearchTranscriptionDocumentResponseCase();
+            caseResponse.setId(transcriptionDocumentResponse.caseId());
+            caseResponse.setCaseNumber(transcriptionDocumentResponse.caseNumber());
+            transformedMediaDetails.setCase(caseResponse);
+        }
+
+        // prioritise the courthouse that is connected directly transcription
+        if (transcriptionDocumentResponse.courthouseId() != null) {
```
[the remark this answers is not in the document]
chrisbellingham-hmcts: @hemantasharma1129, do you know if we'll always have a `transcription > courtroom > courthouse` link? Even for migrated data?

src/main/java/uk/gov/hmcts/darts/common/repository/TranscriptionDocumentRepository.java
hemantasharma1129: Agree with @chrisbellingham-hmcts explanation. Dont we have the method "getCourtHouse" already in TranscriptionEntity to fetch the courthouse which will cater for bot association table? 
@mestebanez

src/main/java/uk/gov/hmcts/darts/transcriptions/mapper/TranscriptionResponseMapper.java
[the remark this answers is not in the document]
hemantasharma1129: Are we trying to fetch the courthouse via the ctr_id of transcription table? Its a nullable column and migration may not be able to populate it for few cases. Use the method "getCourtHouse" from TranscriptionEntity to fetch the courthouse associated with a Tranacription. @mestebanez

src/main/java/uk/gov/hmcts/darts/common/repository/TranscriptionDocumentRepository.java
mestebanez: ok so by the looks of it transcription.courtroom is not needed to ascertain the court house. I will change.

mestebanez: we do have convenience methods for getting a courthouse and getting a court case but they are not useable in hql also they only give one record in the case of multiple existing. I have changed the query as per @chrisbellingham-hmcts comments. Hopefully this aligns

chrisbellingham-hmcts (approved):

PR #1505, opened 2024-05-29
DMP-3060: set updcase handler to null handler

tonymort:
Set updcase handler to null handler

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/db/migration/common/V1_304__update_updcase_handler.sql
+++ b/src/main/resources/db/migration/common/V1_304__update_updcase_handler.sql
@@ -1,3 +1,19 @@
 update event_handler
 set handler = 'DartsEventNullHandler'
 where (event_type = 'UPDCASE' and event_sub_type is null);
+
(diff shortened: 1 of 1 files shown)


src/main/resources/db/migration/common/V1_304__update_updcase_handler.sql
```diff
@@ -0,0 +1,19 @@
...
+set event_name = 'Appellant Read'
+where event_type = '20935' and event_sub_type = '10633';
+
+update event_handler
+set event_name = 'Appellant Read'
+where event_type = '20936' and event_sub_type = '10633';
+
+update event_handler
+set event_name = 'Defendant subject to an electronically monitored curfew'
+where event_type = '20936' and event_sub_type = '10633';
```
jackmaloney, line 19: This doesn't look right.

jackmaloney (approved):

PR #1518, opened 2024-05-31
DMP-2593: Auditing user and group management  

tonymort:
https://tools.hmcts.net/jira/browse/DMP-2593

Auditing user and group management  

```
[ ] Yes
[x] No
```


cmagowan-hmcts (approved):

src/main/java/uk/gov/hmcts/darts/common/entity/SecurityGroupEntity.java
```diff
@@ -33,41 +41,52 @@ public class SecurityGroupEntity extends CreatedModifiedBaseEntity {
     @SequenceGenerator(name = "grp_gen", sequenceName = "grp_seq", allocationSize = 1)
     private Integer id;
 
+    @NotAudited
```
chrisbellingham-hmcts, line 44: This should probably be audited - when we create a new security group it would be valuable to know which role was assigned to it (could be either TRANSCRIBER or TRANSLATION_QA).

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/UserManagementServiceImpl.java
```diff
@@ -79,6 +83,7 @@ public UserWithId createUser(User user) {
         userEntity.setLastModifiedDateTime(now);
 
         var createdUserEntity = userAccountRepository.save(userEntity);
+        auditApi.recordForUserAndCase(AuditActivity.CREATE_USER, authorisationApi.getCurrentUser(), null);
```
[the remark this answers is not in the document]
tonymort: yeah makes sense

src/main/java/uk/gov/hmcts/darts/common/entity/SecurityGroupEntity.java
chrisbellingham-hmcts: Disregard - role cannot be changed so there is no point in auditing it.

PR #1533, opened 2024-06-05
DMP-3254 - correct /admin/medias error messages

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-3254

Correct various error messages in the /admin/medias endpoint


chrisbellingham-hmcts (approved):
Would be worth adding some intTests covering the original ACs on https://tools.hmcts.net/jira/browse/DMP-2847.

PR #1536, opened 2024-06-05
DMP-2799: Auditing transcription workflows

tonymort:
https://tools.hmcts.net/jira/browse/DMP-2799

Auditing transcription workflows

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/transcriptions/auditing/TranscriptionUpdateAuditActivityProvider.java
```diff
@@ -0,0 +1,45 @@
...
+        return auditActivities;
+    }
+
+    private boolean isWorkflowStatusTransitioning(TranscriptionEntity entity, UpdateTranscriptionRequest patch) {
+        var latestWorkflow = entity.getTranscriptionWorkflowEntities().stream()
+            .min(comparing(TranscriptionWorkflowEntity::getWorkflowTimestamp));
+
+        return latestWorkflow.isEmpty()
+            || !Objects.equals(latestWorkflow.get().getTranscriptionStatus().getId(), patch.getTranscriptionStatusId());
+    }
```
mario-paniccia, line 44: this method could be on the `TranscriptionEntity` itself, if we are ok to have rich entities that are not just dump data holders

src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/TranscriptionServiceImpl.java
```diff
@@ -238,6 +241,9 @@ public UpdateTranscriptionAdminResponse updateTranscriptionAdmin(Integer transcr
             transcriptionStatusEntity,
             updateTranscription.getWorkflowComment()
         );
+
+        auditApi.recordAll(auditActivityProvider);
```
mario-paniccia, line 245: just for my knowledge, what is the benefit of having an `AuditApi` in addition to an `AuditService`? The service itself is providing an api for auditing.

davet1985: @mario-paniccia this was a pattern adopted fairly early on in the project. Exposing methods that are used outside of the package via the "Api" class (questionable naming). The idea behind this was to make it easier if we ever wanted to move packages out to some separate service, it would be easier to find and fix usages if they all go through one interface.

src/main/java/uk/gov/hmcts/darts/transcriptions/auditing/TranscriptionUpdateAuditActivityProvider.java
tonymort: It could, but so could many other methods in the codebase and we generally haven't taken that approach on this project.  I prefer a rich model design tbh especially instead of the endless private service method approach. Happy to discuss further.

src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/TranscriptionServiceImpl.java
tonymort: I'm not sure really, I think was an attempt to segregate the domain areas in a DDD style but never really took off.  Again happy to look at this.

src/main/resources/db/migration/common/V1_308__advanced_auditing_transcription_workflow.sql
```diff
@@ -0,0 +1,52 @@
+create table transcription_aud
```
mario-paniccia, line 1: these versions need to be bumped up, we have in master a V1_308 already

src/integrationTest/java/uk/gov/hmcts/darts/transcriptions/service/TranscriptionWorkflowAuditTest.java
```diff
@@ -0,0 +1,82 @@
...
+        assertThat(transcriptionCommentRevisions.getLatestRevision().getMetadata().getRevisionType()).isEqualTo(INSERT);
+    }
+
+    private AuditEntity findAuditActivity(String activity, List<AuditEntity> audits) {
+        return audits.stream()
+            .filter(audit -> activity.equals(audit.getAuditActivity().getName()))
+            .findFirst().orElseThrow();
+    }
+
+    private TranscriptionRequestDetails createTranscriptionRequestDetailsWithDefaults() {
```
mario-paniccia, line 67: could this be moved into one of the Test data classes ?

src/integrationTest/java/uk/gov/hmcts/darts/transcriptions/service/TranscriptionWorkflowAuditTest.java
```diff
@@ -0,0 +1,82 @@
...
+import static uk.gov.hmcts.darts.transcriptions.enums.TranscriptionUrgencyEnum.STANDARD;
+
+class TranscriptionWorkflowAuditTest extends IntegrationBase {
+
+    @Autowired
+    private TranscriptionService transcriptionService;
+
+    @Autowired
+    private GivenBuilder given;
+
```
mario-paniccia, line 28: shall we add a test that proves that non-admin transitions are not audited. If for the business is important to audit only operations triggered by admin

mario-paniccia (approved):
some minor comments if you want

src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/TranscriptionServiceImpl.java
mario-paniccia: I see. Seems a bit of boilerplate to me, and increases code complexity. But ok thanks for explaining this

src/main/java/uk/gov/hmcts/darts/transcriptions/auditing/TranscriptionUpdateAuditActivityProvider.java
mario-paniccia: If I had to chose between having this method on the entity and code duplication I'd go for having it in the entity.
I've seen a lot of code duplication in our codebase. Code duplication is hell

src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/TranscriptionServiceImpl.java
davet1985: This was the original presentation, by @chrisbellingham-hmcts Not sure if it helps at all.

[access_patterns.pdf](https://github.com/user-attachments/files/15772555/access_patterns.3.pdf)

src/integrationTest/java/uk/gov/hmcts/darts/transcriptions/service/TranscriptionWorkflowAuditTest.java
tonymort: That I'm aware of we don't currently have test data classes for api model data class, only entities.  I have been thinking about introducing something, I'm doing another auditing ticket so will look at introducing something then

src/integrationTest/java/uk/gov/hmcts/darts/transcriptions/service/TranscriptionWorkflowAuditTest.java
tonymort: For advanced auditing it will as we rely on annotations so the context of the route to the change has gone

PR #1543, opened 2024-06-06
DMP-2891- Dmp 2891 transcription document hide

mestebanez:

https://tools.hmcts.net/jira/browse/DMP-2891

### Change description ###

Add a new endpoint to show and hide transcription documents

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/AdminTranscriptionServiceImpl.java
```diff
@@ -134,4 +152,56 @@ public GetTranscriptionDocumentByIdResponse getTranscriptionDocumentById(Integer
...
+            objectAdminActionEntity.setHiddenDateTime(OffsetDateTime.now());
+            objectAdminActionEntity.setMarkedForManualDeletion(false);
+            objectAdminActionEntity.setMarkedForManualDelBy(userIdentity.getUserAccount());
+            objectAdminActionEntity.setMarkedForManualDelDateTime(OffsetDateTime.now());
+
+            objectAdminActionEntity = objectAdminActionRepository.saveAndFlush(objectAdminActionEntity);
+
+            response = transcriptionMapper.mapHideOrShowResponse(documentEntity, objectAdminActionEntity);
+        } else {
+            List<ObjectAdminActionEntity> objectAdminActionEntityLst = objectAdminActionRepository.findByTranscriptionDocument_Id(transcriptionDocumentId);
```
mestebanez, line 192: @chrisbellingham-hmcts in the case of a show i.e. not a hide should we return the existing admin action data (or is this only for hiding).

chrisbellingham-hmcts: Hey @mestebanez, so I think the scenario you're describing is "AC2 - Happy Path, Unhide a transcription document".

If we're going from Hidden to Unhidden, the response should reflect the new state of the system which would be Unhidden, i.e `is_hidden: false` with `admin_reason: null` (equivalent to no `admin_reason` being shown in the response).

hemantasharma1129 (approved):

PR #1551, opened 2024-06-10
DMP-3126 use azcopy to push raw data from unstructured to arm

mario-paniccia:
https://tools.hmcts.net/jira/browse/DMP-3126

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractUnstructuredToArmProcessor.java
+++ b/src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractUnstructuredToArmProcessor.java
@@ -1,7 +1,5 @@
 package uk.gov.hmcts.darts.arm.service.impl;
 
-import com.azure.core.util.BinaryData;
-import com.azure.storage.blob.models.BlobStorageException;
 import lombok.extern.slf4j.Slf4j;
 import uk.gov.hmcts.darts.arm.api.ArmDataManagementApi;
 import uk.gov.hmcts.darts.arm.service.UnstructuredToArmProcessor;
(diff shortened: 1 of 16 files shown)


src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractUnstructuredToArmProcessor.java
```diff
@@ -185,19 +181,11 @@ protected boolean copyRawDataToArm(ExternalObjectDirectoryEntity unstructuredExt
...
                 externalObjectDirectoryRepository.saveAndFlush(armExternalObjectDirectory);
             }
-        } catch (BlobStorageException e) {
-            if (e.getStatusCode() == BLOB_ALREADY_EXISTS_STATUS_CODE) {
-                log.info("BLOB raw data already exists {}", e.getMessage());
-            } else {
-                log.error("Failed to move BLOB data for file {} due to {}", unstructuredExternalObjectDirectory.getExternalLocation(), e.getMessage());
-                recoveryAction.run();
-                return false;
-            }
```
mario-paniccia, line 195: AzCopy will overwrite the files at the destination if they already exist, it won't throw an exception so this is not needed

src/main/resources/application.yaml
```diff
@@ -249,9 +249,11 @@ darts:
         unstructured: darts-unstructured
         inbound: darts-inbound-container
         outbound: darts-outbound
+        arm: dropzone
       sas-url:
         darts-inbound-container: ${DARTS_INBOUND_STORAGE_SAS_URL}
         darts-unstructured: ${DARTS_UNSTRUCTURED_STORAGE_SAS_URL}
+        dropzone: ${ARM_SAS_ENDPOINT}
```
hemantasharma1129, line 256: why are we using a new attribute for a SAS URL , I thought we already have one "sas-endpoint: ${ARM_SAS_ENDPOINT:}"

mario-paniccia: because the existing service we are re-using, `DataManagementService.copyBlobData`, expects sas urls to be in  `darts.storage.blob.sas-url.<container_name>`. See [here](https://github.com/hmcts/darts-api/blob/130362d6cf9161077b32af38323eebb39156f5c5/src/main/java/uk/gov/hmcts/darts/datamanagement/config/DataManagementConfiguration.java#L61).
I didn't want to change the existing code and add a special exception for arm, therefore making it less generic.

hemantasharma1129 (approved):

PR #1586, opened 2024-06-17
Dmp 3318.duplicate hearing media pk

cmagowan-hmcts:
remove duplicate case numbers before processing addAudio
 https://tools.hmcts.net/jira/browse/DMP-3318

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ bin/generated-sources/
 
 applicationinsights-agent-*.jar
 *.log
+/src/main/resources/application-perf.yaml
+/src/main/resources/application-dets.yaml
+/src/main/resources/application-demo.yaml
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -30,3 +30,6 @@ bin/generated-sources/
 
 applicationinsights-agent-*.jar
 *.log
+/src/main/resources/application-perf.yaml
+/src/main/resources/application-dets.yaml
+/src/main/resources/application-demo.yaml
```
davet1985, line 35: What are these for?

cmagowan-hmcts: these are files i have locally that i don't want to ever accidentally commit, so i am telling git to ignore.

davet1985: It would be better to add these to your `~/.gitignore` file, so they are only ignored for you.

chrisbellingham-hmcts (approved):

.gitignore
cmagowan-hmcts: removed

PR #1587, opened 2024-06-17
DMP-3260: Fix transcriber-view query to consider only latest transcri…

chrisbellingham-hmcts:
# [DMP-3260](https://tools.hmcts.net/jira/browse/DMP-3260)

The current implementation of the query underpinning `GET /transcriptions/transcriber-view&assigned=false` does not account for the possibility of a given transcription having many transcription_workflow records with the same workflow status. This scenario is possible now that Admin Portal allows a transcription to be rolled back, where trs progression could look something like `1 (REQUESTED) -> 2 (AWAITING AUTHORISATION) -> 3 (APPROVED) -> 5 (WITH TRANSCRIBER) -> 3 (APPROVED)`.

This fix covers only the issue reported by DMP-3260, however it is likely this issue exists in similar queries e.g. the `GET /transcriptions/transcriber-view&assigned=true` flow. A seperate review of such areas should be performed.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/transcriptions/controller/TranscriptionControllerGetTranscriberTranscriptsIntTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/transcriptions/controller/TranscriptionControllerGetTranscriberTranscriptsIntTest.java
@@ -158,7 +158,7 @@ class TranscriptionControllerGetTranscriberTranscriptsIntTest extends Integratio
                 "transcription_type": "Specified Times",
                 "status": "Approved",
                 "requested_ts": "2023-11-23T16:25:55.304517Z",
-                "state_change_ts": "2023-11-23T16:26:20.441633Z",
+                "state_change_ts": "2023-11-23T16:40:00Z",
                 "is_manual": true,
                 "transcription_urgency": {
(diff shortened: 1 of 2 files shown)


src/integrationTest/java/uk/gov/hmcts/darts/transcriptions/controller/TranscriptionControllerGetTranscriberTranscriptsIntTest.java
```diff
@@ -286,6 +286,10 @@ private void setupData(boolean generatedUrgency) {
                                 VALUES (42, 41, 2, -10, '2023-11-23 16:25:55.338405+00');
                                 INSERT INTO darts.transcription_workflow (trw_id, tra_id, trs_id, workflow_actor, workflow_ts)
                                 VALUES (43, 41, 3, -10, '2023-11-23 16:26:20.441633+00');
+                                INSERT INTO darts.transcription_workflow (trw_id, tra_id, trs_id, workflow_actor, workflow_ts)
```
mestebanez, line 289: Understand its not your problem but we may want to add a backlog ticket to refactor some of this. I guess we should be using stubs for data insertion

mestebanez (approved):

PR #1637, opened 2024-06-27
DMP-2623 Issue fixes

mario-paniccia:
https://tools.hmcts.net/jira/browse/DMP-2623

### Change description ###
fixed some little issues found in the generation of the case document

- null `createdDateTime` and `lastModifiedDateTime` CaseDocument fields
- null `createdBy` and `lastModifiedBy` CaseDocument fields
- removed "Entity" suffix from some case document fields
- renamed `id` to `caseId`
- renamed package from `...casedocument.template` to `...casedocument.model`

```
[ ] Yes
[ ] No
```


mestebanez (commented):
At first glance, I notice the package name template is not used anywhere else. Should this be in a model package

hemantasharma1129 (approved):

mestebanez (approved):
Only question I would have is whether src/integrationTest/java/uk/gov/hmcts/darts/casedocument/service/CourtCaseDocumentMapperIntTest.java can be a unit tested instead. 

As a rule we dont tend to integration test the mappers, although there are currently 5 of them:-

ArmResponseUploadFileMapperTest
CaseMapperTest
CasesMapperTest
CasesAnnotationMapperTest

I will approve the PR but put a question on the best practices channel.

mario-paniccia:
> CourtCaseDocumentMapperIntTest

it's a valid question. The case document mapper is different from the existing mappers you listed, because is a complex mapper that relies on 2 other mappers. Therefore we must have an integration test which verifies the correct integration of the 3 mappers that constitute the case document mapper.
At this point, it seemed pointless to me writing 3 separate unite tests for the 3 individual mappers, it would be a lot of effort to basically test the same things already tested by the integration test. So I've skipped them

PR #1642, opened 2024-06-28
DMP-3096: Using entity graph to force load lazy fields

tonymort:
https://tools.hmcts.net/jira/browse/DMP-3096?devStatusDetailDialog=branch

Using entity graph to force load lazy fields

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/repository/TransformedMediaRepository.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/repository/TransformedMediaRepository.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.darts.common.repository;
 
+import org.springframework.data.jpa.repository.EntityGraph;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.stereotype.Repository;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/darts/common/repository/TransformedMediaRepository.java
```diff
@@ -49,13 +50,14 @@ public interface TransformedMediaRepository extends JpaRepository<TransformedMed
     List<TransformedMediaDetailsDto> findTransformedMediaDetails(Integer userId, boolean expired);
 
 
+    @EntityGraph(attributePaths = {"mediaRequest.currentOwner.securityGroupEntities"})
```
chrisbellingham-hmcts, line 53: Not a comment for this PR, but I didn't realise this was a thing. Is this something we could be using more broadly to better control performance? Something worth exploring some best practices around?

@cmagowan-hmcts you aware of this one?

chrisbellingham-hmcts (approved):

src/main/java/uk/gov/hmcts/darts/common/repository/TransformedMediaRepository.java
tonymort: Yeah it is something we should have probably employed more, I've not used it much before but, there maybe other use cases other than forcing lazy loading outside of transactions.  Worth looking into.

davet1985 (approved):
Def looks like something we should be using more, good stuff

PR #1655, opened 2024-07-01
DMP-3054: creating event linked case for all case numbers on an event

tonymort:
https://tools.hmcts.net/jira/browse/DMP-3054

 creating event linked case for all case numbers on an event

```
[ ] Yes
[x] No
```


tonymort:
Duplicated

PR #1686, opened 2024-07-09
DMP-3283 Defendant unallocated case regex check

davet1985:

https://tools.hmcts.net/jira/browse/DMP-3283

### Change description ###

Do not add defendant name to case if matches unallocated case number regex

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/darts/cases/mapper/CasesMapper.java
```diff
@@ -43,6 +45,9 @@ public class CasesMapper {
     private final HearingReportingRestrictionsRepository hearingReportingRestrictionsRepository;
     private final CaseRetentionRepository caseRetentionRepository;
     private final AuthorisationApi authorisationApi;
+    private final LogApi logApi;
+    @Value("${darts.log.unallocated-case-regex}")
+    private final String unallocatedCaseRegex;
```
chrisbellingham-hmcts, line 50: Consider:

```suggestion
    private final Pattern unallocatedCaseRegex;
```
This has the advantage of validating the provided regex sytax upon application startup (when the property value is injected and the `CaseMapper` bean is created). If a junk regex is provided the application will fail fast, rather than waiting until the method is invoked at runtime.

src/main/java/uk/gov/hmcts/darts/cases/mapper/CasesMapper.java
```diff
@@ -85,7 +90,11 @@ public ScheduledCase mapToScheduledCase(HearingEntity hearing) {
     public CourtCaseEntity addDefendantProsecutorDefenderJudge(CourtCaseEntity caseEntity, AddCaseRequest caseRequest) {
 
         emptyIfNull(caseRequest.getDefendants()).forEach(newDefendant -> {
-            caseEntity.addDefendant(createNewDefendant(newDefendant, caseEntity));
+            if (newDefendant.matches(unallocatedCaseRegex)) {
```
chrisbellingham-hmcts, line 93: If going with the `Pattern` suggestion then:

```suggestion
            if (unallocatedCaseRegex.matcher(newDefendant).matches()) {
```

chrisbellingham-hmcts (approved):

PR #1687, opened 2024-07-09
DMP-3312: Add audio, use checksum to detect incoming duplicates

chrisbellingham-hmcts:
# [DMP-3312](https://tools.hmcts.net/jira/browse/DMP-3312)

Introduce a checksum validation to determine if the incoming "add audio" binary data is unique or a duplicate of previously uploaded data.

Ticket mentions checking the performance impact of the checksum computation. 

Comparison has been performed locally by attempting to add a duplicate using the `master` code vs the `dmp-3312-duplicate-audio-checksum` code on a development macbook with a 336MB audio file, which is near the max allowable size limit.

`master`: Request duration (average s): 2.8
`dmp-3312-duplicate-audio-checksum`: Request duration (average s): 1.7

Impact under these conditions is approx 1 second.

```
[ ] Yes
[X] No
```


hemantasharma1129 (approved):

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioUploadServiceImpl.java
```diff
@@ -175,24 +158,26 @@ private void versionUpload(List<MediaEntity> mediaToReplace,
...
-        List<MediaEntity> mediaEntitiesToVersion = new ArrayList<>();
-        for (MediaEntity mediaEntity : mediaEntities) {
-
-            // if file size is not expected
-            if (expectedFileSize != mediaEntity.getFileSize()) {
-                mediaEntitiesToVersion.add(mediaEntity);
-            }
+    private String computeChecksum(MultipartFile audioFileStream) {
+        try (var digestInputStream = new DigestInputStream(new BufferedInputStream(audioFileStream.getInputStream()), audioDigest.getMessageDigest())) {
+            ByteStreams.exhaust(digestInputStream); // We must consume the stream before calculating the checksum
```
mestebanez, line 170: Do we need to do this?  Doesnt the DigestInputStream allow us to get a checksum from the stream?

mestebanez (commented):
Non essential but I did wonder if would be worth introducing a utility to generate randomised bytes of varying sizes for test purposes, rather than checking in the specific media files.

mestebanez (approved):

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioUploadServiceImpl.java
chrisbellingham-hmcts: The checksum is computed incrementally as the stream is consumed. So to compute the checksum for the entire data, we first need to consume the entire stream.

chrisbellingham-hmcts:
> Non essential but I did wonder if would be worth introducing a utility to generate randomised bytes of varying sizes for test purposes, rather than checking in the specific media files.

Good idea, I'll look into that here.

PR #1689, opened 2024-07-09
DMP-3057 Return latest versioned events

jackmaloney:
https://tools.hmcts.net/jira/browse/DMP-3057

### Change description ###
When returning events to DARTS Portal, only return the latest version based on the is_current flag and if multiple exist, use the latest created date. 

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/hearings/mapper/GetEventsResponseMapper.java
```diff
@@ -26,5 +33,30 @@ public EventResponse mapToEvent(EventEntity eventEntity) {
...
+            .collect(Collectors.groupingBy(event -> Optional.ofNullable(event.getEventId())));
+
+        return groupedEvents.values().stream()
+            .flatMap(GetEventsResponseMapper::getLatestEvent)
+            .filter(Objects::nonNull)
+            .toList();
+    }
+
+    private Stream<EventEntity> getLatestEvent(List<EventEntity> group) {
+        if (group.get(0).getEventId() == null ||  group.get(0).getEventId() == 0) {
```
chrisbellingham-hmcts, line 48: Does this logic assume the incoming list of events is sorted by timestamp? If so, there's a risk of picking up the wrong event if the calling code doesn't ensure the data is pre-sorted. Consider explictly sorting the list here (or in `filterNonLatestEvents()`) before doing the conditional checks.

src/main/java/uk/gov/hmcts/darts/cases/mapper/EventMapper.java
```diff
@@ -5,20 +5,23 @@
...
 @UtilityClass
 public class EventMapper {
 
-    public List<Event> mapResponse(List<EventEntity> eventEntities) {
-        List<Event> response = new ArrayList<>();
-        for (EventEntity eventEntity : eventEntities) {
-            response.add(map(eventEntity));
-        }
-        return response;
+    public List<Event> mapToEvents(List<EventEntity> eventEntities) {
```
tonymort, line 19: Realise it was already like this but can be private

src/main/java/uk/gov/hmcts/darts/hearings/mapper/GetEventsResponseMapper.java
chrisbellingham-hmcts: After discussion with @jackmaloney , the above comment can be ignored. The order at this point does not matter - the EventIds will all be identical since they've been grouped within `filterNonLatestEvents()`.

chrisbellingham-hmcts (approved):

tonymort (approved):

PR #1720, opened 2024-07-15
DMP-3312: Simplify computing a checksum for streams that can be fully…

chrisbellingham-hmcts:
This change is related to DMP-3312 and has arisen after a discussion with @mario-paniccia.

An `InputStream` may be wrapped with a `DigestInputStream`. This is convienient if we wish to compute a checkum at the same time as reading the stream for some other purpose (e.g. uploading the data to server).

However, for applications where it is acceptable to consume the entire stream only for the purpose of computing a checksum (e.g. in a verification check), then a simpler approach can be taken that avoids the boilerplate of wrapping the stream with a `DigestInputStream`.

This PR implements that simplification, and removes methods that accept a `DigestInputStream` as there are no current usages.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/util/FileContentChecksum.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/util/FileContentChecksum.java
@@ -1,16 +1,12 @@
 package uk.gov.hmcts.darts.common.util;
 
-import com.google.common.hash.HashCode;
-import com.google.common.hash.Hashing;
-import com.google.common.io.ByteSource;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Component;
 
-import java.io.IOException;
+import java.io.FileInputStream;
 import java.io.InputStream;
 import java.nio.file.Path;
-import java.security.DigestInputStream;
 
 import static org.apache.commons.codec.digest.DigestUtils.md5;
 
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/darts/common/util/FileContentChecksum.java
```diff
@@ -40,18 +36,9 @@ public String calculate(InputStream inputStream) {
...
-    public String calculate(DigestInputStream digestInputStream) throws IOException {
-        return encodeToString(digestInputStream.getMessageDigest().digest());
-    }
-
     @SneakyThrows
     public String calculate(Path filePath) {
-        ByteSource byteSource = com.google.common.io.Files.asByteSource(filePath.toFile());
-        HashCode hc = byteSource.hash(Hashing.md5());
-        return encodeToString(hc.asBytes());
+        return calculate(new FileInputStream(filePath.toFile()));
```
chrisbellingham-hmcts, line 41: Author note: @mario-paniccia I've taken opportunity here to rework this method to use the same underlying `org.apache.commons.codec.digest.DigestUtils.md5()` method as the others in this class.

src/main/java/uk/gov/hmcts/darts/common/util/FileContentChecksum.java
```diff
@@ -40,18 +36,9 @@ public String calculate(InputStream inputStream) {
         return encodeToString(md5(inputStream));
     }
 
-    /**
-     * Calculates the digest when the source data has already been consumed.
-     */
-    public String calculate(DigestInputStream digestInputStream) throws IOException {
-        return encodeToString(digestInputStream.getMessageDigest().digest());
-    }
```
chrisbellingham-hmcts, line 48: Author note: Removed as there no usages now.

src/main/java/uk/gov/hmcts/darts/common/util/FileContentChecksum.java
mario-paniccia: nice one cheers @chrisbellingham-hmcts

mario-paniccia (approved):

PR #1721, opened 2024-07-15
Dmp 3496: event flag processing

mestebanez:

https://tools.hmcts.net/jira/browse/DMP-3496

### Change description ###

Add a new task to process a batch of events (containing the same event id) so that so that only one (the most recent event)  is set as the most current. All others are false

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/darts/event/service/impl/EventProcessorImpl.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Getter
+@RequiredArgsConstructor
+public class EventProcessorImpl implements EventProcessor {
+    private final Integer batchSize;
+    private final EventRepository eventRepository;
+    private final CurrentTimeHelper currentOffsetDateTime;
```
mario-paniccia, line 17: `CurrentTimeHelper` is not used, shall we remove it?

src/main/java/uk/gov/hmcts/darts/common/repository/EventRepository.java
```diff
@@ -74,4 +76,27 @@ Page<EventSearchResult> searchEventsFilteringOn(
...
         LocalDate hearingEndDate,
         Pageable pageable);
-}
+
+    @Query(value = """
+            select distinct evnt.event_id from darts.event evnt
+                inner join (SELECT distinct count(event_id) as eventCount, event_id  FROM
+                darts.event where is_current=true GROUP BY event_id) eventIdCntResult ON evnt.event_id = eventIdCntResult.event_id
+            where evnt.event_id <> 0 and eventIdCntResult.eventCount > 1
+            LIMIT :batchSize
```
mario-paniccia, line 85: rather than setting a limit manually ourselves I suggest we let JPA handle that automatically by using the standard JPA approach of doing limit queries using Pageable. We've already used this approach in other repos, and example [here](https://github.com/hmcts/darts-api/blob/78981fd7db29a3608ec8920f81039416f3a7ebc1/src/main/java/uk/gov/hmcts/darts/common/repository/CaseRepository.java#L60)

mario-paniccia: Pagination should work also with native queries in recent versions of spring boot, see [here](https://stackoverflow.com/questions/38349930/spring-data-and-native-query-with-pagination)

mario-paniccia: also I think this query can be replaced with more performant one which replaces the inner join with a HAVING clause:
```
SELECT distinct count(event_id) as eventCount, event_id  FROM
                darts.event where is_current=true 
				AND event_id <> 0 
				GROUP BY event_id
				having count(event_id) > 1
				order by event_id ASC
```

src/main/java/uk/gov/hmcts/darts/event/service/EventProcessor.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.hmcts.darts.event.service;
+
+import java.util.List;
+
+public interface EventProcessor {
+    List<Integer> processCurrentEvent();
```
mario-paniccia, line 6: shall we use more specific name? In the future we might introduce other types of events processors. 
We could call this for example `CleanupCurrentFlagEventProcessor` or sim

src/main/resources/db/migration/common/V1_332__add_task_cleanup_current_event.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO darts.automated_task (aut_id,task_name,task_description,cron_expression,cron_editable, batch_size)
+VALUES (nextval('aut_seq'),'CleanupCurrentEvent','Cleans up the event','0 * 0/4 * * *',true,50);
```
mario-paniccia, line 2: can we provide a more detailed description of what this process is doing please?
e.g. 
"Cleans up the event is_current flag"

mario-paniccia, line 2:
```suggestion
VALUES (nextval('aut_seq'),'CleanupEventCurrentFlag','Cleans up the event is_current flag','0 * 0/4 * * *',true,50);
```

src/integrationTest/java/uk/gov/hmcts/darts/common/repository/EventRepositoryTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Test
+    void testEventProcessing() {
+        dartsDatabase.createCase("Bristol", "case1");
+        dartsDatabase.createCase("Bristol", "case2");
+        dartsDatabase.createCase("Bristol", "case3");
+
+        Map<Integer, List<EventEntity>> eventIdMap = eventStub.generateEventIdEventsIncludingZeroEventId(3);
+
+        List<Integer> eventEntityReturned = eventRepository.getCurrentEventIdsToBeProcessed(1);
```
mario-paniccia, line 32: can we rename the variable please? the repository is not returning an entity but the event id
```suggestion
        List<Integer> eventIds = eventRepository.getCurrentEventIdsToBeProcessed(1);
```

src/main/java/uk/gov/hmcts/darts/common/repository/EventRepository.java
```diff
@@ -74,4 +76,27 @@ Page<EventSearchResult> searchEventsFilteringOn(
...
+        """, nativeQuery = true)
+    List<Integer> getCurrentEventIdsToBeProcessed(int batchSize);
+
+    @Query(value = """
+             select eve_id FROM darts.event e where event_id=:eventId order by created_ts desc LIMIT 1
+        """, nativeQuery = true)
+    Integer getTheCurrentEventPrimaryKeyForEventId(int eventId);
```
mario-paniccia, line 92:
```suggestion
    Integer getLatestEventPrimaryKeyForEventId(int eventId);
```

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventProcessorImpl.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Override
+    public List<Integer> processCurrentEvent() {
+        List<Integer> currentEventLst = new ArrayList<>();
```
mario-paniccia, line 21:
```suggestion
        List<Integer> processedEventLst = new ArrayList<>();
```

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventProcessorImpl.java
```diff
@@ -0,0 +1,31 @@
...
+        List<Integer> eventEntityReturned = eventRepository.getCurrentEventIdsToBeProcessed(batchSize);
+        eventEntityReturned.stream().forEach(event -> {
```
mario-paniccia, line 23:
```suggestion
        eventEntityReturned.forEach(event -> {
```

mario-paniccia, line 23: for a batch size of X this would result in 2X distinct sql statements to the DB which can be a bit slow. I was wondering if it's possible to remove this loop and instead send the DB 1 or 2 sqls only that update in one go all the necessary rows

src/main/resources/db/migration/common/V1_332__add_task_cleanup_current_event.sql
mario-paniccia: is this cron correct? for the `ApplyRetention` task which needs to run every hours we've used this expressions: ```
0 0 * * * *
``` 
which seems to be ok according to [here](https://stackoverflow.com/questions/70107416/running-cron-job-task-at-every-hour)

**Review by mario-paniccia — COMMENTED** [review]
some comments please

File: src/main/resources/db/migration/common/V1_332__add_task_cleanup_current_event.sql (hunk 4 above)
Comment (mestebanez) [reply src/main/resources/db/migration/common/V1_332__add_task_cleanup_current_event.sql:2]: ok i'll change wasnt sure about he correct cron expression to use

File: src/main/java/uk/gov/hmcts/darts/common/repository/EventRepository.java (hunk 6 above)
Comment (mestebanez) [reply src/main/java/uk/gov/hmcts/darts/common/repository/EventRepository.java:92]: Was trying to use the terminology of the database. This is the primary key of the current event

File: src/main/java/uk/gov/hmcts/darts/event/service/impl/EventProcessorImpl.java (hunk 8 above)
Comment (mestebanez) [reply src/main/java/uk/gov/hmcts/darts/event/service/impl/EventProcessorImpl.java:31]: i'll look but your only going to get batch size at most which shouldnt be that large.

File: src/main/java/uk/gov/hmcts/darts/event/service/impl/CleanupCurrentFlagEventProcessorImpl.java [hunk 9]
```diff
@@ -0,0 +1,49 @@
...
+@RequiredArgsConstructor
+@Slf4j
+public class CleanupCurrentFlagEventProcessorImpl implements CleanupCurrentFlagEventProcessor {
+    private final Integer batchSize;
+    private final EventRepository eventRepository;
+
+    @Override
+    public List<Integer> processCurrentEvent() {
+        List<Integer> processedEventIdLst = new ArrayList<>();
+        log.info("Batch size to process event ids for {}", batchSize);
```
mario-paniccia, line 24: is all this INFO logging really needed? I'm thinking that it could produce too much noise in production. Perhaps we can set some of these to DEBUG?

mario-paniccia (approved):
one minor comment about reducing info logging if you want

src/main/java/uk/gov/hmcts/darts/event/service/impl/CleanupCurrentFlagEventProcessorImpl.java
mestebanez: I will change to debug for now. Thanks

PR #1724, opened 2024-07-16
DMP-3330: Add validation for PATCH /admin/retention-policy-types to p…

chrisbellingham-hmcts:
# [DMP-3330](https://tools.hmcts.net/jira/browse/DMP-3330)

Add validation to preempt a `retention_policy_type_type_unq` constraint violation upon changing a fixed policy key.

```
[ ] Yes
[ ] No
```


src/integrationTest/java/uk/gov/hmcts/darts/testutils/PostgresIntegrationBase.java
```diff
@@ -30,13 +30,16 @@ public class PostgresIntegrationBase {
...
     static void configureProperties(DynamicPropertyRegistry registry) {
-
         registry.add("spring.datasource.url", POSTGRES::getJdbcUrl);
         registry.add("spring.datasource.username", POSTGRES::getUsername);
         registry.add("spring.datasource.password", POSTGRES::getPassword);
     }
 
     static {
+        // Keep an eye on this, if it needs to be increased as more tests utilise PostgresIntegrationBase then we may have a connection leak somewhere
+        final int serverMaxConnections = 50;
```
mestebanez, line 40: Why do we need to do this when running a single threaded test? Just curious.

mestebanez, line 40: This could be added to the integration test config or at least a final static of the class.

src/test/java/uk/gov/hmcts/darts/retention/service/impl/RetentionGetPolicyTypesServiceImplTest.java
```diff
@@ -77,6 +80,7 @@ private void setupStubs() {
             authorisationApi,
             policyDurationValidator,
             createPolicyTypeValidator,
+            editPolicyTypeValidator,
```
mestebanez, line 83: We could also change the tests to validate that the mockito validator object is being called.

src/integrationTest/resources/application-intTest.yaml
```diff
@@ -33,6 +33,9 @@ spring:
     multipart:
       max-file-size: 1KB
       max-request-size: 10KB #this affects addAudio as well, so needs to handle a 350mb mp2 file
+  datasource:
```
mestebanez, line 36: Is the default pool size of 50 causing the tests problems? Just curious. Maybe worth explaining it in a comment.

src/integrationTest/java/uk/gov/hmcts/darts/testutils/PostgresIntegrationBase.java
chrisbellingham-hmcts: There's a bit of a story to this! When I moved the `RetentionGetPolicyTypesServiceImplTest` over to use Postgres, the `./gradlew build` would consistently fail at the integration test stage with:

```
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'entityManagerFactory' defined in class path resource [org/springframework/boot/autoconfigure/orm/jpa/HibernateJpaConfiguration.class]: Failed to initialize dependency 'flywayInitializer' of LoadTimeWeaverAware bean 'entityManagerFactory': Error creating bean with name 'flywayInitializer' defined in class path resource [org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration$FlywayConfiguration.class]: Unable to obtain connection from database: FATAL: sorry, too many clients already
---------------------------------------------------------------------------------
SQL State  : 53300
Error Code : 0
Message    : FATAL: sorry, too many clients already
```

As I understand it, there are a two aspects to the connection config:
- `spring.datasource.hikari.maximum-pool-size` defines the maximum number of connections that our test client can impose against the DB server. This value is set in `application.yaml` to 50.
- `max-connections` can be configured on the DB side, if the number of client connections exceeds this value then we'll see the above error. The value defaults to 100 for the `postgres:alpine-15` image.

To avoid the "too many clients" issue, this PR:
- Sets `maximum-pool-size` to 10, which is a downgrade from the 50 configured for the main app. This seems reasonable (and maybe excessive), as you say the tests are single threaded so we shouldn't ever need 50 connections (or even 10?).
- Sets `max-connections` DB-side to 50, which is a downgrade from the 100 configured by default. Interestingly, this is about as low as I can go without seeing the exception above. I do know that Postgres expects some (non-client) connections to be reserved for internal use, so maybe that's why the extra overhead (of 40??) connections is needed, but not clear.

There remains the root cause of why I had to do this, and I'm not sure on that. The fact the test suite fails when I added a new test class using `PostgresIntegrationBase` leads me to believe the new test has increased the connection demand on the db server, which shouldn't happen. If we truely do have a connection leak somewhere then we should see this problem again when usage of `PostgresIntegrationBase` becomes more common - this is why I've lowered `max-connections` to the lowest tolerable value, so that we get that feedback relatively quickly.

chrisbellingham-hmcts: Will declare as a class constant. I thought about adding it to the config yaml, but I decided against it as I don't want to encourage changes to it  - I think it shouldn't need to be changed unless we have a connection leak problem.

src/integrationTest/resources/application-intTest.yaml
chrisbellingham-hmcts: Add comment:

```yaml
maximum-pool-size: 10 # Overrides the value in `application.yaml` to a smaller value more appropriate for integration tests
```

src/test/java/uk/gov/hmcts/darts/retention/service/impl/RetentionGetPolicyTypesServiceImplTest.java
chrisbellingham-hmcts: I have now added `verifyNoInteractions(editPolicyTypeValidator)` to each test in here - the test is concerned only with "getting" the policies, not editing them.

src/integrationTest/java/uk/gov/hmcts/darts/testutils/PostgresIntegrationBase.java
mestebanez: ah ok makes sense to me. Like you say ideally we would work out why we need to do this but realise thats outside the scope of this ticket

mestebanez (approved):

PR #1726, opened 2024-07-16
DMP-3501 - Advanced search - allow a courthouse and specific date to …

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-3501

Adding an extra point for a specific date.

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/cases/util/RequestValidator.java
+++ b/src/main/java/uk/gov/hmcts/darts/cases/util/RequestValidator.java
@@ -40,6 +40,7 @@ public class RequestValidator {
         totalPoints += StringUtils.length(request.getJudgeName()) >= 3 ? 1 : 0;
         totalPoints += StringUtils.length(request.getDefendantName()) >= 3 ? 1 : 0;
         totalPoints += (request.getDateFrom() != null || request.getDateTo() != null) ? 1 : 0;
+        totalPoints += (request.getDateFrom() != null && request.getDateFrom().equals(request.getDateTo())) ? 1 : 0;//specific date
         totalPoints += StringUtils.length(request.getEventTextContains()) >= 3 ? 1 : 0;
 
         int three = 3;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/darts/cases/util/RequestValidator.java
```diff
@@ -40,6 +40,7 @@ private static void checkComplexity(GetCasesSearchRequest request) {
         totalPoints += StringUtils.length(request.getJudgeName()) >= 3 ? 1 : 0;
         totalPoints += StringUtils.length(request.getDefendantName()) >= 3 ? 1 : 0;
         totalPoints += (request.getDateFrom() != null || request.getDateTo() != null) ? 1 : 0;
+        totalPoints += (request.getDateFrom() != null && request.getDateFrom().equals(request.getDateTo())) ? 1 : 0;//specific date
```
mestebanez, line 43: It would be good to make some of these hard coded values final statics. The concept of points is not all together clear to me. If we used a final static we could avoid the use of magic numbers and explain their purpose i.e. 0 = CRITERIA_TOO_BROAD, 1 = CRITERIA_IS_SPECIFIC. 3 = SPECIFICITY_THRESHOLD.

mestebanez (approved):

src/main/java/uk/gov/hmcts/darts/cases/util/RequestValidator.java
mestebanez: Approved the PR regardless.

cmagowan-hmcts: 0 and 1 are not considered magic numbers, e.g.:-
https://codehs.com/glossary/term/14

there is a confluence pahe here to try to explain why we are using the points system:-
https://tools.hmcts.net/confluence/display/DMP/Advanced+Search+Restrictions+in+Modernised

This page was updated for the new rule

cmagowan-hmcts: i had thought about extracting the IF-statements out to a named boolean to explain what each one was, but i wasn't sure if there was a lot of benefit, but can do if you think it would aid understanding.

PR #1727, opened 2024-07-17
DMP-3448 - populate security Groups

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-3448

### DIFF
--- a/src/main/resources/db/migration/postgres/V1_333_1__insert_security_groups.sql
+++ b/src/main/resources/db/migration/postgres/V1_333_1__insert_security_groups.sql
@@ -1,0 +1,33 @@
+DO $$
+DECLARE
+	recordToBeMigrated record;
+	newId integer;
+	oldId integer;
+BEGIN
+FOR recordToBeMigrated IN
+    SELECT grp_id from security_group where grp_id >0
+    order by 1
+	LOOP
+	select recordToBeMigrated.grp_id+1000000 into newId;
+	select recordToBeMigrated.grp_id into oldId;
+	RAISE NOTICE 'migrating % to %', oldId, newId;
+
+	--insert copy
+	INSERT INTO security_group(
+		grp_id, rol_id, group_name, is_private, description, group_global_unique_id, global_access, display_state, use_interpreter, display_name, dm_group_s_object_id, created_ts, created_by, last_modified_ts, last_modified_by)
+		select newid, rol_id, group_name, is_private, description, group_global_unique_id, global_access, display_state, use_interpreter, display_name, dm_group_s_object_id, created_ts, created_by, last_modified_ts, last_modified_by from security_group where grp_id = oldId;
+
+
+	--move entries
+	--"security_group_courthouse_ae"
+    update security_group_courthouse_ae set grp_id = newId where grp_id = oldId;
+    	--"security_group_user_account_ae"
+	update security_group_user_account_ae set grp_id = newId where grp_id = oldId;
+
+
+	--delete old record
+	delete from security_group where grp_id = oldId;
+
+    END LOOP;
+
+END $$;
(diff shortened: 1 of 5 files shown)


hemantasharma1129 (approved):

src/main/resources/db/migration/postgres/V1_333_1__insert_security_groups.sql
```diff
@@ -0,0 +1,33 @@
...
+DECLARE
+	recordToBeMigrated record;
+	newId integer;
+	oldId integer;
+BEGIN
+FOR recordToBeMigrated IN
+    SELECT grp_id from security_group where grp_id >0
+    order by 1
+	LOOP
+	select recordToBeMigrated.grp_id+1000000 into newId;
```
mestebanez, line 11: Its probably never likely to be a problem but it may be worth making the number negative rather than positive. The sequence number we use always increments so by inserting negative numbers we are guranteed to never clash

src/main/resources/db/migration/postgres/V1_333_3__insert_security_groups.sql
```diff
@@ -0,0 +1,63 @@
...
+DECLARE
+	recordToBeMigrated record;
+	newId integer;
+	oldId integer;
+	maxId integer;
+BEGIN
+perform setval('grp_seq', 1000, false);
+RAISE NOTICE 'migrating duplicate records';
+FOR recordToBeMigrated IN
+    SELECT sg1.grp_id newId, sg2.group_name, sg2.grp_id oldId FROM security_group sg1, security_group sg2
```
mestebanez, line 11: Add a comment here as to what we are doing. Not clear that old is greater than 1000 and new is less than 1000

mestebanez: Maybe worth assigning 1000 a variable name and using that

src/main/resources/db/migration/postgres/V1_333_2__insert_security_groups.sql
```diff
@@ -0,0 +1,1054 @@
+INSERT INTO darts.security_group(
```
mestebanez, line 1: Should we use current_timestamp for the date rather than hard coding?

src/main/resources/db/migration/postgres/V1_333_1__insert_security_groups.sql
cmagowan-hmcts: i've been asked to leave the negative numbers alone, hence the `and sg1.grp_id > 0` clause in src/main/resources/db/migration/postgres/V1_333_3__insert_security_groups.sql

Remember, these scripts are onyl going to be run against the test databases now, so unless there is data right now that is over 1000000, then it wont matter.

src/main/resources/db/migration/postgres/V1_333_3__insert_security_groups.sql
cmagowan-hmcts: comment and constant added.

src/main/resources/db/migration/postgres/V1_333_3__insert_security_groups.sql
```diff
@@ -0,0 +1,63 @@
...
+	where upper(sg1.group_name) = upper(sg2.group_name)
+    and sg1.grp_id < sg2.grp_id
+	 and sg1.grp_id > 0
+     and sg1.grp_id < 1000
```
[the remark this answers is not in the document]
cmagowan-hmcts: done

src/main/resources/db/migration/postgres/V1_333_2__insert_security_groups.sql
cmagowan-hmcts: we wanted to make it obvious that the data was taken from Production and done via migration pre-go-live. current_timestamp would be more applicable after-go-live i think. i.e. it is really obvious that the data has been manually populated as the date will be exactly 1st January 00:00.

PR #1742, opened 2024-07-22
DMP-3365: Remove duplicate events

tonymort:
https://tools.hmcts.net/jira/browse/DMP-3365

Remove duplicate events 

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/event/service/impl/RemoveDuplicateEventsProcessorImpl.java
```diff
@@ -0,0 +1,103 @@
...
+    @Override
+    @Transactional
+    public void processRemoveDuplicateEvents() {
+        var today = currentTimeHelper.currentLocalDate();
+        var startDateTime = today.minusDays(clearUpWindow).atStartOfDay().atOffset(UTC);
+        if (startDateTime.isBefore(earliestRemovableEventDate)) {
+            startDateTime = earliestRemovableEventDate;
+        }
+        var endDateTime = today.atTime(23, 59).atOffset(UTC);
+        var duplicateEvents = new ArrayList<EventEntity>();
```
chrisbellingham-hmcts, line 67: I'm wondering if the approach of taking the `startDateTime` as `00:00:00` has some consequences if `clear-up-window` is ever configured to `0`.

The task would execute at 22:00, capturing all events between 00:00 and 23:59 that day. Practically there should be no events (?) between 22:00 and 23:59, as they'd be in the future, so really we're getting events between 00:00 and 22:00.

Events would continue to flow into the system between 22:00 and 23:59, but the next invocation of the task (the next day) wouldn't capture them. So our cleanup would only be partially cleaning up the day's events.

I think if `clear-up-window` is > 0, then the problem is solved because the previous day will captured in it's entirety, which will scoop up any missed events.

Do we need some validation to ensure `clear-up-window` is > 0?

src/main/java/uk/gov/hmcts/darts/common/repository/EventRepository.java
```diff
@@ -104,4 +106,13 @@ HAVING count(event_id) > 1
...
         List<Integer> eventIdsPrimaryKeysLst, List<Integer> eventIdLst);
+
+    @EntityGraph(attributePaths = {"id", "timestamp", "eventId", "messageId", "eventText"})
+    @Query("""
+        SELECT ee
+        FROM EventEntity ee
+        WHERE ee.timestamp >= :startDateTime
+        AND ee.timestamp <= :endDateTime
+        """)
+    List<EventEntity> findAllForDuplicateProcessing(OffsetDateTime startDateTime, OffsetDateTime endDateTime);
```
chrisbellingham-hmcts, line 117: Maybe a more generic method name would be better, as stricly speaking the query doesn't care about dupes. E.g. `findAllBetweenDateTimesInclusive()`

src/main/java/uk/gov/hmcts/darts/event/service/impl/RemoveDuplicateEventsProcessorImpl.java
tonymort: Good point, will have think about this.

src/main/resources/application.yaml
```diff
@@ -187,6 +187,9 @@ darts:
   dar-pc-notification:
     url-format: http://{0}/VIQDARNotifyEvent/DARNotifyEvent.asmx
   events:
+    duplicates:
+      earliest-removable-event-date: ${MODERNISED_DARTS_START_DATE:2024-07-16}
+      clear-up-window: 2
```
davet1985, line 192: Is this specified in days?

Think it might be worth adding a env var here so that it can be overridden in flux if needed.

```suggestion
      clear-up-window: ${DUPLICATE_EVENTS_CLEAR_UP_WINDOW_DAYS:2}
```

davet1985 (changes requested):
A few comments on the config

src/main/resources/application.yaml
```diff
@@ -187,6 +187,9 @@ darts:
   dar-pc-notification:
     url-format: http://{0}/VIQDARNotifyEvent/DARNotifyEvent.asmx
   events:
+    duplicates:
+      earliest-removable-event-date: ${MODERNISED_DARTS_START_DATE:2024-07-16}
```
jackmaloney, line 191: Have you got a flux change to set this per environment? On prod, we want to make sure this is clear it needs to be set to the go live date, so probably needs a comment explaining.

src/main/java/uk/gov/hmcts/darts/event/service/impl/RemoveDuplicateEventsProcessorImpl.java
```diff
@@ -0,0 +1,109 @@
...
+            List<EventEntity> chunk;
+
+            while (spliterator.estimateSize() > 0) {
+                chunk = stream(spliterator, false)
+                    .limit(CHUNK_SIZE)
+                    .collect(toList());
+
+                if (!chunk.isEmpty()) {
+                    var caseManagementIdsToBeDeleted = caseManagementRetentionRepository.getIdsForEvents(chunk);
+                    caseRetentionRepository.deleteAllByCaseManagementIdsIn(caseManagementIdsToBeDeleted);
```
jackmaloney, line 95: Should we add a log message indicating which events have been removed for support purposes?

jackmaloney: Ok, maybe move it and make clear that we are deleting these events in the message?

src/main/resources/application.yaml
tonymort: https://github.com/hmcts/sds-flux-config/pull/5139

jackmaloney (approved):

PR #1744, opened 2024-07-22
DMP-3292: Task that manages the inbound store transcription and annotation deletion

mestebanez:

https://tools.hmcts.net/jira/browse/DMP-3292

### Change description ###

A task that deletes the inbound transcription and annotation arm media that exceeds duration of x hours

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
```diff
@@ -367,9 +368,32 @@ boolean hasMediaNotBeenCopiedFromInboundStorage(MediaEntity media, ObjectRecordS
...
+        value = """
+         SELECT od.eod_id FROM darts.external_object_directory od WHERE od.med_id IN
+            (
+                SELECT med_id FROM darts.external_object_directory
+                WHERE elt_id=1 AND ors_id=2
+            ) AND elt_id=3 AND ors_id=2 
+            AND (floor (extract (epoch from (current_timestamp::timestamp - last_modified_ts::timestamp)) / 60 / 60) >= :hoursInArm)
+        """, nativeQuery = true
+    )
+    List<Integer> findAllInboundArmMediaExceedingHours(Pageable resultLimit, Integer hoursInArm);
```
cmagowan-hmcts, line 381: i think it's better to reuse the method that a few of the other jobs are already using:-
findMediaFileIdsIn2StorageLocationsBeforeTime

I don't see a batch size mentioned in the ticket. is this a requirement?

See InboundAudioDeleterProcessorImpl as an example

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
```diff
@@ -367,9 +368,32 @@ boolean hasMediaNotBeenCopiedFromInboundStorage(MediaEntity media, ObjectRecordS
...
+    @Modifying
+    @Transactional
+    @Query(
+        value = """
+        UPDATE ExternalObjectDirectoryEntity eod
+        SET status=:status, lastModifiedBy = :userToUpdate, lastModifiedDateTime=current_timestamp
+        WHERE id in :externalObjectDirectoryIdLst
+        """
+    )
+    void updateStatusAndUserOfObjectDirectory(List<Integer> externalObjectDirectoryIdLst, ObjectRecordStatusEntity status, UserAccountEntity userToUpdate);
```
cmagowan-hmcts, line 392: there already exists a very similar updateStatus method.

src/main/java/uk/gov/hmcts/darts/arm/service/impl/InboundTranscriptionAndAnnotationDeleterProcessorImpl.java
```diff
@@ -0,0 +1,49 @@
...
+    public List<Integer> processDeletionIfAfterHours(int batch, int hourThreshold) {
+        List<Integer> armRecordToBeMarkedForDeletion
+            = externalObjectDirectoryRepository.findAllInboundArmMediaExceedingHours(Pageable.ofSize(batch), hourThreshold);
+
+        ObjectRecordStatusEntity status = objectRecordStatusRepository.getReferenceById(ObjectRecordStatusEnum.MARKED_FOR_DELETION.getId());
+
+        log.debug("Identified records to be deleted  {}",  armRecordToBeMarkedForDeletion.stream().map(Object::toString));
+        externalObjectDirectoryRepository.updateStatusAndUserOfObjectDirectory(armRecordToBeMarkedForDeletion, status,
+                                                                               systemUserHelper.getHousekeepingUser());
+        log.debug("Records have been marked as deleted");
```
cmagowan-hmcts, line 45: They have not been `marked as deleted`, they have been `marked for deletion`

src/main/resources/db/migration/common/V1_334__add_task_inbound_transcription_annotation_deletion.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO darts.automated_task (aut_id,task_name,task_description,cron_expression,cron_editable, batch_size)
+VALUES (nextval('aut_seq'),'InboundTranscriptionAnnotationDeleter','Marks for deletion the transcription and annotation that is stored in inbound for the longer than a specific date.','0 0 22 * * *',true,50);
```
cmagowan-hmcts, line 2: `specific time` instead of `specific date`?

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
mestebanez: nah I was told to use pagination as opposed to batch size

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
```diff
@@ -210,15 +210,33 @@ boolean existsMediaFileIn2StorageLocations(MediaEntity media,
             AND eod2.status = :status2
             AND eod.externalLocationType = :location1
             AND eod2.externalLocationType = :location2
-            AND eod.lastModifiedDateTime < :lastModifiedBefore
+            AND eod2.lastModifiedDateTime <= :lastModifiedBefore
```
mestebanez, line 213: @cmagowan-hmcts correct me if i'm wrong but my ticket seems to suggest:-

1) the media record is the record against which the date check should be done
2) the date check should include an equals

cmagowan-hmcts: i wouldn't change the logic of this method, as that will have repercussions on the methods already using it.
would you not just have location1 as `ARM` and location2 as `incoming` so that this method does not change?

1. I think it should be against how long it has been in ARM, only if it has been in ARM for 24hrs, should we then delete the incoming file. If ARM was down for a week, then the media record could have been sitting there for a week, and not migrated to ARM, so we should not be using the media record for anything.

2. i don't think this will matter...we are talking about nanosecond precision! its unlikely there will be a file that was transferred to ARM at that specific point, and if there was, we woun;t care too much, it will just get marked for deletion the next time the job runs.

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
cmagowan-hmcts: my question is why we are limiting the result at all?
who told you to use pagination?

hemantasharma1129: There is no requirement of using a batch or pagesize in the ticket.

mestebanez: in mind mind it was always good to have the flexibility of pagination. if null is passed through pagination is disabled. In that way it adds flexibility to the underlying database call and fits in with the underlying concept of the task. I will remove it.

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
cmagowan-hmcts: just thinking about this, i can see that your above change may be required

src/main/resources/db/migration/common/V1_334__add_task_inbound_transcription_annotation_deletion.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO darts.automated_task (aut_id,task_name,task_description,cron_expression,cron_editable, batch_size)
+VALUES (nextval('aut_seq'),'InboundTranscriptionAnnotationDeleter','Marks for deletion the transcription and annotation that is stored in inbound for the longer than a specific time.','0 0 22 * * *',true,50);
```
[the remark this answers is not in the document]
mestebanez: so i no longer use this value. I will change to null

hemantasharma1129 (approved):

PR #1767, opened 2024-07-29
DMP-3303 Hide isHearingActual false cases from darts portal

jackmaloney:
https://tools.hmcts.net/jira/browse/DMP-3303

### Change description ###
Hide isHearingActual false cases from darts portal in search results and on hearings on case screen

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
@@ -102,11 +102,15 @@ public class CaseServiceImpl implements CaseService {
 
         List<HearingEntity> hearingList = hearingRepository.findByCaseIds(List.of(caseId));
 
-        if (hearingList.isEmpty()) {
-            getCourtCaseById(caseId);
+        if (hearingList.isEmpty() && !caseRepository.existsById(caseId)) {
+            throw new DartsApiException(CaseApiError.CASE_NOT_FOUND);
         }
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
```diff
@@ -102,11 +102,15 @@ public List<Hearing> getCaseHearings(Integer caseId) {
 
         List<HearingEntity> hearingList = hearingRepository.findByCaseIds(List.of(caseId));
 
-        if (hearingList.isEmpty()) {
-            getCourtCaseById(caseId);
+        if (hearingList.isEmpty() && !caseRepository.existsById(caseId)) {
```
cmagowan-hmcts, line 105: Do we want this caseRepository check?
If no hearing exist, then that means that no events appeared, which means that the case has been added by an addCase request, but never happened. Do we want to return an empty array in this scenario?

cmagowan-hmcts (approved):

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
jackmaloney: As discussed, this is existing functionality implemented in DMP-753 that we don't want to change as part of this PR. That ticket specifically says to change this from an empty response to an error.

PR #1775, opened 2024-07-30
DMP-3252 Return 404 error when hearing is actual false

jackmaloney:
https://tools.hmcts.net/jira/browse/DMP-3252

### Change description ###
When a user tries to access a hearing where the hearing is actual is false, return a 404 error.

Also, added some missing unit tests covering scenarios not covered anywhere else in codebase.

```
[ ] Yes
[ x] No
```


sabahirfan (approved):
LGTM

PR #1776, opened 2024-07-30
DMP-3673: GET /userstate globalAccess fix

chrisbellingham-hmcts:
# [DMP-3673](https://tools.hmcts.net/jira/browse/DMP-3673)

This PR resolves a bug whereby the reported `globalAccess` value in the `GET /userstate` response could be incorrect if a user has multiple groups assigned with the same role, where those groups have a mixture of globalAccess values.

Additionally, the existing `AuthorisationServiceImpl` code is rewritten to remove a class of bugs relating to the assumed ordering of a user's security groups and roles, and the underlying queries reworked to obtain data in a shape more friendly to being processed into the `UserState` response. Usages are also tidied up to remove dead code.

A set of broad integration tests (at endpoint level) are also added to validate behaviour and correct integration across components, as currently there is heavy reliance on component-level tests and mocks.

Is this a breaking change?
```
[ ] Yes
[x] No
```


src/integrationTest/java/uk/gov/hmcts/darts/authentication/controller/impl/AuthenticationControllerTest.java
```diff
@@ -1,79 +0,0 @@
...
-import java.net.URI;
-
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
-import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
-
-@AutoConfigureMockMvc
-class AuthenticationControllerTest extends IntegrationBase {
```
chrisbellingham-hmcts, line 26: Author note: This test was in the wrong package (`authentication`), and has been moved (and reworked) into `authorisation` (see `src/integrationTest/java/uk/gov/hmcts/darts/authorisation/controller/AuthorisationControllerIntTest.java`)

src/integrationTest/java/uk/gov/hmcts/darts/testutils/PostgresIntegrationBase.java
```diff
@@ -50,7 +50,9 @@ static void configureProperties(DynamicPropertyRegistry registry) {
 
     @BeforeEach
     void clearDb() {
+        dartsDatabase.resetSequences();
         dartsDatabase.clearDatabaseInThisOrder();
+        dartsDatabase.resetTablesWithPredefinedTestData();
```
chrisbellingham-hmcts, line 55: Author note: Aligning the `PostgresIntegrationBase` setup logic with `IntegrationBase`

src/main/java/uk/gov/hmcts/darts/authorisation/controller/impl/AuthorisationControllerImpl.java
```diff
@@ -24,10 +20,7 @@ public class AuthorisationControllerImpl implements AuthorisationController {
     @Override
     public UserState getUserState() {
         UserAccountEntity currentUser = authorisationApi.getCurrentUser();
-        Optional<UserState> userStateOptional = authorisationApi.getAuthorisation(currentUser.getEmailAddress());
-        if (userStateOptional.isEmpty()) {
-            throw new DartsApiException(AuthorisationError.USER_NOT_AUTHORISED_FOR_ENDPOINT);
```
chrisbellingham-hmcts, line 29: Author note: If the user email provided in the JWT does not match an email in the database, then `authorisationApi.getCurrentUser()` will throw an exception, so this line can never be reached.

src/test/java/uk/gov/hmcts/darts/authorisation/controller/impl/AuthorisationControllerTest.java
```diff
@@ -54,25 +53,24 @@ void getUserStateOk() {
...
 
     @Test
     void getUserStateForbiddenWhenNoEmailAddress() {
-        UserAccountEntity userAccountEntity = CommonTestDataUtil.createUserAccount();
-        when(authorisationApi.getCurrentUser()).thenReturn(userAccountEntity);
-        when(authorisationApi.getAuthorisation(userAccountEntity.getEmailAddress())).thenReturn(Optional.empty());
+        when(authorisationApi.getCurrentUser()).thenThrow(new DartsApiException(AuthorisationError.USER_DETAILS_INVALID));
         var exception = assertThrows(DartsApiException.class, () -> controller.getUserState());
-        assertEquals(AuthorisationError.USER_NOT_AUTHORISED_FOR_ENDPOINT, exception.getError());
+        assertEquals(AuthorisationError.USER_DETAILS_INVALID, exception.getError());
```
chrisbellingham-hmcts, line 73: Author note: These mocks set up a scenario that wasn't possible, so correcting here.

src/integrationTest/java/uk/gov/hmcts/darts/authorisation/controller/AuthorisationControllerIntTest.java
```diff
@@ -0,0 +1,551 @@
...
+                                      "isActive": true
+                                    }""",
+                                mvcResult.getResponse().getContentAsString(),
+                                JSONCompareMode.NON_EXTENSIBLE
+        );
+    }
+
+    @Test
+    @WithMockSecurityContextWithEmailAddress(emailAddress = EMAIL_ADDRESS)
+    void getUserStateShouldFailWhenUserIsInactive() throws Exception {
```
chrisbellingham-hmcts, line 455: @hemantasharma1129 this is the current behaviour also. But is it correct? The `GET /userstate` response has an `isActive` field, but since the endpoint will always throw a 403 for inactive users, the `isActive` value can never be false, which makes the field essentially redundant.

hemantasharma1129: @chrisbellingham-hmcts I agree for /userstate the flag is redundant but the same service method is used in handlOauthCode method which expect the flag so don't think we have to make any changes for it.

PR #1786, opened 2024-07-31
DMP-3632 Fix index out of bounds error on empty transcription workflows

jackmaloney:
https://tools.hmcts.net/jira/browse/DMP-3632

### Change description ###
Fix for when a transcription contains no workflows and an admin tries to load the transcription. 
Looking at the code, this should never actually happen on prod as a transcription will always have a workflow. However, on the lower environments, it appears there are some test transcriptions without workflows. 

```
[ ] Yes
[ x] No
```


sabahirfan (approved):
LGTM

PR #1788, opened 2024-07-31
Possible Refactor related to DMP-3713

mestebanez:

n/a

### Change description ###

This spike has fallen out of bug https://tools.hmcts.net/jira/browse/DMP-3713. 

_**Problem**_

There are many pieces of logic within the darts codebase that have to be context aware i.e. the logic has to understand how it is being executed e.g. in ATS or default (user mode). As such, certain methods have to explicitly by pass the use of the user identity when in ATS mode. 

A better pattern would be for the implementation to always be user agnostic leaving the user identity injection up to the Di framework. This will allow the associated darts logic to be more reusable than it would otherwise be.

**_Solution_**

Introduce the concept a darts mode and load an alternative user identity if we are running in ATS mode. In this manner the underlying code that gets executed does not need to care about its execution context. This naturally promotes reuse of methods as well as reducing the scope for bugs occurring when executing those methods.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/AtsModeTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/AtsModeTest.java
@@ -1,0 +1,23 @@
+package uk.gov.hmcts.darts;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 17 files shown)


mario-paniccia:
I like this proposal 👍

mestebanez:
Closing see tag https://github.com/hmcts/darts-api/releases/tag/UserIdentityAndModes

PR #1795, opened 2024-08-01
DMP-3473 - increase advanced search performance

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-3473


src/main/java/uk/gov/hmcts/darts/audio/helper/PostAdminMediasSearchHelper.java
```diff
@@ -128,8 +128,8 @@ private List<Predicate> addDateCriteria(CriteriaBuilder criteriaBuilder, Root<Me
         return predicateList;
     }
 
-    private String surroundWithPercentagesUpper(String value) {
-        return surroundValue(value.toUpperCase(Locale.ROOT), "%");
+    private String surroundWithPercentages(String value) {
+        return surroundValue(value, "%");
```
karen-hedges, line 132: Did you mean to remove the toUpperCase?

cmagowan-hmcts: yes, we are changing the queries to an `ilike`, which is case insensitive by default, so we don't need to uppercase anything

PR #1809, opened 2024-08-05
 DMP-3535 set retention confidence attributes on case and case retent…

mario-paniccia:
…ion on case closure

https://tools.hmcts.net/jira/browse/DMP-3535

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/retention/enums/RetentionConfidenceScoreEnumConverter.java
+++ b/src/main/java/uk/gov/hmcts/darts/retention/enums/RetentionConfidenceScoreEnumConverter.java
@@ -1,0 +1,30 @@
+package uk.gov.hmcts.darts.retention.enums;
+
+import jakarta.persistence.AttributeConverter;
+import jakarta.persistence.Converter;
+
+import java.util.stream.Stream;
+
+@Converter(autoApply = true)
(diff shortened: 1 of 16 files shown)


src/main/java/uk/gov/hmcts/darts/retention/enums/RetentionConfidenceScoreEnumConverter.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.stream.Stream;
+
+@Converter(autoApply = true)
+public class RetentionConfidenceScoreEnumConverter implements AttributeConverter<RetentionConfidenceScoreEnum, Integer> {
+
+    @Override
+    public Integer convertToDatabaseColumn(RetentionConfidenceScoreEnum retentionConfidenceScoreEnum) {
+        if (retentionConfidenceScoreEnum == null) {
+            return null;
```
karen-hedges, line 14: Instead of returning null, would it make sense to use 0 as NOT_SET for instance?

mario-paniccia: I don't think setting 0 for absence of value is correct. absence of value should be represented with a `null`

chrisbellingham-hmcts (approved):

PR #1811, opened 2024-08-06
DMP-2839 Write to log - Track ARM Push-Pull flow

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-2839

### Change description ###

Added logging to when ARM push and pull is successful or when it fails

```
[ ] Yes
[x ] No
```


cmagowan-hmcts (approved):

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractUnstructuredToArmProcessor.java
```diff
@@ -135,6 +148,9 @@ protected void updateTransferAttempts(ExternalObjectDirectoryEntity externalObje
             externalObjectDirectoryEntity.getId()
         );
         externalObjectDirectoryEntity.setTransferAttempts(currentNumberOfAttempts + 1);
+        if (currentNumberOfAttempts >= armDataManagementConfiguration.getMaxRetryAttempts()) {
+            logApi.armPushFailed(externalObjectDirectoryEntity.getId());
+        }
```
chrisbellingham-hmcts, line 153: Wondering if using `currentNumberOfAttempts` in the conditional here is a bit misleading, because at this point the actual "number of attempts" is actually `currentNumberOfAttempts + 1`.

Would it be clearer to pre-increment in L150, and then use the updated value (and tweak the logic) in the check in L151? Something like:

```java
        externalObjectDirectoryEntity.setTransferAttempts(++currentNumberOfAttempts);
        if (currentNumberOfAttempts > armDataManagementConfiguration.getMaxRetryAttempts()) {
            logApi.armPushFailed(externalObjectDirectoryEntity.getId());
        }
```

PR #1812, opened 2024-08-06
DMP-3748 Populate empty fields in audio file download ReadMe file

mario-paniccia:
https://tools.hmcts.net/jira/browse/DMP-3748

### Change description ###
PR includes some cleanup

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
```diff
@@ -95,32 +79,6 @@ public List<MediaEntity> getMediaMetadata(Integer hearingId) {
             .toList();
     }
 
-    @Override
-    public Optional<UUID> getMediaLocation(MediaEntity media, Integer containerLocationId) {
```
mario-paniccia, line 99: author note:
this method was used in the past by this class but now is no longer used (only by tests). So I've removed it

PR #1825, opened 2024-08-08
Dmp 3602 adjust refresh and lifetime periods of jwks

mestebanez:

https://tools.hmcts.net/jira/browse/DMP-3602

### Change description ###

This ticket performs the following:-

* Introduce the concept of a refresh period for jwks keys together with associated tests that prove the refresh is happening
* Add 4 new vault keys to store the associated jwks refresh values
* Add tests to accurately test the token validation. In this way we ensure we fully validate the security layer

As part of the above tasks I have added support so that we :-

* Allow the Spring security layer to be tested end to end using a Spring profile - In an ideal world all tests would use this same mechanism in order to accurately test the darts logic end to end rather than using mocks
* Have the abiliy to generate a genuine RSA signed token that we can feed into the Darts api, together with the associated JWKS keys in order to successfully validate the token
* Allow Wiremock stubs to simulate the Idp calls

NOTE: There would seem to be  a bug in the Spring token validation layer, whereby we do not authorise audience claims. The code to fix this is currently commented out as I believe it would break the functional tests. This needs some discussion.

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/authentication/component/impl/TokenValidatorImpl.java
```diff
@@ -59,11 +59,15 @@ public JwtValidationResult validate(String accessToken, AuthProviderConfiguratio
...
+            log.error("JWT Token Validation failed", e);
             return new JwtValidationResult(false, e.getMessage());
         }
 
         return new JwtValidationResult(true, null);
     }
 
-}
+    protected boolean shouldIgnoreValidationException(Exception e) {
+        return false;
```
hemantasharma1129, line 70: Why its hardcoded to FALSE?

mestebanez:
This has been tagged under https://tools.hmcts.net/jira/browse/DMP-3888. A followup ticket https://tools.hmcts.net/jira/browse/DMP-3888 has been created.

PR #1826, opened 2024-08-08
DMP-3398 changed audio requests download endpoint to return "Content-length" header

mario-paniccia:
https://tools.hmcts.net/jira/browse/DMP-3398

### Change description ###
This PR is a first attempt to fix DMP-3398. As of now the endpoint does not set a "Content-length" header resulting in the server using chunked transfer encoding to transmit the data to the user.
It seems that specifying "Content-Length" is always preferable when downloading binary files like in our case compared to chunked encoding. See [here](https://stackoverflow.com/questions/2419281/content-length-header-versus-chunked-encoding) for example.
I'm not sure whether this change would improve the download speed and fix the issue, since it's not possible to reproduce this issue locally. But this seems to me a good improvement anyway.

```
[ ] Yes
[ x] No
```


sabahirfan (approved):
LGTM

PR #1848, opened 2024-08-13
DMP-3764 added index on external_object_directory manifest_file column

mario-paniccia:
https://tools.hmcts.net/jira/browse/DMP-3764

### Change description ###
By introducing this index the query execution time on Perf env goes down from 3 mins to 300ms

```
[ ] Yes
[ x] No
```


cmagowan-hmcts:
i've a funny feeling that postgres didn't use the standard index with a % at teh end, that it needed to specify the type, let me double check this by looking at the explain plan

src/main/resources/db/migration/common/V1_352__add_eod_manifest_file_index.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX IF NOT EXISTS eod_manifest_file_idx ON external_object_directory (manifest_file)
```
cmagowan-hmcts, line 1: i've a funny feeling that postgres didn't use the standard index with a % at teh end, that it needed to specify the type, let me double check this by looking at the explain plan

mario-paniccia: ok cheers. According to documentation it should use the standard index (B-tree) for LIKE 'DARTS%'. 
If that's not the case then there's something weird going on with our Postgres

cmagowan-hmcts: as discussed, we need the `text_pattern_ops` as part of the index

```sql
CREATE INDEX IF NOT EXISTS eod_manifest_file_idx ON external_object_directory (manifest_file text_pattern_ops);```

cmagowan-hmcts (changes requested):

src/main/resources/db/migration/common/V1_352__add_eod_manifest_file_index.sql
mario-paniccia: nice spot thanks

cmagowan-hmcts (approved):

sabahirfan (approved):

PR #1859, opened 2024-08-14
DMP-3765 create unique index on user_account.user_email_address

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-3765

### Change description ###

Added unique index on user_email_address where is_active is true

```
[ ] Yes
[x ] No
```


src/main/resources/db/migration/postgres/V1_357__add_unique_email_index_tables.sql
```diff
@@ -0,0 +1,8 @@
+update darts.user_account ua1 set is_active = false where ua1.usr_id not in
```
cmagowan-hmcts, line 1: this might be better:-
```sql
update darts.user_account ua1 set is_active = false where exists (select 1
from darts.user_account ua2
where ua2.user_email_address = ua1.user_email_address
and ua2.is_active = true
and ua1.is_active = true
	and ua1.created_ts < ua2.created_ts);
```

cmagowan-hmcts (approved):

PR #1861, opened 2024-08-14
DMP-3735 attempt to fix slow query by forcing a specific SQL query using IN operator

mario-paniccia:
https://tools.hmcts.net/jira/browse/DMP-3735

### Change description ###
On the Perf DB I can see the following extremely slow SQL query:
```
select eode1_0.med_id 
from darts.external_object_directory eode1_0 
where (? = case when eode1_0.med_id is not null then ? end) and eode1_0.ors_id=? and eode1_0.elt_id=?
```
I can't figure out why this SQL is being produced from [this](https://github.com/hmcts/darts-api/blob/a8e0240675574d1ed044c3d5cbc81e97868dd8ca/src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java#L73) JPQL query. Maybe because the JPQL isn't correct since it's using `eod.media.id` rather than `eod.media` ?  I was expecting a query with the IN operator:
```
select * from darts.external_object_directory
where med_id  in (?)
and ors_id=? and elt_id=?
```
Interestingly the above is being produced only when I run locally against a local Postgres, not in integration tests (both H2 and Testcontainers)
I've tested the IN operator version on Perf env and that is fast, therefore as the quickest solution I'm using a native query to force IN operator query to be used.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
@@ -62,16 +62,17 @@ public interface ExternalObjectDirectoryRepository extends JpaRepository<Externa
     List<ExternalObjectDirectoryEntity> findByTranscriptionDocumentEntityAndExternalLocationType(TranscriptionDocumentEntity transcriptionDocument,
                                                                                                  ExternalLocationTypeEntity externalLocationType);
 
-    @Query(
-        """
-                SELECT eod.media.id FROM ExternalObjectDirectoryEntity eod
-                WHERE eod.media.id in :mediaIdList
-                AND eod.status = :status
-                AND eod.externalLocationType = :externalLocationType
-            """
+    @Query(value = """
+                SELECT eod.med_id 
+                FROM darts.external_object_directory eod
+                WHERE eod.med_id in (:mediaIdList)
+                AND eod.ors_id = :status
+                AND eod.elt_id = :externalLocationType
+            """,
+        nativeQuery = true
     )
-    List<Integer> findMediaIdsByInMediaIdStatusAndType(List<Integer> mediaIdList, ObjectRecordStatusEntity status,
-                                                       ExternalLocationTypeEntity externalLocationType);
+    List<Integer> findMediaIdsByInMediaIdStatusAndType(List<Integer> mediaIdList, Integer status,
+                                                       Integer externalLocationType);
 
     @Query(
         """
--- a/src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
@@ -62,17 +62,16 @@ public interface ExternalObjectDirectoryRepository extends JpaRepository<Externa
     List<ExternalObjectDirectoryEntity> findByTranscriptionDocumentEntityAndExternalLocationType(TranscriptionDocumentEntity transcriptionDocument,
                                                                                                  ExternalLocationTypeEntity externalLocationType);
 
-    @Query(value = """
-                SELECT eod.med_id 
-                FROM darts.external_object_directory eod
-                WHERE eod.med_id in (:mediaIdList)
-                AND eod.ors_id = :status
-                AND eod.elt_id = :externalLocationType
-            """,
-        nativeQuery = true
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
```diff
@@ -62,16 +62,17 @@ List<ExternalObjectDirectoryEntity> findByMediaStatusAndType(MediaEntity media,
     List<ExternalObjectDirectoryEntity> findByTranscriptionDocumentEntityAndExternalLocationType(TranscriptionDocumentEntity transcriptionDocument,
                                                                                                  ExternalLocationTypeEntity externalLocationType);
 
-    @Query(
-        """
-                SELECT eod.media.id FROM ExternalObjectDirectoryEntity eod
-                WHERE eod.media.id in :mediaIdList
```
cmagowan-hmcts, line 68: you mentioned in the description that you don't know why eod.media.id was used here, its because mediaIdList is a list of med_ids's , not a list of mediaEntities, eod.media refers to the entity, not the id.

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
```diff
@@ -62,16 +62,17 @@ List<ExternalObjectDirectoryEntity> findByMediaStatusAndType(MediaEntity media,
...
-                AND eod.externalLocationType = :externalLocationType
-            """
+    @Query(value = """
+                SELECT eod.med_id 
+                FROM darts.external_object_directory eod
+                WHERE eod.med_id in (:mediaIdList)
+                AND eod.ors_id = :status
+                AND eod.elt_id = :externalLocationType
+            """,
+        nativeQuery = true
```
cmagowan-hmcts, line 72: @tonymort / @mario-paniccia  is there a downside to using native sql?
does it get validated at start-up etc?
Is it as maintainable?

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
mario-paniccia: yes I noticed that, but we could have generated a list of MediaEntity by using mediaRepository.getByReference()

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
mario-paniccia: we have some other native queries already, so I assumed it's accepted. 
The alternative is spending more time to figure out why that weird SQL is being generated from the JPQL. Perhaps I can try replacing `eod.media.id` with `eod.media` and passing a List<MediaEntity>. 
I didn't go with that solution because is a bigger change, and we don't have much time on these tickets

cmagowan-hmcts: i've been doing some digging, it looks like when the original query:-
```sql
SELECT eod.media.id FROM ExternalObjectDirectoryEntity eod
WHERE eod.media.id in ( :mediaIdList )
AND eod.status = :status
AND eod.externalLocationType = :externalLocationType
```

was provided with an *empty* list, it produced this SQL:-
```sql
select
        eode1_0.med_id 
    from
        darts.external_object_directory eode1_0 
    where
        (
            1 = case 
                when eode1_0.med_id is not null 
                    then 0 
            end
        ) 
        and eode1_0.ors_id=2
        and eode1_0.elt_id=1;
        ```
        
        Which seemed to be really badly performing.
        But when provided with a populated list, it produces this:-
        ```sql
    select
        eode1_0.med_id 
    from
        darts.external_object_directory eode1_0 
    where
        eode1_0.med_id in (1) 
        and eode1_0.ors_id=2
        and eode1_0.elt_id=1;
        ```
        
        which performs really well.
        
so i think it may be best to bypass calling this method if the list is empty.

mario-paniccia: nice finding @cmagowan-hmcts, what you suggested is the best solution then. I'll make the changes cheers

cmagowan-hmcts (approved):

PR #1896, opened 2024-08-19
DMP-3687 - missing db constraints

cmagowan-hmcts:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/resources/db/migration/postgres/V1_369_2__missing_db_changes.sql
+++ b/src/main/resources/db/migration/postgres/V1_369_2__missing_db_changes.sql
@@ -1,0 +1,2 @@
+DROP INDEX user_account_user_email_address_unq;
+CREATE UNIQUE INDEX user_account_user_email_address_unq ON user_account (upper(user_email_address)) where is_active;
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -6,7 +6,7 @@
   ],
   "timezone": "Europe/London",
   "schedule": [
-    "after 6pm every weekday",
-    "before 8am every weekday"
+    "after 7am and before 8am every weekday",
+    "after 6pm and before 8pm every weekday"
   ]
 }
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -6,7 +6,7 @@
   ],
   "timezone": "Europe/London",
   "schedule": [
-    "after 7am and before 8am every weekday",
+    "after 7am and before 9am every weekday",
     "after 6pm and before 8pm every weekday"
   ]
 }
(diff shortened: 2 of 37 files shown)


src/main/resources/db/migration/postgres/V1_369_2__missing_db_changes.sql
```diff
@@ -0,0 +1,2 @@
+DROP INDEX user_account_user_email_address_unq;
+CREATE UNIQUE INDEX user_account_user_email_address_unq ON user_account (upper(user_email_address)) where is_active;
```
chrisbellingham-hmcts, line 2: The create user process will not allow a new account to be created if the email address already exists in the database, regardless of `is_active` state, so should we drop the `where is_active` clause here? @hemantasharma1129, thoughts?

cmagowan-hmcts: Although the API prevents it, it doesn;t stop someone accidentally creating a 2nd user in the database, which is what happened a few weeks ago and caused some problems, but if we don't see it as an ongoing issue, then it can be dropped.

sabahirfan: The `where is_active` creates a partial index. This means the index only includes rows where the is_active column is true. It improves the performance though. It also allows for scenarios where we might want to keep inactive user records but reuse their email addresses for new accounts

This index ensures that email addresses are unique only among active users. This could be beneficial only if our application allows reuse of email addresses for inactive accounts.

If we need to ensure global uniqueness of email addresses (active and inactive), this index alone won't be sufficient. Queries that don't include the is_active condition in their WHERE clause might not be able to use this index effectively.

chrisbellingham-hmcts: Upon further discussion, there is a small risk of email duplicates arising from the migration process from legacy. On that basis, the `WHERE` clause will let us accept those duplicate addresses, whilst protecting the application from having multiple active accounts.

cmagowan-hmcts: There has been some chatter going on in the background regarding this.
Current plan is that in order to support migration, we allow multiple accounts with the same email_Address, but only 1 active.
After go-live, we will look at the data and remove any duplicate accounts then change the index to not include the is_Active flag:-
https://tools.hmcts.net/jira/browse/DMP-3819

Any time the code looks up this table, it should be using the is_Active flag AFAIK

chrisbellingham-hmcts (approved):

PR #1915, opened 2024-08-23
DMP-3821: Re-enable audio tests part 1

tonymort:
https://tools.hmcts.net/jira/browse/DMP-3821

Re-enable audio tests part 1

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioRequestsControllerAddAudioRequestDownloadIntTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioRequestsControllerAddAudioRequestDownloadIntTest.java
@@ -16,7 +16,6 @@ import uk.gov.hmcts.darts.audio.enums.MediaRequestStatus;
 import uk.gov.hmcts.darts.audiorequests.model.AudioRequestDetails;
 import uk.gov.hmcts.darts.audiorequests.model.AudioRequestType;
 import uk.gov.hmcts.darts.authorisation.component.UserIdentity;
-import uk.gov.hmcts.darts.common.entity.CourtCaseEntity;
 import uk.gov.hmcts.darts.common.entity.CourthouseEntity;
 import uk.gov.hmcts.darts.common.entity.HearingEntity;
 import uk.gov.hmcts.darts.common.entity.UserAccountEntity;
@@ -74,11 +73,11 @@ class AudioRequestsControllerAddAudioRequestDownloadIntTest extends IntegrationB
             SOME_COURTROOM,
             LocalDateTime.parse(HEARING_DATETIME)
         );
-        CourtCaseEntity courtCase = hearingEntity.getCourtCase();
-        courtCase.addProsecutor("aProsecutor");
-        courtCase.addDefendant("aDefendant");
-        courtCase.addDefence("aDefence");
-        dartsDatabase.save(courtCase);
+        //        CourtCaseEntity courtCase = hearingEntity.getCourtCase();
+        //        courtCase.addProsecutor("aProsecutor");
+        //        courtCase.addDefendant("aDefendant");
+        //        courtCase.addDefence("aDefence");
+        //        dartsDatabase.save(courtCase);
 
         testUser = dartsDatabase.getUserAccountStub()
             .createAuthorisedIntegrationTestUser(hearingEntity.getCourtroom().getCourthouse());
--- a/src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioRequestsControllerAddAudioRequestIntTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioRequestsControllerAddAudioRequestIntTest.java
@@ -16,7 +16,6 @@ import uk.gov.hmcts.darts.audio.enums.MediaRequestStatus;
 import uk.gov.hmcts.darts.audiorequests.model.AudioRequestDetails;
 import uk.gov.hmcts.darts.audiorequests.model.AudioRequestType;
 import uk.gov.hmcts.darts.authorisation.component.UserIdentity;
-import uk.gov.hmcts.darts.common.entity.CourtCaseEntity;
 import uk.gov.hmcts.darts.common.entity.CourthouseEntity;
 import uk.gov.hmcts.darts.common.entity.HearingEntity;
 import uk.gov.hmcts.darts.common.entity.UserAccountEntity;
@@ -75,11 +74,11 @@ class AudioRequestsControllerAddAudioRequestIntTest extends IntegrationBase {
             SOME_COURTROOM,
             LocalDateTime.parse(HEARING_DATETIME)
         );
-        CourtCaseEntity courtCase = hearingEntity.getCourtCase();
-        courtCase.addProsecutor("aProsecutor");
-        courtCase.addDefendant("aDefendant");
-        courtCase.addDefence("aDefence");
-        dartsDatabase.save(courtCase);
+        //        CourtCaseEntity courtCase = hearingEntity.getCourtCase();
+        //        courtCase.addProsecutor("aProsecutor");
+        //        courtCase.addDefendant("aDefendant");
+        //        courtCase.addDefence("aDefence");
+        //        dartsDatabase.save(courtCase);
 
         testUser = dartsDatabase.getUserAccountStub()
             .createAuthorisedIntegrationTestUser(hearingEntity.getCourtroom().getCourthouse());
--- a/src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioRequestsControllerAddAudioRequestPlaybackIntTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioRequestsControllerAddAudioRequestPlaybackIntTest.java
@@ -16,7 +16,6 @@ import uk.gov.hmcts.darts.audio.enums.MediaRequestStatus;
 import uk.gov.hmcts.darts.audiorequests.model.AudioRequestDetails;
 import uk.gov.hmcts.darts.audiorequests.model.AudioRequestType;
 import uk.gov.hmcts.darts.authorisation.component.UserIdentity;
-import uk.gov.hmcts.darts.common.entity.CourtCaseEntity;
 import uk.gov.hmcts.darts.common.entity.HearingEntity;
 import uk.gov.hmcts.darts.common.entity.UserAccountEntity;
 import uk.gov.hmcts.darts.notification.api.NotificationApi;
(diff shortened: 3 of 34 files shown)


src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerGetCaseByIdTest.java
```diff
@@ -53,8 +55,8 @@ void setUp() {
             DateConverterUtil.toLocalDateTime(SOME_DATE_TIME)
         );
         CourtCaseEntity courtCase = hearingEntity.getCourtCase();
-        courtCase.addProsecutor("aProsecutor");
-        courtCase.addDefendant("aDefendant");
+        courtCase.addProsecutor(someMinimalProsecutorForCase(courtCase));
```
cmagowan-hmcts, line 58: this seems quite messy, the fact that we are passing the courtcase into the method, but then add the prosecutor to the courtCase, so its been liked twice.
I thought this normally led to issues in tests.

src/testCommon/java/uk/gov/hmcts/darts/test/common/data/ProsecutorTestData.java
```diff
@@ -37,9 +41,14 @@ private static ProsecutorEntity createProsecutorWithCaseBasedName(int index, Cou
...
     public static ProsecutorEntity createProsecutorForCaseWithName(CourtCaseEntity courtCase, String name) {
-        var prosecutor = new ProsecutorEntity();
-        prosecutor.setCourtCase(courtCase);
+        var prosecutor = someMinimalProsecutorForCase(courtCase);
         prosecutor.setName(name);
         return prosecutor;
     }
+
+    public static ProsecutorEntity someMinimalProsecutorForCase(CourtCaseEntity courtCase) {
+        var prosecutor = someMinimalProsecutor();
```
cmagowan-hmcts, line 50: someMinimalProsecutor also creates its own courtcase, so we are creating multiple courtcases here, woudl something like this be better? :-

```java
public static ProsecutorEntity someMinimalProsecutor() {
    return someMinimalProsecutor(createSomeMinimalCase());
}

public static ProsecutorEntity someMinimalProsecutor(CourtCaseEntity courtCase) {
    var prosecutor = new ProsecutorEntity();
    prosecutor.setCourtCase(courtCase);
    prosecutor.setName("some-prosecutor");
    var accountEntity = minimalUserAccount();
    prosecutor.setCreatedBy(accountEntity);
    prosecutor.setLastModifiedBy(accountEntity);
    return prosecutor;
}
```

src/testCommon/java/uk/gov/hmcts/darts/test/common/data/HearingTestData.java
```diff
@@ -50,7 +50,10 @@ public static HearingEntity someMinimalHearing() {
     }
 
     public static HearingEntity createHearingWith(CourtCaseEntity courtCase, CourtroomEntity courtroom, LocalDate hearingDate) {
-        return createHearingWithDefaults(courtCase, courtroom, hearingDate, null);
+        var hearing = createHearingFor(courtCase);
```
cmagowan-hmcts, line 53: `createHearingFor` calls `someMinimalHearing()` which in turn creates its own case, courtroom etc, yet we are then instantly overriding them. Could we do something similar like what i mentioned about Prosecutor?
I notice you removed the call to `createHearingWithDefaults`, curious as to why it caused problems.

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerGetCaseByIdTest.java
tonymort: This replaces a method that was added specifically for test in production code which is one of things that needs to be removed.  What issues in test?

src/testCommon/java/uk/gov/hmcts/darts/test/common/data/ProsecutorTestData.java
tonymort: The the courtcase in from the minimal method is overwritten before being persisted when call this method.  This is the point of this pattern over the stubs which couples data generation with persistence.  minimal* is a base minimal persistence for which other methods can build upon for convenience and readability of tests.

src/testCommon/java/uk/gov/hmcts/darts/test/common/data/HearingTestData.java
tonymort: same a above

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerGetCaseByIdTest.java
cmagowan-hmcts: I can;t remember it was last year, but i think there was persistence problems because the 2 objects were being linked twice.
If someMinimalProsecutorForCase() links the Prosecutor to the case, why do we need `courtCase.addProsecutor` also? Can we call  someMinimalProsecutorForCase by itself?

src/testCommon/java/uk/gov/hmcts/darts/test/common/data/ProsecutorTestData.java
cmagowan-hmcts: is what i have written not functionally teh same, but without creating superfluous entities?

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerGetCaseByIdTest.java
tonymort: Its a bidirectional relationship so it is by nature linked both ways.  The build is passing so there doesn't seem to be any issues.

No because when we call save(case) we want the save method to iterate through its fields and save the ones that are relevant, in this case save(prosecuter)

src/testCommon/java/uk/gov/hmcts/darts/test/common/data/ProsecutorTestData.java
tonymort: The entity is overwritten immediately so isn't causing any issues.  Having 1 place that each entity is created is desirable from a maintainability perspective.  This is another issue with the integration tests, mainly the stubs having lots of places where we new objects up.  The minimal* pattern gives an entity that is gauranteed to be persistable, so all methods should build upon that.

davet1985: @tonymort @cmagowan-hmcts I'm not sure there's a huge amount to be gained by overloading these methods here. If a case object is created, linked and overwritten then it's all done in memory and during a test run.

We would like to get these tests re-enabled as soon as possible, we should continue with what Tony has and any concerns can be raised in tickets.

src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioRequestsControllerAddAudioRequestDownloadIntTest.java
```diff
@@ -74,11 +73,11 @@ void beforeEach() {
...
-        CourtCaseEntity courtCase = hearingEntity.getCourtCase();
-        courtCase.addProsecutor("aProsecutor");
-        courtCase.addDefendant("aDefendant");
-        courtCase.addDefence("aDefence");
-        dartsDatabase.save(courtCase);
+        //        CourtCaseEntity courtCase = hearingEntity.getCourtCase();
+        //        courtCase.addProsecutor("aProsecutor");
+        //        courtCase.addDefendant("aDefendant");
+        //        courtCase.addDefence("aDefence");
+        //        dartsDatabase.save(courtCase);
```
chrisbellingham-hmcts, line 80: Just checking if this commenting is intentional.

src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioRequestsControllerAddAudioRequestIntTest.java
```diff
@@ -75,11 +74,11 @@ void beforeEach() {
...
-        CourtCaseEntity courtCase = hearingEntity.getCourtCase();
-        courtCase.addProsecutor("aProsecutor");
-        courtCase.addDefendant("aDefendant");
-        courtCase.addDefence("aDefence");
-        dartsDatabase.save(courtCase);
+        //        CourtCaseEntity courtCase = hearingEntity.getCourtCase();
+        //        courtCase.addProsecutor("aProsecutor");
+        //        courtCase.addDefendant("aDefendant");
+        //        courtCase.addDefence("aDefence");
+        //        dartsDatabase.save(courtCase);
```
chrisbellingham-hmcts, line 81: Just checking if this commenting is intentional.

src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioRequestsControllerAddAudioRequestPlaybackIntTest.java
```diff
@@ -73,11 +72,11 @@ void beforeEach() {
...
-        CourtCaseEntity courtCase = hearingEntity.getCourtCase();
-        courtCase.addProsecutor("aProsecutor");
-        courtCase.addDefendant("aDefendant");
-        courtCase.addDefence("aDefence");
-        dartsDatabase.save(courtCase);
+        //        CourtCaseEntity courtCase = hearingEntity.getCourtCase();
+        //        courtCase.addProsecutor("aProsecutor");
+        //        courtCase.addDefendant("aDefendant");
+        //        courtCase.addDefence("aDefence");
+        //        dartsDatabase.save(courtCase);
```
chrisbellingham-hmcts, line 79: Just checking if this commenting is intentional.

src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioRequestsControllerAddAudioRequestDownloadIntTest.java
tonymort: Yes, they were using a method that was only being used in test which has been deleted.  These tests are fixed in part 2 of re-enabling the audio tests

src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioRequestsControllerAddAudioRequestIntTest.java
tonymort: same

src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioRequestsControllerAddAudioRequestPlaybackIntTest.java
tonymort: same

PR #1916, opened 2024-08-23
 NO_JIRA: log Darts feign clients requests/responses full details on Demo

mario-paniccia:
### Change description ###
During QA it it sometimes important to see the full details of http requests and responses Darts is making to external services, e.g. ARM. 
This PR configures Feign clients to log headers, body, and metadata for both requests and responses, and replaces an existing config that is not working.
For now we will enable full logging only on Demo:
https://github.com/hmcts/sds-flux-config/pull/5343

Note:
we have [this ticket](https://tools.hmcts.net/jira/browse/DMP-3854) to remove this setting before we go to Prod to avoid sensitive data leak in logs:

as part of that ticket we can consider having different configs depending on the environments, so that full logging is enabled on Staging (which would help QA) but disabled on Prod

Example:
[
Screenshot 2024-08-23 at 09 52 51
Screenshot 2024-08-23 at 10 25 41
](url)

```
[ ] Yes
[x ] No
```

### DIFF
--- a/charts/darts-api/Chart.yaml
+++ b/charts/darts-api/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "1.0"
 description: A Helm chart for darts-api App
 name: darts-api
 home: https://github.com/hmcts/darts-api
-version: 0.0.84
+version: 0.0.85
 maintainers:
   - name: HMCTS darts team
 dependencies:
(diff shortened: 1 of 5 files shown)


hemantasharma1129 (approved):

sabahirfan (approved):
LGTM

mestebanez (approved):

PR #1930, opened 2024-08-29
DMP-3323 refactor retrieve core object service

sabahirfan:
DMP-3323 

### Change description ###
This refactored structure addresses the following issues:

- Removes the need for @SuppressWarnings("PMD.TooManyMethods") by splitting the functionality into smaller, focused services.

- Eliminates self-invocation of @Transactional methods within the same class.

- Improves separation of concerns by creating dedicated services for each entity type.

- Maintains the transactional and retryable behavior for the appropriate methods.

- Keeps the RetrieveCoreObjectServiceImpl as a facade that coordinates between the different services.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/common/service/impl/CaseCommonServiceImpl.java
```diff
@@ -0,0 +1,58 @@
...
+
+@RequiredArgsConstructor
+@Service
+public class CaseCommonServiceImpl implements CaseCommonService {
+
+    private final CaseRepository caseRepository;
+    private final CourthouseCommonService courthouseCommonService;
+
+    @Override
+    @Transactional
```
cmagowan-hmcts, line 23: If we are already inside a Transaction, one that was started outside of this class, will it create a new one, or continue to use the encapsulating one?
i.e. do we need propagation = Propagation.REQUIRES_NEW ?

src/main/java/uk/gov/hmcts/darts/common/service/impl/HearingCommonServiceImpl.java
```diff
@@ -0,0 +1,71 @@
...
+                                                 UserAccountEntity userAccount) {
+        Optional<HearingEntity> foundHearing = hearingRepository.findHearing(courthouseName, courtroomName, caseNumber, hearingDate.toLocalDate());
+
+        return foundHearing.map(hearingEntity -> setHearingLastDateModifiedBy(hearingEntity, userAccount))
+            .orElseGet(() -> createHearing(courthouseName, courtroomName, caseNumber, hearingDate, userAccount, null));
+    }
+
+    @Override
+    @Transactional
+    public HearingEntity retrieveOrCreateHearingWithMedia(String courthouseName, String courtroomName, String caseNumber, LocalDateTime hearingDate,
```
cmagowan-hmcts, line 39: i was confused by this method, and so looked into its usages and realised ti seems to be some helper method for tests. it probably shouldn't be in here.
Can we add a comment in the code for it to be removed at a later date?

src/main/java/uk/gov/hmcts/darts/common/service/impl/JudgeCommonServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+    @Override
+    @Transactional
+    public JudgeEntity retrieveOrCreateJudge(String judgeName, UserAccountEntity userAccount) {
+        Optional<JudgeEntity> foundJudge = judgeRepository.findByNameIgnoreCase(judgeName);
+        return foundJudge.orElseGet(() -> createJudge(judgeName, userAccount));
+    }
+
+    private JudgeEntity createJudge(String judgeName, UserAccountEntity userAccount) {
+        JudgeEntity judge = new JudgeEntity();
+        String upperCaseJudgeName = judgeName != null ? judgeName.toUpperCase(Locale.ROOT) : null;
```
cmagowan-hmcts, line 30: If judgeName is null, i don't think we should be adding it at all!
Perhaps a null check before line 24?

src/main/java/uk/gov/hmcts/darts/common/service/impl/CaseCommonServiceImpl.java
sabahirfan: No we don't need propagation = Propagation.REQUIRES_NEW

src/main/java/uk/gov/hmcts/darts/common/service/impl/CourthouseCommonServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Optional;
+
+@RequiredArgsConstructor
+@Service
+public class CourthouseCommonServiceImpl implements CourthouseCommonService {
+
+    private final CourthouseRepository courthouseRepository;
+
+    @Override
+    @Transactional(readOnly = true)
```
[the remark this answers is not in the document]
sabahirfan: I believe so.

src/main/java/uk/gov/hmcts/darts/common/service/impl/HearingCommonServiceImpl.java
sabahirfan: sure added @Deprecated with detail

src/main/java/uk/gov/hmcts/darts/common/service/impl/JudgeCommonServiceImpl.java
sabahirfan: done

src/main/java/uk/gov/hmcts/darts/common/service/impl/JudgeCommonServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+@RequiredArgsConstructor
+@Service
+public class JudgeCommonServiceImpl implements JudgeCommonService {
+
+    private final JudgeRepository judgeRepository;
+
+    @Override
+    @Transactional
+    public JudgeEntity retrieveOrCreateJudge(String judgeName, UserAccountEntity userAccount) {
+        if (judgeName == null || judgeName.trim().isEmpty()) {
```
cmagowan-hmcts, line 24: For future reference,this could have also been achieved using apache commons StringUtils.isBlank

PR #1947, opened 2024-09-03
Dmp 3825 tune automated tasks

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-3825

### Change description ###

Updated the automated tasks cron expressions so that they are less likely to interfere with one another and added/changed the batch size

```
[ ] Yes
[x ] No
```


sabahirfan (approved):
LGTM

src/main/resources/db/migration/common/V1_375__update_automated_tasks_default_values.sql
```diff
@@ -0,0 +1,75 @@
+-- Run every hour at 10 mins past the hour between 8pm and 8am with batch size 5,000
```
davet1985, line 1: @cmagowan-hmcts @jackmaloney this adds another flyway script, it doesn't change the table definitions, but not sure if the migration team need to be aware?

jackmaloney: Yeh, they're already aware.

PR #1949, opened 2024-09-03
DMP-3892 - Ensure event cleanup current flag job only cleans events linked to same hearing

Ben-Edwards-cgi:
### Change description ###
When the event cleanup current flag [job|https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=1689786831#TaskautomationLLD-Cleanupis_currentflagforevents] runs, it is currently setting the flag to false for any event that has the same event_id as another event (except 0). 

Currently, there is an issue on lower environments where events are linked to multiple hearings and this job cleans up all but one of the events so these hearings appear to have no events associated to them in the portal. 

This needs to be tweaked slightly so that it only sets the flag to false for events that have the same event_id and the same associated hearing entities. 

Example of a case with this issue on staging [https://darts.staging.apps.hmcts.net/case/173845/hearing/182750?tab=Transcripts] Note there are no events showing for this case.

```
[ ] Yes
[X] No
```


src/integrationTest/java/uk/gov/hmcts/darts/common/repository/EventRepositoryTest.java
```diff
@@ -34,23 +33,46 @@ void testEventProcessing() {
...
+        Assertions.assertTrue(eventIdMap.containsKey(eventIdsToBeProcessed2.getFirst()));
         Assertions.assertNotEquals(eventIdsToBeProcessed1, eventIdsToBeProcessed2);
 
         // ENSURE WE DONT PROCESS THE THIRD BATCH I.E. THE ZERO EVENT ID
         List<Integer> eventIdsToBeProcessed3 = eventRepository.getCurrentEventIdsToBeProcessed(Pageable.ofSize(1));
         Assertions.assertTrue(eventIdsToBeProcessed3.isEmpty());
     }
+
+    @Test
+    void testEventProcessingWithMultipleHearings() {
```
jackmaloney, line 57: Any chance you can make it a bit clearer what this test is actually trying to achieve? It's not clear to me what this is trying to do.
This example [PR](https://github.com/mario-paniccia/darts-api-fork/pull/1/files) gives examples of the sort of format we want to use (although it's not been implemented throughout the project, it would be good if we can try and align to this as much as possible).

src/main/java/uk/gov/hmcts/darts/event/service/impl/CleanupCurrentFlagEventProcessorImpl.java
```diff
@@ -23,27 +23,25 @@ public List<Integer> processCurrentEvent() {
...
 
         eventEntityReturned.forEach(event -> {
-            Integer eventIdPrimaryKey = eventRepository.getTheLatestCreatedEventPrimaryKeyForTheEventId(event);
-            log.debug("Current event primary key is {}", eventIdPrimaryKey);
-
-            eventPrimaryKeysLst.add(eventIdPrimaryKey);
-            eventIdsLst.add(event);
-
+            EventRepository.EventIdAndHearingIds eventIdAndHearingIds = eventRepository.getTheLatestCreatedEventPrimaryKeyForTheEventId(event);
+            log.debug("Current event primary key is {}", eventIdAndHearingIds);
```
jackmaloney, line 36: Is this going to print the whole object? Or do you want to print just one of the primary keys?

src/integrationTest/java/uk/gov/hmcts/darts/common/repository/EventRepositoryTest.java
Ben-Edwards-cgi: Updated :)

src/main/java/uk/gov/hmcts/darts/event/service/impl/CleanupCurrentFlagEventProcessorImpl.java
Ben-Edwards-cgi: I have updated the log after this one to include more detail and removed this log as they both showed the same information

src/main/java/uk/gov/hmcts/darts/event/service/impl/CleanupCurrentFlagEventProcessorImpl.java
```diff
@@ -14,6 +14,7 @@
 @Getter
 @RequiredArgsConstructor
 @Slf4j
+//TODO add integration tests for this class
```
jackmaloney, line 17: Remove Todo

jackmaloney (approved):

src/main/java/uk/gov/hmcts/darts/event/service/impl/CleanupCurrentFlagEventProcessorImpl.java
Ben-Edwards-cgi: Removed :)

PR #1983, opened 2024-09-11
Updated getCurrentEventIdsToBeProcessed() to exclude duplicate event_ids with unique hearing ids

Ben-Edwards-cgi:
https://tools.hmcts.net/jira/browse/DMP-3892

### Change description ###
Updated getCurrentEventIdsToBeProcessed() to exclude processed event_ids that have multiple current eve_ids but are associated to different hearings

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/repository/EventRepository.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/repository/EventRepository.java
@@ -84,13 +84,13 @@ public interface EventRepository extends JpaRepository<EventEntity, Integer> {
     @Query(value = """
         SELECT distinct e2.event_id
         from (
-        	SELECT e.event_id, he.eve_id, string_agg(he.hea_id::varchar, ',' order by he.hea_id) as hearing_ids FROM darts.event e
-            LEFT JOIN darts.hearing_event_ae he
-            ON he.eve_id = e.eve_id
-            where e.is_current = true
-        	AND e.event_id <> 0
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/darts/common/repository/EventRepository.java
```diff
@@ -82,15 +82,21 @@ Page<EventSearchResult> searchEventsFilteringOn(
...
+          AND e.event_id <> 0
+          AND e.event_id IS NOT null
+          GROUP by he.eve_id, e.event_id
+        ) e2
+        GROUP BY e2.event_id, e2.hearing_ids
+        having count(e2.event_id) > 1
+        limit = :limit
         """, nativeQuery = true)
-    List<Integer> getCurrentEventIdsToBeProcessed(Pageable pageable);
+    List<Integer> getCurrentEventIdsToBeProcessed(long limit);
```
karen-hedges, line 99: Currently the rest of the application uses pageable instead of limit. Is there a reason you are detracting from this pattern?

Ben-Edwards-cgi: Yeah, pageable is for when we want to grab keep our place in the search, or use multiple pages of items.

Here we just process the first X amount and we never care about what is on the other pages, so limit is more performant as we don't need to count all the items, work out the number of pages, deal with offsets, etc

karen-hedges (approved):

PR #1985, opened 2024-09-11
DMP-3876 Inbound Store Deleter Job - Replace ARM RPO with Unstructured criteria

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-3876

### Change description ###

Updated the inbound store data deleter tasks to check against unstructured instead or ARM

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/darts/datamanagement/service/impl/InboundAnnotationTranscriptionDeleterProcessorImpl.java
```diff
@@ -61,7 +61,7 @@ public List<Integer> markForDeletion(int hourBeforeCurrentDate) {
             currentTimeHelper.currentOffsetDateTime()
         );
 
-        log.debug("Records have been marked for deletion");
+        recordsMarkedForDeletion.stream().forEach(eodId -> log.info("Set status of unstructured EOD {} to be marked for deletion", eodId));
```
Ben-Edwards-cgi, line 64: Will be worth wrapping this in a if(log.isInfoEnabled() {...}

karen-hedges: As far as I know log level will never be less than info so its not really required unless we want to reduce the strings on the stack on start up. I am outputting individual eod id's as its easier for support to read them rather than a long string seperated by commas.

src/integrationTest/resources/application-intTest.yaml
```diff
@@ -3,6 +3,8 @@ spring:
     retention-period:
       inbound:
         arm-minimum: 24
+        unstructured-minimum:
+          hours: 24
```
davet1985, line 7: This is a different format to above, it would be good to keep this consistent.

src/main/resources/application.yaml
```diff
@@ -181,6 +181,8 @@ darts:
     retention-period:
       inbound:
         arm-minimum: 24
+        unstructured-minimum:
+          hours: 24
```
davet1985, line 185: Same comment as above regarding the format here.

davet1985 (commented):
Just one minor comment.

src/main/resources/application.yaml
karen-hedges: Its consistent with the naming convention for the one below, lines 186-188, which I think is a better naming convention. If you they should all be consistent which way is best to go for?

src/main/java/uk/gov/hmcts/darts/datamanagement/service/impl/InboundAnnotationTranscriptionDeleterProcessorImpl.java
Ben-Edwards-cgi: I'm happy with the above :)

hemantasharma1129 (approved):

PR #1988, opened 2024-09-12
DMP-3933 - (Dev Only) -> Refactor automated task service to use spring dependency injection

Ben-Edwards-cgi:
### Change description ###
Refactor automated task service to use spring dependency injection

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/task/service/impl/AutomatedTaskServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/task/service/impl/AutomatedTaskServiceImpl.java
@@ -132,6 +132,7 @@ public class AutomatedTaskServiceImpl implements AutomatedTaskService {
         }
         boolean found = false;
         for (AutoloadingAutomatedTask autoloadingAutomatedTask : autoloadingAutomatedTasks) {
+            System.out.println("TMP: " + autoloadingAutomatedTask.getAutomatedTaskName() + " -- " + automatedTaskName);
             if (automatedTaskName.equals(autoloadingAutomatedTask.getAutomatedTaskName())) {
                 found = true;
                 TriggerAndAutomatedTask triggerAndAutomatedTask = getTriggerAndAutomatedTask(autoloadingAutomatedTask.getTaskName());
@@ -145,7 +146,7 @@ public class AutomatedTaskServiceImpl implements AutomatedTaskService {
                 break;
             }
         }
-        if (found) {
+        if (!found) {
             throw new DartsApiException(FAILED_TO_FIND_AUTOMATED_TASK);
         }
     }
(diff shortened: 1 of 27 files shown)


src/main/resources/application.yaml
```diff
@@ -37,7 +37,7 @@ spring:
             response-type: code
             response-mode: form_post
             prompt: login
-            issuer-uri: ${ACTIVE_DIRECTORY_B2C_BASE_URI}/${AAD_B2C_TENANT_ID:00000000-0000-0000-0000-000000000000}/v2.0/
+            issuer-uri: ${ACTIVE_DIRECTORY_B2C_BASE_URI:https://hmctsstgextid.b2clogin.com}/${AAD_B2C_TENANT_ID:00000000-0000-0000-0000-000000000000}/v2.0/
```
karen-hedges, line 40: Is this default url mean't to be exposed?

Ben-Edwards-cgi: This is already in the readme document I have defaulted here for simplicity

src/main/resources/application.yaml
```diff
@@ -166,7 +166,7 @@ darts:
     temp-blob-workspace: ${user.home}/audiotransform/tempworkspace
     trim-workspace: ${user.home}/audiotransform/trim
   azure:
-    active-directory-b2c-base-uri: ${ACTIVE_DIRECTORY_B2C_BASE_URI}
+    active-directory-b2c-base-uri: ${ACTIVE_DIRECTORY_B2C_BASE_URI:https://hmctsstgextid.b2clogin.com}
```
karen-hedges, line 169: @hemantasharma1129  Is this default url mean't to be exposed?

src/main/java/uk/gov/hmcts/darts/task/service/impl/AutomatedTaskServiceImpl.java
```diff
@@ -132,6 +132,7 @@ public void reloadTaskByName(String taskName) {
         }
         boolean found = false;
         for (AutoloadingAutomatedTask autoloadingAutomatedTask : autoloadingAutomatedTasks) {
+            System.out.println("TMP: " + autoloadingAutomatedTask.getAutomatedTaskName() + " -- " + automatedTaskName);
```
karen-hedges, line 135: I take it you are using system.out instead of logging here because logging is not yet setup? if so, might be worth adding a comment so someone later doesn't come along and think its a mistake and change it to logging

Ben-Edwards-cgi: I have removed this as this was left over from some dev testing :)

karen-hedges (approved):

PR #1996, opened 2024-09-13
DMP-2543 - Automated task to remove Personally Identifiable Information from case on expiry

Ben-Edwards-cgi:
### Change description ###
A record of the case's existence must still be available in DARTS following the retain until date being reached. 

No personally Identifiable Information data on citizens can be visible once the retain date has been reached for the case but the case must still be returned in search results.

An automated task should be written that removes the following Personally Identifiable Information from the case:
 * defendant name(s)
 * defence name(s)
 * prosecutor name(s)
 * event text
 * transcription comment

These fields should be replaced with UUIDs to denote text used to exist but has since been removed. 

The following fields should be set on the court_case record when this job runs:
 * is_data_anonymised = true
 * data_anonymised_by = automated task user
 * data_anonymised_ts = current timestamp

*Task configuration and feature flag*

For the automated task configuration in the database ensure the following are set:
 * task_name = "CaseExpiryDeletion"
 * task_enabled = false
 * cron_expression = "0 0 9 30 2 *" (At 09:00 AM, on day 30 of the month, only in February)
 * batch_size = 1000

Ensure this automated task is controlled through a feature flag environment variable via Flux that is disabled on production, read in via spring application.yaml, as below.
{code:java}
automated:
  task:
    expiry-deletion:
      enabled: ${CASE_EXPIRY_DELETION_ENABLED:false}{code}
In addition to setting task_enabled to false, use this flag to condition the logic of the task so that it will not run unless it's also specifically enabled by this feature flag. This is to prevent the task from being run manually by accident.

```
[ ] Yes
[X] No
```


sabahirfan (approved):

sabahirfan:
Please check the failing checks otherwise LGTM.

PR #1999, opened 2024-09-13
DMP-3023 - Close active transcript workflow when case expires

Ben-Edwards-cgi:
### Change description ###
If a transcript request is mid workflow when a case expires then the workflow will automatically be moved to closed.

```
[ ] Yes
[X] No
```


sabahirfan (approved):
LGTM

PR #2000, opened 2024-09-13
DMP-2559 - Write to log when a case is deleted due to expiry for Dynatrace purposes

Ben-Edwards-cgi:
### Change description ###
h3. Background:

The client has requested that we have a Dynatrace dashboard tile that proves that the Deletion Service is working. Outputs should show the number of cases expired in the selected timeframe.
h3. Technical:

When a case has been expired successfully, write the following log statement at INFO level:
{noformat}
"Case expired: cas_id={cas_id}, case_number={case_number}"{noformat}
Dynamic data:
||field||value||
|cas_id|Case id for the case that has been deleted |
|case_number|Case number for the case that has been deleted|

Please add the log entry in a service under the [log package|https://github.com/hmcts/darts-api/tree/master/src/main/java/uk/gov/hmcts/darts/log], exposing the service method via the "api" sub-package. Add unit tests to cover the log entry scenarios.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/entity/CourtCaseEntity.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/entity/CourtCaseEntity.java
@@ -18,6 +18,7 @@ import jakarta.persistence.SequenceGenerator;
 import jakarta.persistence.Table;
 import lombok.Getter;
 import lombok.Setter;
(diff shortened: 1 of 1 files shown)


sabahirfan (approved):
LGTM

PR #2001, opened 2024-09-13
DMP-2544 - An audit log must be kept for case deletion due to expiry

Ben-Edwards-cgi:
### Change description ###
An audit log must be kept of what has been deleted. Items can either be deleted due to retain until date passing.

The audit trail must include, case ID and the date the records were deleted.

In order to facilitate this for expiry deletion, the following should be implemented
 add a new audit activity, named CASE_EXPIRED
 When the automated task defined by DMP-2543 removes data due to the case being expired, create an audit entry using the CASE_EXPIRED activity and the case_id

```
[ ] Yes
[X] No
```


sabahirfan (approved):
LGTM

PR #2003, opened 2024-09-16
DMP-3923 - Hide CaseExpiryDeletion task if disabled by feature flag

Ben-Edwards-cgi:
### Change description ###
The expiry deletion automated task, named "CaseExpiryDeletion" and added by DMP-2543, is configured within the database automated_tasks table. In addition, there is a feature flag within application.yaml driven by an env var set in flux, named CASE_EXPIRY_DELETION_ENABLED

If the task is not enabled then we need to prevent it from being the subject of the following automated tasks endpoints, as follows:

- GET /admin/automated-tasks - do not include the "CaseExpiryDeletion" task in the response
- GET /admin/automated-tasks/\{task_id} - return 404 (AUTOMATED_TASK_NOT_FOUND)
- PATCH /admin/automated-tasks/\{task_id} - return 404 (AUTOMATED_TASK_NOT_FOUND)
- POST /admin/automated-tasks/\{task_id}/run - return 404 (AUTOMATED_TASK_NOT_FOUND)

This will prevent it from showing on the admin portal and prevent consuming the API endpoints directly to amend or run the task.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/repository/CaseRepository.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/repository/CaseRepository.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.darts.common.repository;
 
+import org.springframework.data.domain.Limit;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.Query;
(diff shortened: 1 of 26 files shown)


sabahirfan (approved):
LGTM

src/main/java/uk/gov/hmcts/darts/common/repository/CaseRepository.java
```diff
@@ -83,13 +84,11 @@ List<CourtCaseEntity> findCasesNeedingCaseDocumentForRetentionDateGeneration(Off
...
-        """,
-        nativeQuery = true)
-    List<CourtCaseEntity> findCasesToBeAnonymized(OffsetDateTime maxRetentionDate, long limit);
+        select cc from CourtCaseEntity cc
+        join CaseRetentionEntity cr
+        on cr.courtCase.id = cc.id and cr.currentState = 'COMPLETE'
+        where cc.isDataAnonymised = false
+        and cr.retainUntil < :maxRetentionDate
+        """)
+    List<CourtCaseEntity> findCasesToBeAnonymized(OffsetDateTime maxRetentionDate, Limit limit);
```
sabahirfan, line 93: `Limit` is not used anywhere in the current code effectively and should be replaced with Pageable for limit and pagination functionality

Ben-Edwards-cgi: We use limit instead of Pageable here because it is more performant as Pageable gets the total counts of entries that match the criteria where Limit only grabs the limit so less SQL

sabahirfan (approved):

PR #2007, opened 2024-09-16
DMP-3645 - Prevent expired case URLs from being accessed

Ben-Edwards-cgi:
### Change description ###
When DMP-2909 is implemented, only basic information about a case will be displayed to the user. DARTS should prevent users from manually trying to change URLs to access the tabs that have been hidden.

When a user tries to access hidden parts of the case with the *is_data_anonymised* flag set to true, return a 404 error with the following detail:
 * code = CASE_XXX (replacing XXX with the next number in the sequence)
 * title = Case has expired

This needs to be applied to all DARTS portal endpoints that can be called from the frontend for a case e.g. Hearings, Transcripts, Annotations, Events, Audios etc.

The admin endpoints should remain accessible. 

The frontend part of this ticket can be found in DMP-2989. 

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
@@ -129,8 +129,7 @@ public class CaseServiceImpl implements CaseService {
     @Override
     public CourtCaseEntity getCourtCaseById(Integer caseId) {
         return caseRepository.findById(caseId)
-            .orElseThrow(() -> new DartsApiException(CaseApiError.CASE_NOT_FOUND))
-            .validateIsExpired();
+            .orElseThrow(() -> new DartsApiException(CaseApiError.CASE_NOT_FOUND));
     }
 
     private void createCourtroomIfMissing(List<HearingEntity> hearings, GetCasesRequest request, UserAccountEntity userAccount) {
@@ -199,7 +198,7 @@ public class CaseServiceImpl implements CaseService {
 
     @Override
     public List<Annotation> getAnnotations(Integer caseId) {
-        CourtCaseEntity courtCaseEntity = getCourtCaseById(caseId);
+        CourtCaseEntity courtCaseEntity = getCourtCaseById(caseId).validateIsExpired();
         List<HearingEntity> hearingEntities = courtCaseEntity.getHearings();
 
         if (authorisationApi.userHasOneOfRoles(List.of(SecurityRoleEnum.SUPER_ADMIN))) {
--- a/src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
@@ -145,7 +145,7 @@ public class CaseServiceImpl implements CaseService {
         CourtCaseEntity courtCase = retrieveCoreObjectService.retrieveOrCreateCase(
             addCaseRequest.getCourthouse(),
             addCaseRequest.getCaseNumber()
-        ).validateIsExpired();
+        );
         return updateCase(addCaseRequest, courtCase);
     }
 
@@ -153,7 +153,7 @@ public class CaseServiceImpl implements CaseService {
         CourtCaseEntity updatedCaseEntity = casesMapper.addDefendantProsecutorDefenderJudgeType(
             existingCase,
             addCaseRequest
-        ).validateIsExpired();
+        );
         caseRepository.saveAndFlush(updatedCaseEntity);
         return casesMapper.mapToPostCaseResponse(updatedCaseEntity);
     }
--- a/src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
@@ -90,7 +90,6 @@ public class CaseServiceImpl implements CaseService {
             request.getCourtroom(),
             request.getDate()
         );
-
         UserAccountEntity currentUser = authorisationApi.getCurrentUser();
         createCourtroomIfMissing(hearings, request, currentUser);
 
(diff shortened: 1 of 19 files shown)


src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
```diff
@@ -143,15 +145,15 @@ public PostCaseResponse addCaseOrUpdate(AddCaseRequest addCaseRequest) {
         CourtCaseEntity courtCase = retrieveCoreObjectService.retrieveOrCreateCase(
             addCaseRequest.getCourthouse(),
             addCaseRequest.getCaseNumber()
-        );
+        ).validateIsExpired();
```
davet1985, line 148: I'm not sure about adding this here because this is used by the "addCase" endpoint and we don't want to expose this case expired error to the external services which call that endpoint. 

I think we should ensure this validation is only enforced for endpoints which are used by the portal.

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
```diff
@@ -143,15 +145,15 @@ public PostCaseResponse addCaseOrUpdate(AddCaseRequest addCaseRequest) {
...
         return updateCase(addCaseRequest, courtCase);
     }
 
     private PostCaseResponse updateCase(AddCaseRequest addCaseRequest, CourtCaseEntity existingCase) {
         CourtCaseEntity updatedCaseEntity = casesMapper.addDefendantProsecutorDefenderJudgeType(
             existingCase,
             addCaseRequest
-        );
+        ).validateIsExpired();
```
davet1985, line 156: Same as above.

src/main/java/uk/gov/hmcts/darts/common/service/impl/CaseCommonServiceImpl.java
```diff
@@ -25,16 +25,20 @@ public class CaseCommonServiceImpl implements CaseCommonService {
     public CourtCaseEntity retrieveOrCreateCase(String courthouseName, String caseNumber, UserAccountEntity userAccount) {
         String courthouseNameUC = StringUtils.toRootUpperCase(courthouseName);
         Optional<CourtCaseEntity> foundCase = caseRepository.findByCaseNumberAndCourthouse_CourthouseName(caseNumber, courthouseNameUC);
-        return foundCase.map(entity -> setCourtCaseLastDateModifiedBy(entity, userAccount))
-            .orElseGet(() -> createCase(courthouseName, caseNumber, userAccount));
+        return foundCase
+            .map(entity -> setCourtCaseLastDateModifiedBy(entity, userAccount))
+            .orElseGet(() -> createCase(courthouseName, caseNumber, userAccount))
+            .validateIsExpired();
```
davet1985, line 31: As per other comment, we should keep this validation out of common code that fetches the case.

src/main/java/uk/gov/hmcts/darts/common/service/impl/CaseCommonServiceImpl.java
```diff
@@ -25,16 +25,20 @@ public class CaseCommonServiceImpl implements CaseCommonService {
...
     @Override
     @Transactional
     public CourtCaseEntity retrieveOrCreateCase(CourthouseEntity courthouse, String caseNumber, UserAccountEntity userAccount) {
         Optional<CourtCaseEntity> foundCase = caseRepository.findByCaseNumberAndCourthouse(caseNumber, courthouse);
-        return foundCase.map(entity -> setCourtCaseLastDateModifiedBy(entity, userAccount))
-            .orElseGet(() -> createCase(courthouse, caseNumber, userAccount));
+        return foundCase
+            .map(entity -> setCourtCaseLastDateModifiedBy(entity, userAccount))
+            .orElseGet(() -> createCase(courthouse, caseNumber, userAccount))
+            .validateIsExpired();
```
davet1985, line 41: Same as above.

src/main/java/uk/gov/hmcts/darts/task/service/impl/AutomatedTaskServiceImpl.java
```diff
@@ -56,8 +57,15 @@ public class AutomatedTaskServiceImpl implements AutomatedTaskService {
...
 
+    @Value("${darts.automated.allow-automated-tasks:true}")
+    private final boolean allowAutomatedTasks;
+
     @Override
     public void configureAndLoadAutomatedTasks(ScheduledTaskRegistrar taskRegistrar) {
+        if (!allowAutomatedTasks) {
+            log.info("Automated tasks are disabled");
+            return;
+        }
```
davet1985, line 68: Is this change intentional? It doesn't seem relevant for the ticket.

src/main/resources/application.yaml
```diff
@@ -83,7 +83,7 @@ spring:
       maximum-pool-size: ${DARTS_API_DB_POOL_SIZE:50}
   jpa:
     database: postgresql
-    show-sql: false
+    show-sql: true
```
davet1985, line 86: Not sure we want to turn this on everywhere? Maybe set in a local profile?

davet1985 (changes requested):
I'm concerned about adding this validation in some very common methods called by code that we don't want to expose the error too.

src/main/resources/application.yaml
Ben-Edwards-cgi: Agreed this was added temporarily and should have been removed - Updated

src/main/java/uk/gov/hmcts/darts/task/service/impl/AutomatedTaskServiceImpl.java
Ben-Edwards-cgi: I spoke to Jack on this one and he is happy as it makes local testing easier if these are disabled (It will be enabled on all deployed envs by default)

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
Ben-Edwards-cgi: Updated

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
Ben-Edwards-cgi: Updated

src/main/java/uk/gov/hmcts/darts/common/service/impl/CaseCommonServiceImpl.java
Ben-Edwards-cgi: Updated

src/main/java/uk/gov/hmcts/darts/common/service/impl/CaseCommonServiceImpl.java
Ben-Edwards-cgi: Updated

davet1985 (approved):

PR #2025, opened 2024-09-19
DMP-3947 - add new archive retention policies

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-3947

### DIFF
--- a/src/main/resources/db/migration/common/V1_380__added_additional_retention_policy_type.sql
+++ b/src/main/resources/db/migration/common/V1_380__added_additional_retention_policy_type.sql
@@ -1,5 +1,5 @@
-delete from retention_policy_type where fixed_policy_key in (-3,-4);
+delete from retention_policy_type where fixed_policy_key in ('-3','-4');
 
-INSERT INTO darts.retention_policy_type (rpt_id,policy_name,display_name,fixed_policy_key,duration,policy_start_ts,policy_end_ts,retention_policy_object_id,created_ts,created_by,last_modified_ts,last_modified_by,description) VALUES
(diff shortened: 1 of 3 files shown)


src/main/resources/db/migration/common/V1_380__added_additional_retention_policy_type.sql
```diff
@@ -0,0 +1,5 @@
+delete from retention_policy_type where fixed_policy_key in ('-3','-4');
+
+INSERT INTO retention_policy_type (rpt_id,policy_name,display_name,fixed_policy_key,duration,policy_start_ts,policy_end_ts,retention_policy_object_id,created_ts,created_by,last_modified_ts,last_modified_by,description) VALUES
+(10,'DARTS Archive Permanent Retention','Archive Permanent','-3','30Y0M0D','2024-01-01T00:00:00Z',null,null,current_timestamp,0,current_timestamp,0,'DARTS Archive Permanent Retention'),
```
jackmaloney, line 4: How is this different to what Ben did here? https://github.com/hmcts/darts-api/pull/2016/files

cmagowan-hmcts: they are onyl 10 and 11 on in-meory h2 database, e.g. on staging they have been inserted as 17726 and 17727, so their ids are different in every environemnt, which migration dont want.

image
****

jackmaloney, line 4: Do we not want to remove those ones put in the last PR, otherwise there will be 2 records with the same fixed_policy_key?

cmagowan-hmcts: 1st line does that

jackmaloney (approved):

PR #2029, opened 2024-09-19
DMP-3959 - Set is_anonymised on events and transcription comments when case expiry job runs

Ben-Edwards-cgi:
### Change description ###
When the case expiry job kicks in, the is_anonymised flags should be set on the events and transcription comments linked to a case. 

```
[ ] Yes
[X] No
```


sabahirfan (approved):
LGTM

PR #2033, opened 2024-09-20
 DMP-3911 Update GET /admin/events/{id} to include anonymisation details

mario-paniccia:
https://tools.hmcts.net/jira/browse/DMP-3911

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/audio/service/impl/MediaRequestServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/service/impl/MediaRequestServiceImpl.java
@@ -158,13 +158,6 @@ public class MediaRequestServiceImpl implements MediaRequestService {
         return mediaRequestRepository.saveAndFlush(mediaRequestEntity);
     }
 
-    @Transactional
-    @Override
-    public MediaRequestEntity updateAudioRequestStatus(MediaRequestEntity mediaRequestEntity, MediaRequestStatus status) {
-        mediaRequestEntity.setStatus(status);
-        return mediaRequestRepository.saveAndFlush(mediaRequestEntity);
-    }
-
     @Override
     public boolean isUserDuplicateAudioRequest(AudioRequestDetails audioRequestDetails) {
 
--- a/src/main/java/uk/gov/hmcts/darts/common/entity/EventEntity.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/entity/EventEntity.java
@@ -87,4 +87,9 @@ public class EventEntity extends CreatedModifiedBaseEntity {
     public void addHearing(HearingEntity hearingEntity) {
         hearingEntities.add(hearingEntity);
     }
+
+    public HearingEntity associatedHearing() {
+        // Events can only be associated with one hearing at this stage
+        return this.hearingEntities.getFirst();
+    }
 }
--- a/src/test/java/uk/gov/hmcts/darts/event/mapper/EventMapperTest.java
+++ b/src/test/java/uk/gov/hmcts/darts/event/mapper/EventMapperTest.java
@@ -5,10 +5,12 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.junit.jupiter.MockitoExtension;
+import uk.gov.hmcts.darts.common.entity.CourtCaseEntity;
 import uk.gov.hmcts.darts.common.entity.CourthouseEntity;
 import uk.gov.hmcts.darts.common.entity.CourtroomEntity;
 import uk.gov.hmcts.darts.common.entity.EventEntity;
 import uk.gov.hmcts.darts.common.entity.EventHandlerEntity;
+import uk.gov.hmcts.darts.common.entity.HearingEntity;
 import uk.gov.hmcts.darts.common.entity.UserAccountEntity;
 import uk.gov.hmcts.darts.event.model.AdminGetEventForIdResponseResult;
 
@@ -39,6 +41,7 @@ class EventMapperTest {
         eventHandlerEntity.setId(300);
         eventEntity.setEventType(eventHandlerEntity);
         eventEntity.setLogEntry(false);
+        eventEntity.setDataAnonymised(true);
 
         CourtroomEntity courtroom = new CourtroomEntity();
         courtroom.setId(500);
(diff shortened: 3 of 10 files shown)


src/main/java/uk/gov/hmcts/darts/audio/service/impl/MediaRequestServiceImpl.java
```diff
@@ -158,13 +158,6 @@ public MediaRequestEntity updateAudioRequestStatus(Integer id, MediaRequestStatu
         return mediaRequestRepository.saveAndFlush(mediaRequestEntity);
     }
 
-    @Transactional
-    @Override
-    public MediaRequestEntity updateAudioRequestStatus(MediaRequestEntity mediaRequestEntity, MediaRequestStatus status) {
```
mario-paniccia, line 163: author note:
this method is no longer used in production code so I'm removing it

src/main/java/uk/gov/hmcts/darts/common/entity/EventEntity.java
```diff
@@ -87,4 +87,9 @@ public class EventEntity extends CreatedModifiedBaseEntity {
     public void addHearing(HearingEntity hearingEntity) {
         hearingEntities.add(hearingEntity);
     }
-}
+
+    public HearingEntity associatedHearing() {
```
mario-paniccia, line 91: author note:
I'm not using the standard `get*` notation because I don't want a `hearing` property to be added to the entity (and potentially show up in endpoints JSON responses). This is more an utility method

Ben-Edwards-cgi: I think we can still use the get* convention here as we should never return Spring Entity as responses anyway.
We could even add a @JsonIgnore annotation to ensure it does not get added to Json output

src/main/java/uk/gov/hmcts/darts/common/entity/EventEntity.java
```diff
@@ -87,4 +87,9 @@ public class EventEntity extends CreatedModifiedBaseEntity {
...
+        // Events can only be associated with one hearing at this stage
+        return this.hearingEntities.getFirst();
```
Ben-Edwards-cgi, line 93: I think we should add some logic here something like if(this.getHearingEntities().size() > 1) {
 throw new UnsupportedOperation(...);
}

or something just so if we do associate multiple hearings in the future this scenario becomes obvious

Ben-Edwards-cgi (changes requested):

src/main/java/uk/gov/hmcts/darts/common/entity/EventEntity.java
Ben-Edwards-cgi: The JsonIgnore annotation would prevent this from showing up in a Json Response, and as we are getting the hearing entity I still think the standard naming convention of getHearing() makes sense here?

src/main/java/uk/gov/hmcts/darts/common/entity/EventEntity.java
Ben-Edwards-cgi: We probably need a wider discussion on this as because the data model supports Many-To-Many relationships we should account for these and throw relevant errors where we are getting multiple items returned where we are only expecting a single item.

This just means if we ever get bad data on the system or we add support for multiple hearing / event associations we don't have code that hides failures

mario-paniccia: Happy to throw an Exception if this is ok with Tech Leads. My understanding is that we wanted to handle this gracefully, also looking at other places. @hemantasharma1129 what do you think?

Ben-Edwards-cgi: For additional context we do also have logic to chuck an exception in this instance in other places>
So the question is do we want to be altered when data gets into an invalid state or do we want to gracefully consume it

image

src/test/java/uk/gov/hmcts/darts/event/mapper/EventMapperTest.java
```diff
@@ -84,6 +93,7 @@ void mapsEventToAdminSearchEventResponseResult() {
         Assertions.assertEquals(eventEntity.getCreatedBy().getId(), responseResult.getCreatedBy());
         Assertions.assertEquals(eventEntity.getLastModifiedDateTime(), responseResult.getLastModifiedAt());
         Assertions.assertEquals(eventEntity.getLastModifiedBy().getId(), responseResult.getLastModifiedBy());
-        Assertions.assertEquals(eventEntity.isDataAnonymised(), responseResult.getIsDataAnonymised());
+        Assertions.assertEquals(true, responseResult.getIsDataAnonymised());
```
mestebanez, line 96: Picky but assertTrue ;)

src/main/java/uk/gov/hmcts/darts/common/entity/EventEntity.java
mestebanez: Not really bothered either way but not sure what we mean by _potentially show up in endpoints JSON_? We have mappers to address this? Entity content shouldnt just be automatically added to the json response?

src/main/java/uk/gov/hmcts/darts/common/entity/EventEntity.java
mestebanez: I assume the list will never be empty. In such cases a [NoSuchElementException](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/NoSuchElementException.html) would be thrown

mestebanez (commented):
Maybe worth added an additional test to src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerTest.java in order to ensure that both SUPER_ADMIN and SUPER_USER are the only roles that can validly make this call 

NOTE: The AC for the ticket only calls out SUPER_ADMIN. I assume that this is an omission in the original ticket?

src/main/java/uk/gov/hmcts/darts/common/entity/EventEntity.java
mestebanez: We could protect against all the aforementioned scenarios by adding a check to the EventIdValidator. Better to gracefully handle these issues ourselves rather than throwing exceptions ourselves that will result in a 5xx response.

mestebanez: Also maybe if this exception was to be thrown it should be a darts exception so that the client can understand the nature of the problem

Ben-Edwards-cgi: This exception is already in the application :) But agreed it should be a darts exception.
Reason for chucking the exception instead of just getting the first is to protect the data.

But I think this needs to be redesign following some internal chats we had as, we need to handle many to many relationships between hearing and events now as there are 2,000,000 records in production with this relation

PR #2039, opened 2024-09-23
DMP-3966 - allow ATS to go into a loop if multiple requests to process.

cmagowan-hmcts:
https://tools.hmcts.net/jira/browse/DMP-3966

Added a list to ensure that if a media request that has audio only in inbound can be ignored and not hold up the processing of other requests.
added a check to make sure it doesn't loop forever.
Added a time check to try to avoid a scenario where the pod has been running for 30m and gets killed by keda, and leaves a media_request stuck in "PROCESSING".

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
@@ -56,6 +56,7 @@ import static uk.gov.hmcts.darts.audiorequests.model.AudioRequestType.DOWNLOAD;
 @Slf4j
 @SuppressWarnings({"PMD.ExceptionAsFlowControl"})
 public class AudioTransformationServiceImpl implements AudioTransformationService {
+    public static final int MAX_LOOPS = 200;//Arbitrary high number, just a glass ceiling which we wouldn't expect the loop to reach unless there is a problem.
     private final MediaRequestService mediaRequestService;
     private final OutboundFileProcessor outboundFileProcessor;
     private final OutboundFileZipGenerator outboundFileZipGenerator;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
```diff
@@ -87,12 +94,24 @@ public Path saveBlobDataToTempWorkspace(InputStream mediaFile, String fileName)
...
     @Override
     public void handleKedaInvocationForMediaRequests() {
-        Optional<MediaRequestEntity> mediaRequest = mediaRequestService.retrieveMediaRequestForProcessing();
-
-        mediaRequest.ifPresentOrElse(
-            this::processAudioRequest,
-            () -> log.info("No open requests found for ATS to process.")
-        );
+        int counter = 1;
+        List<Integer> mediaRequestIdsProcessed = new ArrayList<>();
```
mario-paniccia, line 98: in your PR description you've explained that this list list is introduced to skip picking up media requests whose files are not yet ready (in outbound store). 
can we try explaining this in the code a bit better since it's not easy to understand otherwise. Even a comment would be fine

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
```diff
@@ -87,12 +94,24 @@ public Path saveBlobDataToTempWorkspace(InputStream mediaFile, String fileName)
...
+        OffsetDateTime cutoffTime = currentTimeHelper.currentOffsetDateTime().plusMinutes(config.getLoopCutoffMinutes());
+        while (currentTimeHelper.currentOffsetDateTime().isBefore(cutoffTime)) {
+            if (counter++ > 200) {
```
mario-paniccia, line 101: can we remove this magic number and introduce a descriptive constant?

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
cmagowan-hmcts: code comment added now

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
cmagowan-hmcts: done

PR #2044, opened 2024-09-24
DMP-3636 - Automated task to remove all unstructured data at the end of their retention period from all data stores

Ben-Edwards-cgi:
### Change description ###
Unstructured data records must only be stored for as long as their retention period.

Once the retention period has passed then these records must be deleted in their entirety.

During deletion, all records will be removed from DARTS data stores including case document, transcripts, annotations and audio. 

Whenever a case is expired, all associated objects should be cleaned up or deleted as described below. Therefore, merge this logic/criteria with the existing case expiry job implemented in DMP-2543.

Data stores to delete from include: 
 * Inbound data store
 ** Hard Delete from Inbound Data Store
 ** Hard delete from external_object_directory table for elt_id=1.
 ** Soft delete from metadata e.g media by updating the is_deleted flag to true.
 * Unstructured data store
 ** Hard Delete from Unstructured Data Store
 ** Hard delete from external_object_directory table for elt_id=2.
 ** Soft delete from metadata e.g media by updating the is_deleted flag to true.
 * Outbound data store - any super admin, super user or user in media_in_perpetuity group data doesn't get deleted but with this job, on expiry the data will be deleted. 
 ** Hard Delete from Outbound Data Store
 ** Hard delete from transient_object_directory table.
 ** Hard delete from transformed_media table
 ** Soft delete from media_request -> Set status to be expired
 ** Can be done apart of CaseExpiryDeletionAutomatedTask flow instead of this new automated task
 * ARM
 ** delete from external_object_directory table for elt_id=3.

For inbound and unstructured, make sure there is an entry for ARM (elt_id = 3) in the external_object_directory table with a stored status (osr_id = 2) before the data should be deleted. 

 
Ensure this automated task is controlled through a feature flag via Flux that is disabled on production (should be covered in DMP-2543 but confirm). 

h4.User Story
AS A

DARTS System

I WANT TO 

delete all unstructured data from DARTS data stores at the end of the record's retention period

SO THAT

data is not retained any longer than it needs to be

 

```
[ ] Yes
[X] No
```


src/main/resources/db/migration/common/V1_381__UnstructuredDataExpiryDeletionAutomated_task.sql
```diff
@@ -0,0 +1,6 @@
+INSERT INTO darts.automated_task (aut_id,task_name,task_description,cron_expression,cron_editable, batch_size,
+created_ts, created_by,last_modified_ts,last_modified_by)
+VALUES (nextval('aut_seq'),'UnstructuredDataExpiryDeletion',
+ 'Deletes INBOUND, UNSTRUCTURED and ARM data that has reached its expiry date.',
```
cmagowan-hmcts, line 4: if its called `UnstructuredDataExpiryDeletion` it implies it only deleted from unstructured, not the other 2

src/main/java/uk/gov/hmcts/darts/task/runner/impl/UnstructuredDataExpiryDeletionAutomatedTask.java
```diff
@@ -0,0 +1,167 @@
...
+        this.unstructuredDeleter = unstructuredDeleter;
+    }
+
+    @Override
+    public AutomatedTaskName getAutomatedTaskName() {
+        return AutomatedTaskName.UNSTRUCTURED_DATA_EXPIRY_DELETION_TASK_NAME;
+    }
+
+    @Override
+    @Transactional
```
cmagowan-hmcts, line 82: risky puting this Transactional annotation here, is it needed?
when we hit perf/prod this job could take a while every time its run and lead to table locking and performance issues.

src/main/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImpl.java
```diff
@@ -97,8 +122,51 @@ public UserAccountEntity getUserAccount() {
...
 
+    void tidyUpTransformedMediaEntities(UserAccountEntity userAccount, CourtCaseEntity courtCase) {
+        courtCase
+            .getHearings()
+            .stream()
+            .map(HearingEntity::getMediaRequests)
+            .flatMap(Collection::stream)
+            .distinct()
+            .peek(mediaRequestEntity -> expiredMediaRequest(userAccount, mediaRequestEntity))
+            .map(MediaRequestEntity::getTransformedMediaEntities)
```
cmagowan-hmcts, line 133: not sure adding getTransformedMediaEntities to MediaRequestEntity jsut for this is justified, might be better to just call repository method here.

src/main/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImpl.java
```diff
@@ -97,8 +122,51 @@ public UserAccountEntity getUserAccount() {
...
+            .flatMap(Collection::stream)
+            .distinct()
+            .forEach(this::deleteTransformedMediaEntity)
```
cmagowan-hmcts, line 136: this stream is quite long, consider breaking up into logical chunks so that debugging and fault diagnosis is easier.

src/main/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImpl.java
```diff
@@ -97,8 +122,51 @@ public UserAccountEntity getUserAccount() {
...
+        ;
+    }
+
+    void deleteTransformedMediaEntity(TransformedMediaEntity transformedMediaEntity) {
+        if (transformedMediaEntity.getTransientObjectDirectoryEntities().stream()
+            .peek(transientObjectDirectoryEntity -> transientObjectDirectoryEntity.setTransformedMedia(null))
+            .filter(transientObjectDirectoryEntity -> !deleteTransientObjectDirectoryEntity(transientObjectDirectoryEntity))
+            .count() == 0) {
```
cmagowan-hmcts, line 144: noneMatch might be better than `.count() == 0`?

Ben-Edwards-cgi: I don't think this works in this scenario as noneMatch and allMatch will stop processing once there criteria has been invalidated which means it will could fail to delete some entries

src/main/java/uk/gov/hmcts/darts/task/runner/impl/UnstructuredDataExpiryDeletionAutomatedTask.java
Ben-Edwards-cgi: Agreed removed :)

src/main/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImpl.java
cmagowan-hmcts: yeah, thats fine.

src/main/java/uk/gov/hmcts/darts/audio/deleter/impl/ExternalDataStoreDeleterImpl.java
```diff
@@ -28,32 +28,38 @@ public List<T> delete() {
...
+            deleter.delete(externalLocation);
+            deleted = true;
+        } catch (AzureDeleteBlobException e) {
+            log.error(
+                "Failed to delete storage data with externalLocation={} for entityId={} and statusId={}",
+                externalLocation, entityId, statusId, e
+            );
+        }
+        repository.delete(entityToBeDeleted);
+        if (entityToBeDeleted instanceof TransientObjectDirectoryEntity transientObjectDirectoryEntity
```
mestebanez, line 57: Be better to override and add this to ExternalOutboundDataStoreDeleter

src/main/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImpl.java
Ben-Edwards-cgi: As the items are Lazy loaded it only gets loaded when we call getTransformedMediaEntities() so we should not have any performance impacts. 

This also helps link the objects conceptually in code simplifying logic and helping spring understand the relationship (useful when using things like QueryDSL)

src/main/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImpl.java
Ben-Edwards-cgi: Agreed updated

src/main/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImpl.java
Ben-Edwards-cgi: Agreed Updated

src/main/java/uk/gov/hmcts/darts/audio/deleter/impl/ExternalDataStoreDeleterImpl.java
Ben-Edwards-cgi: This logic existed previously :)

I just extracted it into a delete method so we can delete singular entries

src/main/java/uk/gov/hmcts/darts/task/runner/impl/CaseExpiryDeletionAutomatedTask.java
```diff
@@ -54,15 +50,10 @@ public AutomatedTaskName getAutomatedTaskName() {
     @Override
     @Transactional
     public void runTask() {
-        final List<CourtCaseEntity> courtCaseEntities = new ArrayList<>();
-
         caseRepository.findCasesToBeAnonymized(currentTimeHelper.currentOffsetDateTime(), Limit.of(getAutomatedTaskBatchSize()))
```
mestebanez, line 53: What does this task do now?

src/main/java/uk/gov/hmcts/darts/task/runner/impl/UnstructuredDataExpiryDeletionAutomatedTask.java
```diff
@@ -0,0 +1,165 @@
...
+                .stream()
+                .map(entityMapper)
+                .peek(entity -> entity.markAsDeleted(userAccount))
+                .toList()
+        );
+        externalObjectDirectoryRepository.deleteAll(deletedEntities);
+    }
+
+    boolean deleteFromExternalDataStore(ExternalObjectDirectoryEntity externalObjectDirectoryEntity) {
+        if (ExternalLocationTypeEnum.INBOUND.getId().equals(externalObjectDirectoryEntity.getExternalLocationType().getId())) {
```
mestebanez, line 153: Do we need outbound call here?

mestebanez: Also before we call the delete on the inbound and unstructured the ticket mentions:-

For inbound and unstructured, make sure there is an entry for ARM (elt_id = 3) in the external_object_directory table with a stored status (osr_id = 2)

src/main/java/uk/gov/hmcts/darts/task/runner/impl/UnstructuredDataExpiryDeletionAutomatedTask.java
```diff
@@ -0,0 +1,165 @@
...
+            deletedEntities
+                .stream()
+                .map(entityMapper)
+                .peek(entity -> entity.markAsDeleted(userAccount))
+                .toList()
+        );
+        externalObjectDirectoryRepository.deleteAll(deletedEntities);
+    }
+
+    boolean deleteFromExternalDataStore(ExternalObjectDirectoryEntity externalObjectDirectoryEntity) {
```
mestebanez, line 152: Ideally the deleters ExternalDataStoreDeleter would know what ExternalLocationTypeEnum applies to them to avoid this if statement

src/main/java/uk/gov/hmcts/darts/task/runner/SoftDelete.java
```diff
@@ -0,0 +1,27 @@
+package uk.gov.hmcts.darts.task.runner;
+
+import uk.gov.hmcts.darts.common.entity.UserAccountEntity;
+
+import java.time.OffsetDateTime;
+
+public interface SoftDelete {
```
mestebanez, line 7: Java doc here would be good, highlighting how this interface is envisaged to be used i.e. in the context of an Entity.

src/main/java/uk/gov/hmcts/darts/task/runner/impl/UnstructuredDataExpiryDeletionAutomatedTask.java
```diff
@@ -0,0 +1,165 @@
...
+        List<ExternalObjectDirectoryEntity> externalObjectDirectoryEntities,
+        Function<ExternalObjectDirectoryEntity, T> entityMapper) {
+
+
+        List<ExternalObjectDirectoryEntity> deletedEntities = externalObjectDirectoryEntities
+            .stream()
+            .filter(this::deleteFromExternalDataStore)
+            .toList();
+
+        repository.saveAll(
```
mestebanez, line 142: Optional but we could introduce a soft repository interface that soft deletes all soft entities and saves them

src/integrationTest/java/uk/gov/hmcts/darts/task/runner/impl/UnstructuredDataExpiryDeletionAutomatedTaskITest.java
```diff
@@ -0,0 +1,598 @@
+package uk.gov.hmcts.darts.task.runner.impl;
```
mestebanez, line 1: maybe by design but should we have ITest in the name. Should it be Test

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
```diff
@@ -385,4 +386,68 @@ List<ExternalObjectDirectoryEntity> findByExternalLocationTypeAndUpdateRetention
...
+        """)
+    List<ExternalObjectDirectoryEntity> findExpiredTranscriptionDocuments(OffsetDateTime maxRetentionDate, Limit batchSize);
+
+    @Query("""
+               SELECT eod
+               FROM ExternalObjectDirectoryEntity eod
+               JOIN AnnotationDocumentEntity a ON eod.annotationDocumentEntity = a
+               WHERE a.retainUntilTs < :maxRetentionDate
+               AND a.isDeleted = false
+               AND eod.externalLocationType.id in (1,2,3)
```
mestebanez, line 412: We could accept the entity types see findMatchingExternalObjectDirectoryEntityByLocation

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
```diff
@@ -385,4 +386,68 @@ List<ExternalObjectDirectoryEntity> findByExternalLocationTypeAndUpdateRetention
...
+               AND EXISTS (
+                SELECT 1 FROM ExternalObjectDirectoryEntity eod2
+                WHERE eod2.externalLocationType.id = 3
```
mestebanez, line 415: same

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
```diff
@@ -385,4 +386,68 @@ List<ExternalObjectDirectoryEntity> findByExternalLocationTypeAndUpdateRetention
...
+                AND eod2.status.id = 2
```
mestebanez, line 416: same

src/main/java/uk/gov/hmcts/darts/task/runner/impl/CaseExpiryDeletionAutomatedTask.java
Ben-Edwards-cgi: This find's cases to be anonymised and anonymises them

src/main/java/uk/gov/hmcts/darts/task/runner/impl/UnstructuredDataExpiryDeletionAutomatedTask.java
Ben-Edwards-cgi: This logic is done in the select statement  to locate data to be deleted

src/main/java/uk/gov/hmcts/darts/task/runner/impl/UnstructuredDataExpiryDeletionAutomatedTask.java
Ben-Edwards-cgi: This method is used to delete and confirm if deletion was successful so we can filter out unsuccessful deletions :)

The deletion classes will delete anything you provided to it without checking the LocationType the different deletion classes just select the datastore to delete from, as such this is required to ensure we delete from the correct location and only deletes the LocationTypes we want :)

src/integrationTest/java/uk/gov/hmcts/darts/task/runner/impl/UnstructuredDataExpiryDeletionAutomatedTaskITest.java
Ben-Edwards-cgi: This is to differentiate the integration tests from the unit tests

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
Ben-Edwards-cgi: This is a performance thing, in this instance we don't care about the all the data within the LocationTypeEnity so we don't need to fetch the object

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
Ben-Edwards-cgi: See previous

src/main/resources/db/migration/common/V1_381__UnstructuredDataExpiryDeletionAutomated_task.sql
hemantasharma1129: Agree with @cmagowan-hmcts. Unstructured is associated with one of the Data store in DARTS system. Since these all are associated objects may be "AssociatedObjectDataExpiryDeletionAutomatedTask" sounds better?

src/main/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImpl.java
cmagowan-hmcts: well it means you are having to wrap methods in transactional annotations to allow the lazy loading, which isn't great.
e.g. i can see you are adding another one here:-
https://github.com/hmcts/darts-api/pull/2044/files#diff-38e2fe2df168b40916a7e1fc72cc6313bbb56d64952a201c1ba12c27f223a7a0R42

src/main/java/uk/gov/hmcts/darts/task/runner/impl/UnstructuredDataExpiryDeletionAutomatedTask.java
Ben-Edwards-cgi: Agreed updated :)

src/main/resources/db/migration/common/V1_381__UnstructuredDataExpiryDeletionAutomated_task.sql
Ben-Edwards-cgi: Agreed updated :)

src/main/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImpl.java
Ben-Edwards-cgi: We only need to wrap the method if we want to make use of the Lazy loaded item :)

For that second transactional we want this one as we do not want to partially anonymise a case as this could lead to data not getting anonymised in the future or data sticking around in the blob stores

cmagowan-hmcts (approved):

src/main/java/uk/gov/hmcts/darts/audio/deleter/impl/ExternalDataStoreDeleterImpl.java
mestebanez: ok just always cautious about instanceof. Theres generally a better place for this logic to sit. This logic does belong in the ExternalOutboundDataStoreDeleter though? It seems the only place where this condition would be true?

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
mestebanez: ok fine with that maybe in future we could pass though ExternalLocationTypeEnum

src/main/java/uk/gov/hmcts/darts/task/runner/impl/CaseExpiryDeletionAutomatedTask.java
mestebanez: So what saves the entity?

src/main/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImpl.java
```diff
@@ -68,22 +95,22 @@ public void anonymizeHearingEntity(UserAccountEntity userAccount, HearingEntity
     @Override
     public void anonymizeEventEntity(UserAccountEntity userAccount, EventEntity eventEntity) {
```
mestebanez, line 96: Although not related to this PR be good to make these private or default scoped. Could lead to bugs if the caller assumes that the db persisting saving will be done for them. This is not the case

src/main/java/uk/gov/hmcts/darts/task/runner/impl/CaseExpiryDeletionAutomatedTask.java
mestebanez: oh ok didnt see the database save in the anonymiser class

src/main/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImpl.java
Ben-Edwards-cgi: I'm happy to default scope these for now can always pull them back up downstream if they are needed.
Updated

src/main/java/uk/gov/hmcts/darts/audio/deleter/impl/ExternalDataStoreDeleterImpl.java
Ben-Edwards-cgi: I would agree it would be better to have this in the more specific super classes so classes that dont needed it dont have it, but that is outside the scope of this ticket as this is existing logic :)

src/integrationTest/java/uk/gov/hmcts/darts/task/runner/impl/UnstructuredDataExpiryDeletionAutomatedTaskITest.java
mestebanez: isnt the fact its under integrationTest good enough? The only reason I mention this is that nothing else does this

PR #2047, opened 2024-09-24
DMP-3892 - Fixed issue where distinct would ignore order by

Ben-Edwards-cgi:
### Change description ###
Fixed issue where distinct would ignore order by

```
[ ] Yes
[X] No
```


jackmaloney:
Could an integration test be added here to prove that the scenario seen where the wrong event is set as 'is_current' is now fixed given this change?

src/main/java/uk/gov/hmcts/darts/common/repository/EventRepository.java
```diff
@@ -103,13 +103,12 @@ having count(e2.event_id) > 1
 
     @Query(value = """
         select distinct on (event_id, hearing_ids) e.* from (
-            SELECT e.eve_id, event_id, string_agg(he.hea_id::varchar, ',' order by he.hea_id) as hearing_ids FROM darts.event e
+            SELECT e.eve_id, event_id, e.created_ts, string_agg(he.hea_id::varchar, ',' order by he.hea_id) as hearing_ids FROM darts.event e
```
cmagowan-hmcts, line 106: is it possible to add a test for the scenario covered in the ticket?

Ben-Edwards-cgi: Added :)

cmagowan-hmcts (approved):

PR #2051, opened 2024-09-25
DMP-3974 - Update admin event payloads

Ben-Edwards-cgi:
### Change description ###
The timestamps used on the admin portal event screens are not correct. 

The event_ts timestamp should be used in the payload and not the created_at fields currently used. 

Update the *POST /admin/events/search* endpoint and the *GET /admin/events/\{id}* endpoints to include this, returned as 'event_ts'.

Additionally, for the GET /admin/events/\{id} endpoint:
 * is_current boolean field should be added to the payload.

And for the POST /admin/events/search endpoint:
 * the antecedent_id and chronicle_id can be removed from the admin event search endpoint as they are not needed.

Liaise with frontend as this will be a breaking change, covered in DMP-3975.

```
[ ] Yes
[X] No
```


cmagowan-hmcts (approved):

src/main/java/uk/gov/hmcts/darts/common/repository/EventRepository.java
```diff
@@ -53,6 +53,7 @@ SELECT EXISTS (
          SELECT new uk.gov.hmcts.darts.event.model.EventSearchResult(
             e.id,
             e.createdDateTime,
```
jackmaloney, line 55: Would there be another PR to remove createdDateTime, ChronicleID and Antecedent ID from this payload?
You've removed it from the payload for the event endpoint further down, so wondered why that's been included in this PR and not the future one?

src/main/resources/openapi/event.yaml
```diff
@@ -713,14 +713,13 @@ components:
         created_at:
```
jackmaloney, line 713: Remove created_at as well?

src/main/java/uk/gov/hmcts/darts/common/repository/EventRepository.java
jackmaloney, line 55: You've removed ChronicleID and Antecedent ID from the openapi code further down, so wondered why that's been removed here as well? Or will that be in a future PR?

Ben-Edwards-cgi: I have updated the code to also remove this

src/main/resources/openapi/event.yaml
Ben-Edwards-cgi: Removed :)

sabahirfan (approved):

src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventSearchControllerTest.java
```diff
@@ -103,6 +103,7 @@ void returnsAllFieldsCorrectly() throws Exception {
         var response = json.from(mvcResult.getResponse().getContentAsString());
         assertThat(response).hasJsonPathNumberValue("[0].id");
         assertThat(response).hasJsonPathStringValue("[0].created_at");
```
jackmaloney, line 105: You've removed created_at from this endpoint so think should be removed from this test.

PR #1559, opened 2024-06-11
 New automated task scaffolding setup - NO MERGE

mario-paniccia:
This PR illustrates all the scaffolding changes required to setup a new automated task, using as a reference the `GenerateCaseDocumentAutomated` task.
These changes serve as a starting point. This single commit can be cherry-picked and the various classes renamed to the desired task name

PR #2052, opened 2024-09-25
DMP-3479 - (Dev Only) Create DETSToARM Pull Response automated job framework

Ben-Edwards-cgi:
### Change description ###
We have the requirement to replicate the existing Unstructured to ARM flow for DETS to ARM as well. New jobs will be created as part of this requirement which can use any of the existing logic available in DARTS to ARM.
 * Create flyway script for new automated task
 ** Task name :ProcessDETSToArmResponseAutomatedTask
 ** Batchsize : 100000
 ** Cron : '0 49 * * * *'
 ** Rest of the details will be same as ProcessArmResponseFilesAutomatedTask
 * Create ProcessDETSToArmResponseAutomatedTask with empty process method. DETSToARM will always be driven by Batchsize so no Single implementation is required.

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/darts/task/runner/impl/ProcessDetsToArmResponseAutomatedTask.java
```diff
@@ -0,0 +1,38 @@
+package uk.gov.hmcts.darts.task.runner.impl;
```
mario-paniccia, line 1: We typically add some tests for a new automated task, you can see an example if this (old, outdated now :( reference PR we have: https://github.com/hmcts/darts-api/pull/1559/files)

src/main/java/uk/gov/hmcts/darts/task/runner/impl/ProcessDetsToArmResponseAutomatedTask.java
```diff
@@ -0,0 +1,38 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.darts.arm.component.AutomatedTaskProcessorFactory;
+import uk.gov.hmcts.darts.common.repository.AutomatedTaskRepository;
+import uk.gov.hmcts.darts.log.api.LogApi;
+import uk.gov.hmcts.darts.task.api.AutomatedTaskName;
+import uk.gov.hmcts.darts.task.config.AutomatedTaskConfigurationProperties;
+import uk.gov.hmcts.darts.task.runner.AutoloadingManualTask;
```
mario-paniccia, line 11: I'd suggest we also add a Processor class with an empty "execute" method so we can tackle more of the scaffolding stuff as part of this ticket, which is the whole point we usually raise a dedicated scaffolding ticket for automated tasks.
We have a PR that we usually use as reference here: https://github.com/hmcts/darts-api/pull/1559/files
Sadly it's a bit out of date now but still can give you an idea

src/main/java/uk/gov/hmcts/darts/task/runner/impl/ProcessDetsToArmResponseAutomatedTask.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Autowired
+    public ProcessDetsToArmResponseAutomatedTask(AutomatedTaskRepository automatedTaskRepository,
+                                                 AutomatedTaskConfigurationProperties automatedTaskConfigurationProperties,
+                                                 AutomatedTaskProcessorFactory automatedTaskProcessorFactory,
+                                                 LogApi logApi, LockService lockService) {
+        super(automatedTaskRepository, automatedTaskConfigurationProperties, logApi, lockService);
+    }
+
+    @Override
```
mario-paniccia, line 29: I think you also need these changes (see reference PR)

image

src/main/java/uk/gov/hmcts/darts/task/runner/impl/ProcessDetsToArmResponseAutomatedTask.java
```diff
@@ -0,0 +1,38 @@
...
+    public AutomatedTaskName getAutomatedTaskName() {
+        return PROCESS_DETS_TO_ARM_RESPONSE;
+    }
+
+    @Override
```
mario-paniccia, line 34: to be able to run the task manually we also need to update `ManualTaskService`

src/main/java/uk/gov/hmcts/darts/task/runner/impl/ProcessDetsToArmResponseAutomatedTask.java
Ben-Edwards-cgi: I did some work last week to tidy this logic up to reduce boiler plate so it gets automatically created now :)

src/main/java/uk/gov/hmcts/darts/task/runner/impl/ProcessDetsToArmResponseAutomatedTask.java
Ben-Edwards-cgi: I think we are moving away from the processor approach as it does not make much sense abstracting this logic out again

src/main/java/uk/gov/hmcts/darts/task/runner/impl/ProcessDetsToArmResponseAutomatedTask.java
Ben-Edwards-cgi: As we dynamically load these tasks now instead of creating the wrapper code manually, I don't think there are any tests required :)

mario-paniccia: pardon me, I was not aware that my reference PR is now obsolete after the simplifications we've done to the framework. You can ignore my comments then

mario-paniccia (approved):

PR #2054, opened 2024-09-25
DMP-1714 - Return migrated transcription comments in legacy_comments field for get transcription payload

Ben-Edwards-cgi:
### Change description ###
Transcription data being migrated from the legacy system will not be associated to a workflow in modernised DARTS. Therefore, we will just display all transcription comments associated with a transcription where the trw_id is null in a generic 'Migrated legacy data comment' on the portal.

Therefore, a new field needs to be added to the {*}GET /transcriptions/{transcription_id{*}} called 'legacy_comments', which will be an array. 
{code:java}
{
...
   "legacy_comments" : [
      "comment1",
      "comment2"
   ]
...
}{code}

h4.Acceptance Criteria
*AC1: Return comments not associated with a workflow in a new 'legacy comments' list in the payload for GET /transcriptions/\{transcription_id}*

GIVEN

a get transcription request by id is made

WHEN

the request has comments migrated from legacy not associated with a workflow

THEN 

include the comments in a legacy_comments array

 

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/transcriptions/controller/TranscriptionControllerGetTranscriptionTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/transcriptions/controller/TranscriptionControllerGetTranscriptionTest.java
@@ -97,6 +97,34 @@ class TranscriptionControllerGetTranscriptionTest extends IntegrationBase {
     }
 
     @Test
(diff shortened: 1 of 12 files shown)


sabahirfan (approved):
LGTM

PR #2060, opened 2024-09-25
DMP-2559: Write to log when a case is deleted due to expiry 

sabahirfan:
DMP-2559: 

### Change description ###
Write to log when a case is deleted due to expiry 

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImplTest.java
```diff
@@ -178,6 +184,7 @@ void assertPositiveAnonymizeCourtCaseEntity() {
         doNothing().when(dataAnonymisationService).anonymizeProsecutorEntity(any(), any());
         doNothing().when(dataAnonymisationService).anonymizeHearingEntity(any(), any());
         doNothing().when(dataAnonymisationService).tidyUpTransformedMediaEntities(any(), any());
+        doNothing().when(casesLoggerService).caseDeletedDueToExpiry(anyInt(), anyString());
```
mario-paniccia, line 187: aren't these redundant? The default behaviour would be to do nothing

src/main/java/uk/gov/hmcts/darts/log/service/CasesLoggerService.java
```diff
@@ -9,4 +9,6 @@ public interface CasesLoggerService {
     void defendantNameOverflow(AddCaseRequest addCaseRequest);
 
     void defendantNotAdded(String defendant, String caseNumber);
+
+    void caseDeletedDueToExpiry(Integer caseId, String caseNumber);
```
davet1985, line 13: Could you expose this to other packages via the `LogApi` class please. We have a pattern to use an "api" package for methods that are used by other packages. Allowing us to easily see all usages of a method outside the package it's defined within.

sabahirfan: done

src/test/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImplTest.java
sabahirfan: removed all these unnecessary ones.

davet1985 (approved):

PR #2061, opened 2024-09-25
DMP-3984 (Dev Only) Fully implement continuation token

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-3984

### Change description ###

Implemented continuation token so that it iterates until batch size is met

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -332,7 +332,8 @@ darts:
         housekeeping: ecfd1f14-c9b6-4f15-94c7-cc60e53f2c7a
         dailylist-processor: f6f71122-ff85-4ebe-93d9-1706460dbea5
       unstructured-to-arm-batch-processor:
-        max-result-size: 100000
+        max-arm-manifest-items: 100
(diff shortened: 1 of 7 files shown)


src/main/resources/application.yaml
```diff
@@ -331,8 +332,7 @@ darts:
         housekeeping: ecfd1f14-c9b6-4f15-94c7-cc60e53f2c7a
         dailylist-processor: f6f71122-ff85-4ebe-93d9-1706460dbea5
       unstructured-to-arm-batch-processor:
-        max-arm-manifest-items: 100
-        max-arm-single-mode-items: 100
+        max-result-size: 100000
```
hemantasharma1129, line 335: Why do we have the batch size hardcoded in yaml. It should be fetched from DB which the existing code is already doing.

karen-hedges: I think that was a merge conflict. Will correct it.

hemantasharma1129 (approved):

PR #2076, opened 2024-09-30
Dmp 2630 implement dets to arm push flow

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-2630

### Change description ###

Implement the DETS To ARM Push flow

```
[ ] Yes
[ x] No
```


src/main/resources/application.yaml
```diff
@@ -284,6 +284,9 @@ darts:
       sas-endpoint: ${DETS_SAS_URL_ENDPOINT:}
       container-name: darts-migration
       temp-blob-workspace: ${user.home}/dets/tempworkspace
+      delete-timeout: 60
```
cmagowan-hmcts, line 287: whats the unit? do we want to make this into a duration?

src/main/java/uk/gov/hmcts/darts/dets/service/impl/DetsApiServiceImpl.java
```diff
@@ -65,4 +81,60 @@ public UUID saveBlobData(BinaryData binaryData) {
...
     }
+
+    @Override
+    public boolean deleteBlobDataFromContainer(UUID blobId) throws AzureDeleteBlobException {
+        try {
+            BlobServiceClient serviceClient = blobServiceFactory.getBlobServiceClientWithSasEndpoint(configuration.getSasEndpoint());
+            BlobContainerClient containerClient = blobServiceFactory.getBlobContainerClient(configuration.getContainerName(), serviceClient);
+            BlobClient client = blobServiceFactory.getBlobClient(containerClient, blobId);
+            Response<Boolean> response = client.deleteIfExistsWithResponse(DeleteSnapshotsOptionType.INCLUDE, null,
+                                                                           Duration.of(
```
cmagowan-hmcts, line 92: maybe config should just be duration, and use 60s or 1m in application.yaml

src/main/java/uk/gov/hmcts/darts/dets/config/DetsDataManagementConfiguration.java
```diff
@@ -13,4 +13,9 @@
 public class DetsDataManagementConfiguration extends StorageConfiguration {
     private String containerName;
     private String sasEndpoint;
+    private String detsManifestFilePrefix;
```
cmagowan-hmcts, line 16: i dont see this in application.yaml

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
```diff
@@ -100,18 +100,28 @@ Optional<ExternalObjectDirectoryEntity> findMatchingExternalObjectDirectoryEntit
...
+    @Query(
+        """
+            SELECT eod FROM ExternalObjectDirectoryEntity eod
+            WHERE eod.status in :failedStatuses
+            AND eod.externalLocationType = :type
+            AND eod.transferAttempts <= :transferAttempts
+            AND eod.osrUuid is not null
+            """
+    )
+    List<ExternalObjectDirectoryEntity> findNotFinishedAndNotExceededRetryInStorageLocationForDets(List<ObjectRecordStatusEntity> failedStatuses,
```
cmagowan-hmcts, line 120: is this different to line 103?

src/main/resources/application.yaml
```diff
@@ -337,6 +340,9 @@ darts:
       unstructured-to-arm-batch-processor:
         max-arm-manifest-items: 100
         max-arm-single-mode-items: 100
+      dets-to-arm-batch-processor:
+        max-arm-manifest-items: 100
```
cmagowan-hmcts, line 344: is this 100 not common across the whole application?

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmBatchProcessorImpl.java
```diff
@@ -57,22 +51,19 @@ public class UnstructuredToArmBatchProcessorImpl implements UnstructuredToArmBat
...
 
-    private static final String DARTS_STRING = "darts";
-    private static final String DETS_STRING = "dets";
 
     @Override
     @SuppressWarnings({"PMD.AvoidInstantiatingObjectsInLoops", "PMD.CognitiveComplexity", "PMD.CyclomaticComplexity"})
-
     public void processUnstructuredToArm(int taskBatchSize) {
 
         log.info("Started running ARM Batch Push processing at: {}", OffsetDateTime.now());
```
mario-paniccia, line 59: any reason not to use the `currentTimeHelper` that we have available in a field already?

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
```diff
@@ -100,18 +100,28 @@ Optional<ExternalObjectDirectoryEntity> findMatchingExternalObjectDirectoryEntit
             WHERE eod.status in :failedStatuses
             AND eod.externalLocationType = :type
             AND eod.transferAttempts <= :transferAttempts
+            AND eod.osrUuid is null
```
mario-paniccia, line 103: don't we have to review and make a similar update to other queries as well? for example this one:
```
    List<ExternalObjectDirectoryEntity> findByEntityAndStatus(MediaEntity media, ObjectRecordStatusEntity status);
```
This could now return a mix of DARTS and DETS EODs right?

src/main/java/uk/gov/hmcts/darts/arm/helper/DataStoreToArmHelper.java
```diff
@@ -142,7 +153,10 @@ public ExternalObjectDirectoryEntity createArmExternalObjectDirectoryEntity(Exte
         armExternalObjectDirectoryEntity.setCreatedBy(userAccount);
         armExternalObjectDirectoryEntity.setLastModifiedBy(userAccount);
         armExternalObjectDirectoryEntity.setTransferAttempts(1);
-
+        // This is only set for DETS records
+        if (nonNull(externalObjectDirectory.getOsrUuid())) {
+            armExternalObjectDirectoryEntity.setOsrUuid(externalObjectDirectory.getOsrUuid());
+        }
```
mario-paniccia, line 159: since this is meant to be a generic method, can't this DETS specific change be move up a level to the DETS code that is calling this?

src/main/java/uk/gov/hmcts/darts/arm/helper/DataStoreToArmHelper.java
```diff
@@ -208,4 +218,80 @@ public void updateExternalObjectDirectoryFailedTransferAttempts(ExternalObjectDi
...
+        if (matchingEntity.isPresent()) {
+            batchItem.setSourceEod(matchingEntity.get());
+            batchItems.add(batchItem);
+            armEod.setManifestFile(archiveRecordsFile.getName());
+            incrementTransferAttempts(armEod);
+            updateExternalObjectDirectoryStatus(armEod, EodHelper.armIngestionStatus(), userAccount);
+        } else {
+            log.error("Unable to find matching external object directory for {}", armEod.getId());
+            updateExternalObjectDirectoryFailedTransferAttempts(armEod, userAccount);
+            throw new RuntimeException(MessageFormat.format("Unable to find matching external object directory for {0}", armEod.getId()));
```
mario-paniccia, line 249: shall we use `DartsException`?

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
```diff
@@ -466,4 +476,36 @@ AND eod.externalLocationType.id IN (1, 2)
...
+            )
+            union
+            (
+                select file_size as fileSize from darts.transcription_document as trd
+                join darts.external_object_directory as eod on eod.trd_id = trd.trd_id
+                where eod.eod_id = :externalObjectDirectoryId
+            )
+        ) a
+        """, nativeQuery = true)
+    Long findFileSize(Integer externalObjectDirectoryId);
```
mario-paniccia, line 508: could we do this in code since we have already the Eod entity we can use?
If lazy loading works we could navigate from there to the not null associated object (media, transcript...), otherwise we could just fetch the associated object from the repo and get its size?
This way we can avoid this complex native very specific query

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmBatchProcessorImpl.java
```diff
@@ -45,7 +39,7 @@
 public class UnstructuredToArmBatchProcessorImpl implements UnstructuredToArmBatchProcessor {
```
mario-paniccia, line 39: We now have 2 batch push processors:
- `UnstructuredToArmBatchProcessor`
- `DetsToArmBatchPushProcessor`
I think it would be useful if they implemented a common top level `ArmBatchProcessor` "tag" interface. 
It's getting more and more difficult to keep track of all these Arm related processors and their multiple variants

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
mario-paniccia, line 103: this is interesting. I wonder whether some of the other repo methods would need a similar change applied to make sure they filter out the DETS EOD records.
@hemantasharma1129 ?

src/main/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchPushProcessorImpl.java
```diff
@@ -0,0 +1,340 @@
...
+            var batchItem = new ArmBatchItem();
+            try {
+                ExternalObjectDirectoryEntity armEod;
+                if (isEqual(currentEod.getExternalLocationType(), EodHelper.armLocation())) {
+                    //retry existing attempt that has previously failed.
+                    armEod = currentEod;
+                    batchItem.setArmEod(armEod);
+                    dataStoreToArmHelper.updateArmEodToArmIngestionStatus(
+                        currentEod, batchItem, batchItems, archiveRecordsFile, userAccount, EodHelper.detsLocation());
+                    getObjectStateRecordEntity(armEod);
```
mario-paniccia, line 131: was this meant to be:
```
objectStateRecord =  getObjectStateRecordEntity(armEod);
```
?

src/main/java/uk/gov/hmcts/darts/dets/service/impl/DetsApiServiceImpl.java
karen-hedges: This was copied from Marios ticket which has now been merged so will look into this.

src/main/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchPushProcessorImpl.java
```diff
@@ -0,0 +1,340 @@
...
+            objectStateRecordRepository.save(objectStateRecord);
+        } else {
+            if (nonNull(externalObjectDirectory)) {
+                log.error("Unable to find ObjectStateRecordEntity for ARM EOD ID: {}", externalObjectDirectory.getId());
+            }
+        }
+    }
+
+    private ObjectStateRecordEntity getObjectStateRecordEntity(ExternalObjectDirectoryEntity externalObjectDirectory) {
+        if (nonNull(externalObjectDirectory.getOsrUuid())) {
```
mario-paniccia, line 233: I don't think this is necessary, we can never be in a situation where this field is `null` because at the very beginning the processor has only retrieved from the DB EODs with ors uuid not null. 
I would remove it since it complicates an already complex code

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
karen-hedges: yes the difference is whether osrUuid is not null whereas the original one checks if it is null. This maybe the only way to distinguish between migration ARM and DETS ARM

src/main/java/uk/gov/hmcts/darts/arm/helper/DataStoreToArmHelper.java
karen-hedges: I added this just to highlight the difference between DETS and Unstructured. Its doesn't actually need the nonNull check.

src/main/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchPushProcessorImpl.java
```diff
@@ -0,0 +1,340 @@
...
+        for (var batchItem : batchItems.getFailed()) {
+            String errorMessage = "Manifest file creation failed";
+            updateObjectStateRecordStatus(batchItem.getArmEod(), errorMessage);
+        }
+
+    }
+
+    private void updateObjectStateRecordStatus(ExternalObjectDirectoryEntity externalObjectDirectory, String errorMessage) {
+        ObjectStateRecordEntity objectStateRecord = getObjectStateRecordEntity(externalObjectDirectory);
+        if (nonNull(objectStateRecord)) {
```
mario-paniccia, line 218: this null check is not needed since `getObjectStateRecordEntity` throws Exception if it can't find the OSR.
I would removed and simplify the code

src/main/resources/application.yaml
karen-hedges: This can be changed per automated task and is specific to arm so its how many items they can accept in batch mode.

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
karen-hedges: This was considered the most optimal query wise. If we add @Transactional it will implode the amount of queries called.

src/main/java/uk/gov/hmcts/darts/arm/helper/DataStoreToArmHelper.java
karen-hedges: can do

src/main/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchPushProcessorImpl.java
```diff
@@ -0,0 +1,340 @@
...
+            if (nonNull(objectStateRecord.getObjectStatus())) {
+                objectStateRecord.setObjectStatus(objectStateRecord.getObjectStatus() + " " + errorMessage);
```
mario-paniccia, line 220: if we append don't we have the issue that we would be preserving "old" errors that happened on previous retries? Those would be no longer relevant in the next retries

src/main/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchPushProcessorImpl.java
```diff
@@ -0,0 +1,340 @@
...
+                        currentEod, batchItem, batchItems, archiveRecordsFile, userAccount, EodHelper.detsLocation());
+                    getObjectStateRecordEntity(armEod);
+
+                } else {
+
+                    armEod = dataStoreToArmHelper.createArmEodWithArmIngestionStatus(
+                        currentEod, batchItem, batchItems, archiveRecordsFile, userAccount);
+                    objectStateRecord = getObjectStateRecordEntity(currentEod);
+                    // Save the new ARM EOD to the object state record
+                    if (nonNull(objectStateRecord)) {
```
mario-paniccia, line 139: not needed null check, I would remove

src/main/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchPushProcessorImpl.java
karen-hedges: yes when I added this it was because I wanted to make sure the cascading would pick up all errors as its possible to get more than one error but primarily it would only happen if something went wrong and then when its trying to clean up something else goes wrong. This was what I was referring to on your ticket about appending.

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
cmagowan-hmcts: i'm not sure whats best here. I'm not the biggest fan of native queries, but in this scenario, it means 1 query can be used instead of 4, so it is more performant. 
Its a question i have asked before, but no real guidance.

src/main/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchPushProcessorImpl.java
```diff
@@ -0,0 +1,340 @@
...
+            return false;
+        }
+
+        return true;
+    }
+
+    private void setRawDataPushed(ExternalObjectDirectoryEntity detsExternalObjectDirectory, ObjectStateRecordEntity objectStateRecord) {
+        if (nonNull(objectStateRecord)) {
+            objectStateRecord.setFlagFileTransferToDets(true);
+            objectStateRecord.setDateFileTransferToDets(currentTimeHelper.currentOffsetDateTime());
```
mario-paniccia, line 308: the ticket says to update the below fields instead:
`flag_file_transf_toarml`
`date_file_transf_toarml`
I guess the ticket if wrong, can we get that updated if that's the case please

src/main/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchPushProcessorImpl.java
```diff
@@ -0,0 +1,340 @@
...
+
+    private void updateObjectStateRecordManifestSuccessOrFailure(ArmBatchItems batchItems, File archiveRecordsFile) {
+        for (var batchItem : batchItems.getSuccessful()) {
+            ObjectStateRecordEntity objectStateRecord = getObjectStateRecordEntity(batchItem.getArmEod());
+            if (nonNull(objectStateRecord)) {
+                objectStateRecord.setFlagFileMfstCreated(true);
+                objectStateRecord.setDateFileMfstCreated(currentTimeHelper.currentOffsetDateTime());
+                objectStateRecord.setIdManifestFile(archiveRecordsFile.getName());
+                objectStateRecord.setFlagMfstTransfToArml(true);
+                objectStateRecord.setDateMfstTransfToArml(currentTimeHelper.currentOffsetDateTime());
```
mario-paniccia, line 204: the ticket says that these 2 fields should be set only after the manifest file has been pushed to ARM, which hasn't happened at this stage yet if I'm not wrong. 
Should these be moved inside the next step instead?:
```
copyMetadataToArm(archiveRecordsFile);
```

src/main/resources/application.yaml
karen-hedges: This was copied from Marios PR when I needed the delete code he had added but it appears this was replaced with a different config setting.

src/main/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchPushProcessorImpl.java
karen-hedges: We aren't using the object state record here but we are verifying that it exists. If not an exception is thrown and this EOD will be skipped.

src/main/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchPushProcessorImpl.java
mario-paniccia: the dets fields will be populated by migration we don't touch them

karen-hedges: I picked the wrong fields

src/main/java/uk/gov/hmcts/darts/arm/service/impl/UnstructuredToArmBatchProcessorImpl.java
hemantasharma1129: yea its work optimising it in future but not part of this ticekt.

src/main/java/uk/gov/hmcts/darts/arm/helper/DataStoreToArmHelper.java
mario-paniccia: I was not referring to the null check, the whole block is very DETS specific I think doesn't really belong here for that reason

karen-hedges: No it isn't. This is shared by both unstructured and dets where they copy the original eod and make a new arm one. All I was trying to highlight is that the setting the setOsrUuid should be null for unstructured but have a value for dets.

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
hemantasharma1129: Im not sure if UNION is performant. I am inclined towards @mario-paniccia solution i.e. achieving this via entity association rather than native query with UNION @cmagowan-hmcts. Happy to discuss it further.

mario-paniccia: with my suggested solution it's going to be only one query to get by id one of media, transcript, annotation or case document. Since we already have the EOD entity

src/main/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchPushProcessorImpl.java
```diff
@@ -0,0 +1,376 @@
...
+                annotationDocument -> Long.valueOf(annotationDocument.getFileSize())).orElse(null);
+        } else if (nonNull(detsExternalObjectDirectory.getCaseDocument())) {
+            fileSize = caseDocumentRepository.findById(detsExternalObjectDirectory.getCaseDocument().getId()).map(
+                caseDocument -> Long.valueOf(caseDocument.getFileSize())).orElse(null);
+        } else if (nonNull(detsExternalObjectDirectory.getTranscriptionDocumentEntity())) {
+            fileSize = transcriptionDocumentRepository.findById(detsExternalObjectDirectory.getTranscriptionDocumentEntity().getId()).map(
+                transcriptionDocument -> Long.valueOf(transcriptionDocument.getFileSize())).orElse(null);
+        }
+        return fileSize;
+    }
```
mario-paniccia, line 373: this could live in the EodService, would reduce the complexity of this already complex class

src/main/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchPushProcessorImpl.java
```diff
@@ -0,0 +1,376 @@
...
+            String errorMessage = "Manifest file creation failed";
+            updateObjectStateRecordStatus(batchItem.getArmEod(), errorMessage);
+        }
+    }
+
+    private String getManifestFilePrefix() {
+        return detsToArmProcessorConfiguration.getManifestFilePrefix();
+    }
+
+    private void updateObjectStateRecordManifestSuccessOrFailure(ArmBatchItems batchItems, File archiveRecordsFile, boolean successfulTransfer) {
```
mario-paniccia, line 221:
```suggestion
    private void updateObjectStateRecordManifestSuccessOrFailure(ArmBatchItems batchItems, File archiveRecordsFile, boolean isSuccessfulTransfer) {
```

src/main/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchPushProcessorImpl.java
```diff
@@ -0,0 +1,376 @@
...
+            } else {
+                log.warn("No EODs were able to be processed, skipping manifest file creation");
+            }
+        } catch (Exception e) {
+            String errorMessage = String.format("Error during generation of DETS batch manifest file %s", archiveRecordsFile.getName());
+            log.error(errorMessage, e);
+            batchItems.getSuccessful().forEach(batchItem -> dataStoreToArmHelper.recoverByUpdatingEodToFailedArmStatus(batchItem, userAccount));
+            final String errorMessageWithStackTrace = errorMessage + " - " + ExceptionUtils.getStackTrace(e);
+            batchItems.getFailed().forEach(batchItem -> updateObjectStateRecordStatus(batchItem.getArmEod(), errorMessageWithStackTrace));
+            updateObjectStateRecordManifestSuccessOrFailure(batchItems, archiveRecordsFile, false);
```
mario-paniccia, line 196: when ANY exception happens this method is:

- setFlagMfstTransfToArml(false)
- setting error message on OSR status to `"Manifest file creation failed"`.

but that exception could NOT be to related to the manifest file copy to ARM, it could have happened in a step before the copy. And so setting that "error status" on the OSR is a bit misleading I think

src/main/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchPushProcessorImpl.java
```diff
@@ -0,0 +1,376 @@
...
+    private void updateObjectStateRecordManifestSuccessOrFailure(ArmBatchItems batchItems, File archiveRecordsFile, boolean successfulTransfer) {
+        for (var batchItem : batchItems.getSuccessful()) {
+            ObjectStateRecordEntity objectStateRecord = getObjectStateRecordEntity(batchItem.getArmEod());
+            objectStateRecord.setFlagMfstTransfToArml(successfulTransfer);
+            objectStateRecord.setDateMfstTransfToArml(currentTimeHelper.currentOffsetDateTime());
+            objectStateRecordRepository.save(objectStateRecord);
+        }
+
+        for (var batchItem : batchItems.getFailed()) {
+            String errorMessage = "Manifest file creation failed";
```
mario-paniccia, line 230:
```suggestion
            String errorMessage = "Manifest file transfer to ARM failed";
```

hemantasharma1129 (approved):

PR #2096, opened 2024-10-02
Fix problem whereby we do not consider inactive users unless the endp…

mestebanez:

NO JIRA

### Change description ###

Always authenticate against the user being inactive.  If the user is inactive we return a 401 across the APIs

This fixes a bug whereby if an endpoint does not have authorisation enabled e.g. /transcriptions/transcriber-view?assigned=false, then an inactive user can use the endpoint successfully.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/docker-compose-local.yml
+++ b/docker-compose-local.yml
@@ -12,7 +12,6 @@ services:
       - DARTS_API_DB_USERNAME=darts
       - DARTS_API_DB_PASSWORD=darts
       - DARTS_API_DB_SCHEMA=darts
-      - SPRING_PROFILES_ACTIVE=local
       - AAD_B2C_CLIENT_ID
       - AAD_B2C_CLIENT_SECRET
       - AAD_B2C_TENANT_ID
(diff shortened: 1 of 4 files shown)


docker-compose-local.yml
```diff
@@ -55,12 +54,14 @@ services:
       - MAX_FILE_UPLOAD_REQUEST_SIZE_MEGABYTES=360
       - DARTS_INBOUND_STORAGE_SAS_URL
       - DARTS_UNSTRUCTURED_STORAGE_SAS_URL
+      - JAVA_TOOL_OPTIONS=-agentlib:jdwp=transport=dt_socket,address=*:8003,server=y,suspend=n
```
mestebanez, line 57: Add a docker debug port. Allows us to debug into the docker container using intellij

mestebanez:
Closing. See PR https://github.com/hmcts/darts-api/pull/2155

mestebanez:
See https://github.com/hmcts/darts-api/pull/2155

src/main/java/uk/gov/hmcts/darts/authorisation/component/impl/UserIdentityImpl.java
```diff
@@ -34,78 +34,88 @@ public class UserIdentityImpl implements UserIdentity {
...
-            if (principalObject instanceof Jwt jwt) {
-                Object emailsAddressesObject = jwt.getClaims().get(EMAILS);
-                if (emailsAddressesObject == null) {
-                    emailsAddressesObject = jwt.getClaims().get(PREFERRED_USERNAME);
+    private String getEmailAddressFromToken(Jwt token) {
+        Object principalObject = token;
+
+        if (principalObject instanceof Jwt jwt) {
+            Object emailsAddressesObject = jwt.getClaims().get(EMAILS);
+            if (emailsAddressesObject == null) {
```
hemantasharma1129, line 42: We have to add "email" claim as well which we have recently added. The sequence should be EMAILS -> EMAIL -> PREFERRED_USERNAME.

PR #2114, opened 2024-10-07
DMP-3996 - Manual obfuscation of event text endpoint

Ben-Edwards-cgi:
### Change description ###
# Summary of Git Diff

This Git diff introduces new functionality for event obfuscation in the DARTS application. It adds tests for obfuscating event texts, updates existing service interfaces, and modifies the controller to include new API endpoints for this feature. Additionally, it implements error handling for scenarios when events are not found.

## Highlights

- **New Tests Added**: 
  - `adminObfuscateEveByIdsSingle`: Tests obfuscation of a single event.
  - `adminObfuscateEveByIdsMultiple`: Tests obfuscation of multiple events.
  - `adminObfuscateEveByIdsNotFound`: Tests API response when trying to obfuscate a non-existent event.
  - `adminObfuscateEveByIdsNotSuperAdmin`: Ensures that non-super admin users cannot obfuscate events.

- **New Methods in Service Interface**:
  - `obfuscateEventByIds(List<Integer> eveIds)`: Added to `DataAnonymisationService`.
  - `getEventEntityById(Integer eveId)`: Retrieves event details by ID.
  - `saveEvent(EventEntity eventEntity)`: Saves the modified event entity.

- **Controller Modification**:
  - New endpoint `/admin/events/obfuscate` for obfuscating event texts added to `EventsController`.

- **Error Handling**:
  - Introduced `DartsApiException.DartsApiErrorCommon.NOT_FOUND` for handling cases where events do not exist.

- **OpenAPI Specification Updated**:
  - The OpenAPI specification (`event.yaml`) reflects the new obfuscation feature and its associated responses.

- **New Unit Tests**: 
  - Added tests for `EventServiceImpl` to validate the behavior of new methods, especially for retrieving and saving events.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerTest.java
@@ -186,7 +186,7 @@ class EventsControllerTest extends IntegrationBase {
             .contentType(MediaType.APPLICATION_JSON_VALUE)
             .content("{\"eve_ids\":[" + event.getId() + "," + event2.getId() + "]}");
 
-        mockMvc.perform(requestBuilder).andExpect(status().isNoContent());
+        mockMvc.perform(requestBuilder).andExpect(status().isOk());
 
         EventEntity editedEventEntity = dartsDatabaseStub.getEventRepository().findById(event.getId()).orElseThrow();
         assertThat(editedEventEntity.getEventText()).matches(UUID_REGEX);
--- a/src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerTest.java
@@ -163,7 +163,7 @@ class EventsControllerTest extends IntegrationBase {
             .contentType(MediaType.APPLICATION_JSON_VALUE)
             .content("{\"eve_ids\":[" + event.getId() + "]}");
 
-        mockMvc.perform(requestBuilder).andExpect(status().isNoContent());
+        mockMvc.perform(requestBuilder).andExpect(status().isOk());
 
         EventEntity editedEventEntity = dartsDatabaseStub.getEventRepository().findById(event.getId()).orElseThrow();
         assertThat(editedEventEntity.getEventText()).matches(UUID_REGEX);
--- a/src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerTest.java
@@ -174,6 +174,28 @@ class EventsControllerTest extends IntegrationBase {
     }
 
     @Test
+    void adminObfuscateEveByIdsMultiple() throws Exception {
+        HearingEntity hearing = dartsDatabaseStub.createHearing("Courthouse", "1", "12345", LocalDateTime.now());
+
+        EventEntity event = dartsDatabaseStub.createEvent(hearing);
+        EventEntity event2 = dartsDatabaseStub.createEvent(hearing);
+
+
+        given.anAuthenticatedUserWithGlobalAccessAndRole(SUPER_ADMIN);
+        MockHttpServletRequestBuilder requestBuilder = post("/admin/events/obfuscate")
+            .contentType(MediaType.APPLICATION_JSON_VALUE)
+            .content("{\"eve_ids\":[" + event.getId() + "," + event2.getId() + "]}");
+
+        mockMvc.perform(requestBuilder).andExpect(status().isNoContent());
+
+        EventEntity editedEventEntity = dartsDatabaseStub.getEventRepository().findById(event.getId()).orElseThrow();
+        assertThat(editedEventEntity.getEventText()).matches(UUID_REGEX);
+
+        EventEntity editedEventEntity2 = dartsDatabaseStub.getEventRepository().findById(event2.getId()).orElseThrow();
+        assertThat(editedEventEntity2.getEventText()).matches(UUID_REGEX);
+    }
+
+    @Test
     void adminObfuscateEveByIdsNotFound() throws Exception {
         given.anAuthenticatedUserWithGlobalAccessAndRole(SUPER_ADMIN);
         MockHttpServletRequestBuilder requestBuilder = post("/admin/events/obfuscate")
--- a/src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerTest.java
@@ -193,6 +193,7 @@ class EventsControllerTest extends IntegrationBase {
 
         EventEntity editedEventEntity2 = dartsDatabaseStub.getEventRepository().findById(event2.getId()).orElseThrow();
         assertThat(editedEventEntity2.getEventText()).matches(UUID_REGEX);
+        assertThat(editedEventEntity2.getEventText()).isNotEqualTo(editedEventEntity.getEventText());
     }
 
     @Test
@@ -207,5 +208,20 @@ class EventsControllerTest extends IntegrationBase {
 
         Problem responseResult = objectMapper.readValue(response.getResponse().getContentAsString(), Problem.class);
         Assertions.assertEquals(DartsApiException.DartsApiErrorCommon.NOT_FOUND.getType(), responseResult.getType());
+    }
+
+    @ParameterizedTest(name = "User with role {0} should not be able to obfuscate events")
+    @EnumSource(value = SecurityRoleEnum.class, names = {"SUPER_ADMIN"}, mode = EnumSource.Mode.EXCLUDE)
+    void adminObfuscateEveByIdsNotSuperAdmin(SecurityRoleEnum securityRoleEnum) throws Exception {
+        given.anAuthenticatedUserWithGlobalAccessAndRole(securityRoleEnum);
+        MockHttpServletRequestBuilder requestBuilder = post("/admin/events/obfuscate")
+            .contentType(MediaType.APPLICATION_JSON_VALUE)
+            .content("{\"eve_ids\":[1]}");
+
+        MvcResult response = mockMvc.perform(requestBuilder).andExpect(status().isForbidden())
+            .andReturn();
+
+        Assertions.assertEquals("{\"type\":\"AUTHORISATION_109\",\"title\":\"User is not authorised for this endpoint\",\"status\":403}",
+                                response.getResponse().getContentAsString());
     }
 }
--- a/src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerTest.java
@@ -15,6 +15,7 @@ import uk.gov.hmcts.darts.audio.api.AudioApi;
 import uk.gov.hmcts.darts.common.entity.EventEntity;
 import uk.gov.hmcts.darts.common.entity.HearingEntity;
 import uk.gov.hmcts.darts.common.enums.SecurityRoleEnum;
+import uk.gov.hmcts.darts.common.exception.DartsApiException;
 import uk.gov.hmcts.darts.event.component.DartsEventMapper;
 import uk.gov.hmcts.darts.event.exception.EventError;
 import uk.gov.hmcts.darts.event.model.AdminGetEventForIdResponseResult;
@@ -28,6 +29,7 @@ import uk.gov.hmcts.darts.testutils.stubs.DartsDatabaseStub;
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.verify;
 import static org.skyscreamer.jsonassert.JSONAssert.assertEquals;
@@ -37,6 +39,7 @@ import static org.springframework.test.web.servlet.request.MockMvcRequestBuilder
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.CPP;
 import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.SUPER_ADMIN;
+import static uk.gov.hmcts.darts.test.common.TestUtils.UUID_REGEX;
 
 @AutoConfigureMockMvc
 class EventsControllerTest extends IntegrationBase {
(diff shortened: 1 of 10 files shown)


src/test/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImplTest.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+
+    @Test
+    void positiveGetEventEntityByIdNotFound() {
+        when(eventRepository.findById(1)).thenReturn(Optional.empty());
+
+        assertThatThrownBy(() -> eventService.getEventEntityById(1))
+            .isInstanceOf(DartsApiException.class)
+            .hasFieldOrPropertyWithValue("error", DartsApiException.DartsApiErrorCommon.NOT_FOUND);
```
mestebanez, line 83: Optional: This is fine but we could use the getError method rather than binding ourself to field name. Ensures that compilation failure soccur on any future change.

Ben-Edwards-cgi: The getError method does not work with the assertThatThrownBy(...) assertJ method :)

src/test/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImplTest.java
```diff
@@ -0,0 +1,87 @@
...
+        eventService.adminObfuscateEveByIds(List.of(1, 2, 3, 4));
+
+
+        verify(dataAnonymisationService, times(1)).anonymizeEvent(event1);
+        verify(dataAnonymisationService, times(1)).anonymizeEvent(event2);
+        verify(dataAnonymisationService, times(1)).anonymizeEvent(event3);
+
+        verify(eventService, times(1)).getEventEntityById(1);
+        verify(eventService, times(1)).getEventEntityById(2);
+        verify(eventService, times(1)).getEventEntityById(3);
```
mestebanez, line 64: Add verify(eventService, times(1)).getEventEntityById(4);

Ben-Edwards-cgi: Id 4 was added to verify that we only process distinct events in this case id 4 relates to event1 so is not unique as such would not expect it to be called :)

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
```diff
@@ -19,11 +23,26 @@ public class EventServiceImpl implements EventService {
...
+    public void adminObfuscateEveByIds(List<Integer> eveIds) {
+        eveIds.stream()
+            .map(this::getEventEntityById)
+            .distinct()
+            .forEach(dataAnonymisationService::anonymizeEvent);
+    }
+
+    EventEntity getEventEntityById(Integer eveId) {
+        return eventRepository.findById(eveId)
+            .orElseThrow(() -> new DartsApiException(DartsApiException.DartsApiErrorCommon.NOT_FOUND));
```
mestebanez, line 46: dont we want an operation specific error code?

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
```diff
@@ -19,11 +23,26 @@ public class EventServiceImpl implements EventService {
...
+    @Transactional
+    public void adminObfuscateEveByIds(List<Integer> eveIds) {
+        eveIds.stream()
+            .map(this::getEventEntityById)
+            .distinct()
+            .forEach(dataAnonymisationService::anonymizeEvent);
+    }
+
+    EventEntity getEventEntityById(Integer eveId) {
+        return eventRepository.findById(eveId)
```
mestebanez, line 45: The requirements seem to talk about:-

When the event_id is zero for the relevant eve_id needs to be obfuscated

src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerTest.java
```diff
@@ -37,6 +39,7 @@
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.CPP;
 import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.SUPER_ADMIN;
+import static uk.gov.hmcts.darts.test.common.TestUtils.UUID_REGEX;
 
 @AutoConfigureMockMvc
```
mestebanez, line 44: Add test to ensure we only process for Admins and it fails for others

src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerTest.java
```diff
@@ -146,4 +149,63 @@ void adminEventsApiGetByIdEndpointFailureNoEventFound() throws Exception {
...
+            .contentType(MediaType.APPLICATION_JSON_VALUE)
+            .content("{\"eve_ids\":[" + event.getId() + "," + event2.getId() + "]}");
+
+        mockMvc.perform(requestBuilder).andExpect(status().isNoContent());
+
+        EventEntity editedEventEntity = dartsDatabaseStub.getEventRepository().findById(event.getId()).orElseThrow();
+        assertThat(editedEventEntity.getEventText()).matches(UUID_REGEX);
+
+        EventEntity editedEventEntity2 = dartsDatabaseStub.getEventRepository().findById(event2.getId()).orElseThrow();
+        assertThat(editedEventEntity2.getEventText()).matches(UUID_REGEX);
```
mestebanez, line 195: Maybe check the the uuid is not the same across the two event entities

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
Ben-Edwards-cgi: So the eve_id zero comes from a old requirement which has since been removed around obfuscating previous versions of events but this has been moved to a new ticket now.

This ticket exclusively obfuscates by eve_id

src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerTest.java
Ben-Edwards-cgi: Updated

src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerTest.java
Ben-Edwards-cgi: Added

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
```diff
@@ -19,11 +23,26 @@ public class EventServiceImpl implements EventService {
...
     }
+
+    @Override
+    @Transactional
+    public void adminObfuscateEveByIds(List<Integer> eveIds) {
+        eveIds.stream()
+            .map(this::getEventEntityById)
+            .distinct()
+            .forEach(dataAnonymisationService::anonymizeEvent);
+    }
```
mario-paniccia, line 42: Why don't we directly expose a new method `anonymiseEvents` on the obfuscation service that takes the list of ids and have the controller call that direclty?
Seems a bit redundant to me going through the EventService

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
```diff
@@ -19,11 +23,26 @@ public class EventServiceImpl implements EventService {
...
     @Override
     public AdminGetEventForIdResponseResult adminGetEventById(Integer eventId) {
         eventIdValidator.validate(eventId);
         Optional<EventEntity> eventEntityOptional = eventRepository.findById(eventId);
         return eventMapper.mapToAdminGetEventsResponseForId(eventEntityOptional);
     }
+
+    @Override
+    @Transactional
+    public void adminObfuscateEveByIds(List<Integer> eveIds) {
```
mario-paniccia, line 37:
```suggestion
    public void adminObfuscateEveByIds(List<Integer> eventIds) {
```

mario-paniccia: here and other places too.

mario-paniccia, line 37: we could consider moving this method to the obfuscation service. And have the controller call that directly without going through the EventService.

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
mario-paniccia: there seems to be already an event specific error code we can reuse, see [here](https://github.com/hmcts/darts-api/blob/a67c3b3234dc52445c5a34e4ef4e78808e2821b9/src/main/java/uk/gov/hmcts/darts/event/validation/EventIdValidator.java#L24).
(even thought the enum name is quite bad :) )

src/main/resources/openapi/event.yaml
```diff
@@ -535,6 +535,37 @@ paths:
...
+              required:
+                - eve_ids
+              properties:
+                eve_ids:
+                  minLength: 1
+                  type: array
+                  items:
+                    type: integer
+      responses:
+        '201':
```
mario-paniccia, line 561: status code for NO_CONTENT is 204

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
Ben-Edwards-cgi: Agreed updated

src/main/resources/openapi/event.yaml
Ben-Edwards-cgi: Good spot updated

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
Ben-Edwards-cgi: Agreed updated

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
Ben-Edwards-cgi: For non specific errors such as 404 I think using common errors across the bored make sense instead of entity specific ones such as EventApiError.NOT_FOUND, CourtApiError.NOT_FOUND, etc

src/test/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImplTest.java
mestebanez: but this method does get called? I verified this verify statement locally. Doesnt harm to add it in

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
mestebanez: If we want to define a common 404 error then thats fine, but the response body should always contain a problem with an associated app error code. The app error code provides the unique code around the specific operation failure. Thats the pattern we have adopted across the system so far.

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
mestebanez: oh ok, thats fine. Can we mention that on the ticket if its not already called out

src/test/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImplTest.java
Ben-Edwards-cgi: Added see above

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
Ben-Edwards-cgi: Error codes do still get added using the prefix COMMON

So not found would be COMMON_NOT_FOUND

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
Ben-Edwards-cgi: @jackmaloney see above

src/main/resources/openapi/event.yaml
```diff
@@ -535,6 +535,37 @@ paths:
...
+              required:
+                - eve_ids
+              properties:
+                eve_ids:
+                  minLength: 1
+                  type: array
+                  items:
+                    type: integer
+      responses:
+        '204':
```
mario-paniccia, line 561: just noticed now that the LLD states response should be 200. If you believe we should be returning 204 instead, it would be good to raise it and get the LLD updated accordingly

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
Ben-Edwards-cgi: @jackmaloney  What are your thoughts?

src/main/resources/openapi/event.yaml
Ben-Edwards-cgi: Good spot updated

mestebanez (approved):

src/main/resources/openapi/event.yaml
```diff
@@ -860,6 +892,7 @@ components:
         MAPPING_IN_USE,
         TOO_MANY_RESULTS,
         EVENT_ID_NOT_FOUND,
+        ALREADY_OBFUSCATED
```
davet1985, line 895: This error doesn't appear to be used and I don't see any use for it. If the event is already obfuscated then we just return 200.

Can it be removed?

davet1985 (commented):
One comment, otherwise LGTM

src/main/resources/openapi/event.yaml
Ben-Edwards-cgi: Agreed updated

src/main/java/uk/gov/hmcts/darts/event/controller/EventsController.java
```diff
@@ -168,6 +171,15 @@ public ResponseEntity<List<AdminSearchEventResponseResult>> adminSearchEvents(Ad
...
-        return  new ResponseEntity<>(eventService.adminGetEventById(eventId), HttpStatus.OK);
+        return new ResponseEntity<>(eventService.adminGetEventById(eventId), HttpStatus.OK);
+    }
+
+    @Override
+    @SecurityRequirement(name = SECURITY_SCHEMES_BEARER_AUTH)
+    @Authorisation(contextId = ANY_ENTITY_ID,
+        globalAccessSecurityRoles = {SUPER_ADMIN})
+    public ResponseEntity<Void> adminObfuscateEveByIds(AdminObfuscateEveByIdsRequest adminObfuscateEveByIdsRequest) {
+        this.dataAnonymisationService.obfuscateEventByIds(adminObfuscateEveByIdsRequest.getEveIds());
```
cmagowan-hmcts, line 182: do you need the "this" bit?

cmagowan-hmcts (approved):

mestebanez (approved):

src/main/java/uk/gov/hmcts/darts/event/controller/EventsController.java
Ben-Edwards-cgi: This is just a preference thing :D

When the code gets compiled it will auto add this I just like it as it makes things more clear

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
jackmaloney: Sorry only just seen this, and now it's been merged. 
I think we should be following the pattern that we have used elsewhere instead of deviating. If we think we need to change to a different pattern then fine, but as Mario says it needs agreement with the project. Why can't we use the same error code that Mario referenced above @Ben-Edwards-cgi ?

chrisbellingham-hmcts: The primary purpose of the DARTS-specific error codes are to inform the caller that a specific failure has occured, such that they can handle it in a specific way. The caller's error-handling code is keyed on the specific DARTS error code.

So it's down to what the API design specifies - if we want the client to be able to disambiguate between an event not found error and (say) a courthouse not found error, then it makes sense to have a specific error code/enum for the event not found scenario. If the client doesn't care, then a generic "not found" is fine. There is no specific pattern for this that I'm aware of, it's use-case/design specific.

However, there is a pattern for how we define our errors in code - If a generic error is suitable, then I suggest it be defined in the existing `src/main/java/uk/gov/hmcts/darts/common/exception/CommonApiError.java` rather than the new inner class within `src/main/java/uk/gov/hmcts/darts/common/exception/DartsApiException.java`. The error code and human-readable message must be defined as enums in the openapi yaml - `src/main/resources/openapi/common.yaml`, and those generated enums used within the enum definition within CommonApiError. The idea is that our error codes and messages are stable and form part of the API contract, therefore are defined in an openapi spec. The convention is the error code (errorTypeNumeric in the error enum) takes the form `FEATURE_XXX`, where XXX is a numerical value unique to the feature. e.g `COMMON_100`, `COMMON_110` etc, and we should stick to that.

I would say if we're going for a generic error then I'd encourage use of the `public DartsApiException(DartsApiError error, String detail)` constructor, and provide some description in the `detail`. The use of a common error code won't allow a client to key their error handling, but the `detail` will at least give devs additional info should we need to troubleshoot the error in the future. E.g:

```
new DartsApiException(DartsApiException.DartsApiErrorCommon.NOT_FOUND, "The provided event id does not correspond to an existing event"));
```

Ben-Edwards-cgi: I did not realise we had a CommonApiError.java :)

I'm happy to either move the existing CommonApiError into the inner class (as these errors are exclusively used within that exception so we don't really need a standard class for it) or to remove them from the inner and stick the items in the common.

I also like the idea of having a changeable detail section as it gives us the ability to use common errors while allowing for more detailed logging but I think it would be good to have a "default" value for when it does not matter

PR #2122, opened 2024-10-10
DMP-4055 Retention Attributes Not Applied to Media, Annotations, Transcriptions, and Case Document

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-4055

### Change description ###

Updated the case document generation to also reset retention data

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/task/service/GenerateCaseDocumentProcessorIntTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/task/service/GenerateCaseDocumentProcessorIntTest.java
@@ -81,10 +81,9 @@ class GenerateCaseDocumentProcessorIntTest extends IntegrationBase {
         List<ExternalObjectDirectoryEntity> eodCaseDocument = eodRepository.findByCaseDocument(caseDocument);
         assertThat(eodCaseDocument.size()).isEqualTo(1);
 
-        caseRepository.findById(courtCase.getId()).ifPresent(c -> {
-            assertThat(c.isRetentionUpdated()).isTrue();
-            assertThat(c.getRetentionRetries()).isEqualTo(0);
-        });
+        CourtCaseEntity courtCaseEntity = caseRepository.findById(caseDocument.getCourtCase().getId()).orElseThrow();
+        assertThat(courtCaseEntity.isRetentionUpdated()).isTrue();
+        assertThat(courtCaseEntity.getRetentionRetries()).isEqualTo(0);
     }
 
 }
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/hmcts/darts/casedocument/service/impl/GenerateCaseDocumentSingleCaseProcessorImplTest.java
```diff
@@ -137,6 +143,8 @@ void testGenerationOfCaseDocument() {
         assertThat(savedCaseDocument.getCreatedBy()).isEqualTo(user);
         assertThat(savedCaseDocument.getLastModifiedBy()).isEqualTo(user);
         assertThat(savedCaseDocument.isHidden()).isEqualTo(false);
+
+        verify(caseRepository).save(courtCaseCaptor.capture());
```
Ben-Edwards-cgi, line 147: Should validate content of the capture to make sure the correct fields are being set on the object

src/integrationTest/java/uk/gov/hmcts/darts/task/service/GenerateCaseDocumentProcessorIntTest.java
```diff
@@ -78,6 +80,11 @@ void testGenerateCaseDocument() {
 
         List<ExternalObjectDirectoryEntity> eodCaseDocument = eodRepository.findByCaseDocument(caseDocument);
         assertThat(eodCaseDocument.size()).isEqualTo(1);
+
+        caseRepository.findById(courtCase.getId()).ifPresent(c -> {
+            assertThat(c.isRetentionUpdated()).isTrue();
+            assertThat(c.getRetentionRetries()).isEqualTo(0);
+        });
```
Ben-Edwards-cgi, line 87: Should add a generic catch incase this id could not be found fail the test

Ben-Edwards-cgi (changes requested):

src/test/java/uk/gov/hmcts/darts/casedocument/service/impl/GenerateCaseDocumentSingleCaseProcessorImplTest.java
karen-hedges: tried but can't as the save method return is mocked so you are basically setting the values and comparing the same which makes it pointless.

src/integrationTest/java/uk/gov/hmcts/darts/task/service/GenerateCaseDocumentProcessorIntTest.java
karen-hedges: technically you have a non null on case so if case document can be found so can case but I can add it just to bomb proof it

hemantasharma1129 (approved):

Ben-Edwards-cgi (approved):

PR #2126, opened 2024-10-10
DMP-3978: Return Batch Size in GET /automated-tasks/{task_id}

sabahirfan:
DMP-3978

### Change description ###
Return Batch Size in GET /automated-tasks/{task_id}

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/openapi/tasks.yaml
+++ b/src/main/resources/openapi/tasks.yaml
@@ -132,13 +132,13 @@ components:
           $ref: '#/components/schemas/CronExpression'
         is_active:
           $ref: '#/components/schemas/IsActive'
-        batchSize:
-          $ref: '#/components/schemas/batchSize'
     DetailedAutomatedTask:
       type: object
       properties:
         is_cron_editable:
           $ref: '#/components/schemas/IsCronEditable'
+        batchSize:
+          $ref: '#/components/schemas/BatchSize'
         created_at:
           $ref: '#/components/schemas/CreatedAt'
         created_by:
@@ -166,7 +166,7 @@ components:
       maxLength: 256
     IsActive:
       type: boolean
-    batchSize:
+    BatchSize:
       type: integer
     IsCronEditable:
       type: boolean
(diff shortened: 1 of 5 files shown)


src/main/resources/openapi/tasks.yaml
```diff
@@ -132,6 +132,8 @@ components:
           $ref: '#/components/schemas/CronExpression'
         is_active:
           $ref: '#/components/schemas/IsActive'
+        batchSize:
```
chrisbellingham-hmcts, line 135: Note the `AutomatedTaskSummary` object is also used by the `GET /admin/automated-tasks` endpoint which returns all automated tasks, whereas DMP-3915 only specifies changing the `GET /admin/automated-tasks/{id}` endpoint, so this change will have broader consequences than what the ticket intends.

Suggest this be moved into the `DetatailedAutomatedTask` object instead, which is only used by the get by id endpoint?

Also the naming convention for API properties is snake_case (see https://hmcts.github.io/restful-api-standards/#118), so the attribute name on L135 should be `batch_size`.

src/main/resources/openapi/tasks.yaml
```diff
@@ -132,6 +132,8 @@ components:
...
+          $ref: '#/components/schemas/batchSize'
```
chrisbellingham-hmcts, line 136: Please capitalise this to BatchSize for consistency with the other component names.

chrisbellingham-hmcts (commented):
Is there any way we can enhance `AdminGetAutomatedTasksByIdTest` to verify that the full (JSON serialised) API response is as expected?

src/main/resources/openapi/tasks.yaml
```diff
@@ -164,6 +166,8 @@ components:
       maxLength: 256
     IsActive:
       type: boolean
+    batchSize:
```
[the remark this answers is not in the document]
sabahirfan: done

src/main/resources/openapi/tasks.yaml
sabahirfan: done

chrisbellingham-hmcts (approved):

jackmaloney (approved):

PR #2132, opened 2024-10-11
DMP-3979 Add Update Batch Size in PATCH /automated-tasks/{task_id}

mario-paniccia:
https://tools.hmcts.net/jira/browse/DMP-3979

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceImpl.java
@@ -19,7 +19,6 @@ import uk.gov.hmcts.darts.tasks.model.DetailedAutomatedTask;
 
 import java.util.List;
 
-import static uk.gov.hmcts.darts.audit.api.AuditActivity.CHANGE_JOB_BATCH_SIZE;
 import static uk.gov.hmcts.darts.audit.api.AuditActivity.ENABLE_DISABLE_JOB;
 import static uk.gov.hmcts.darts.audit.api.AuditActivity.RUN_JOB_MANUALLY;
 import static uk.gov.hmcts.darts.task.exception.AutomatedTaskApiError.AUTOMATED_TASK_ALREADY_RUNNING;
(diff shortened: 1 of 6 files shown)


src/main/resources/openapi/tasks.yaml
```diff
@@ -114,11 +114,11 @@ components:
   schemas:
     AutomatedTaskPatch:
       type: object
-      required:
-        - is_active
       properties:
         is_active:
           $ref: '#/components/schemas/IsActive'
+        batch_size:
+          type: integer
```
mario-paniccia, line 121: author note:
To do things the perfect way we should now validate that at least one of the patch attributes are not null.
The Open Api documentation says that Open Api can enforce that by using `minProperties: 1` but unfortunately it won't work.
I have therefore decided not to enforce this because:

- This is not a public API, and the Portal won't allow sending all `null` patch attributes
- I'd rather avoid overcomplicating the code by adding manual validation logic for this when not really needed
- Even if the endpoint was called with all `null` attributes, the implementation would ignore them and won't change the state of the task
- other PATCH methods in our APIs don't enforce this either

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceImpl.java
```diff
@@ -81,9 +82,15 @@ public void runAutomatedTask(Integer taskId) {
     public DetailedAutomatedTask updateAutomatedTask(Integer taskId, AutomatedTaskPatch automatedTaskPatch) {
         var automatedTask = getAutomatedTaskEntityById(taskId);
 
-        automatedTask.setTaskEnabled(automatedTaskPatch.getIsActive());
+        if (automatedTaskPatch.getIsActive() != null) {
```
Ben-Edwards-cgi, line 85: Should we assume null = false here? (Just so the task is disabled by default?

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceImpl.java
```diff
@@ -81,9 +82,15 @@ public void runAutomatedTask(Integer taskId) {
...
+            automatedTask.setTaskEnabled(automatedTaskPatch.getIsActive());
+            auditApi.record(ENABLE_DISABLE_JOB);
+        }
 
-        auditApi.record(ENABLE_DISABLE_JOB);
+        if (automatedTaskPatch.getBatchSize() != null) {
```
Ben-Edwards-cgi, line 90: if batch size is null should we set the batch size to be null in the automated task?

Ben-Edwards-cgi (approved):

sabahirfan (approved):

Ben-Edwards-cgi (approved):

jackmaloney (approved):

PR #2138, opened 2024-10-14
NO-JIRA: performance fix around loading of eod statuses from DB

mario-paniccia:
### Change description ###
Improving performance by loading EOD statuses once on bean creation rather than every time the task processes an object.
This will reduce the number of SELECTs to the DB

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
@@ -2,6 +2,7 @@ package uk.gov.hmcts.darts.arm.component.impl;
 
 import com.azure.core.util.BinaryData;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import jakarta.annotation.PostConstruct;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections.CollectionUtils;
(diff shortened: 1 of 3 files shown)


Ben-Edwards-cgi (approved):

src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
```diff
@@ -75,10 +76,21 @@ public class ArmResponseFilesProcessSingleElementImpl implements ArmResponseFile
...
+    @PostConstruct
+    public void initialisePreloadedObjects() {
+        storedStatus = objectRecordStatusRepository.findById(STORED.getId()).get();
+        armDropZoneStatus = objectRecordStatusRepository.findById(ARM_DROP_ZONE.getId()).get();
+        armProcessingResponseFilesStatus = objectRecordStatusRepository.findById(ARM_PROCESSING_RESPONSE_FILES.getId()).get();
+        armResponseManifestFailedStatus = objectRecordStatusRepository.findById(ARM_RESPONSE_MANIFEST_FAILED.getId()).get();
+        armResponseProcessingFailedStatus = objectRecordStatusRepository.findById(ARM_RESPONSE_PROCESSING_FAILED.getId()).get();
+        armResponseChecksumVerificationFailedStatus = objectRecordStatusRepository.findById(ARM_RESPONSE_CHECKSUM_VERIFICATION_FAILED.getId()).get();
+
+        userAccount = userIdentity.getUserAccount();
```
chrisbellingham-hmcts, line 88: Initialising the EOD values makes sense as they're basically constants in the sense they're populated by flyway - any changes to that data would drive a new deployment and therefore the application instance would be refreshed, with the value re-read during bean initialisation. Which all seems fine.

Can we say that's the same for the system account however? I.e is there any way that account data can change in the DB during the lifetime of an application instance, meaning the user data read here could get stale?

sabahirfan (approved):

src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
mestebanez: We have a EodHelper for cached Eod info. Could we house this in a similar data object ObjectRecordHelper?

mario-paniccia: > We have a EodHelper for cached Eod info. Could we house this in a similar data object ObjectRecordHelper?

You're totally right Michael, EodHelper would be an even better solution. I've tried using that initially but I was struggling to update the tests to use EodHelper, for some reason they would fails. Since I didn't want to spend much time on this fix, I went for this quicker compromise solution.

mario-paniccia: > Can we say that's the same for the system account however? I.e is there any way that account data can change in the DB during the lifetime of an application instance, meaning the user data read here could get stale?

Yes that's true Chris there is this possibility. Let me revert the change for the user only. Cheers

mestebanez: Just to add it in the mix shouldnt 2nd level db caching solve this for us naturally?

Ben-Edwards-cgi: That is a point could even use spring cache

mario-paniccia: I think we should not rely on DB or caches. If the code is clearly inefficient better remove that inefficiency at the source

Ben-Edwards-cgi: I think this is effectively a custom cache as we are saving into memory on load, by using something like spring cache or what @mestebanez indicated it would allow it to be more responsive to DB changes (As spring can auto refresh on pre-defined intervals or when it detects a change to the root data fresh it)

jackmaloney (approved):

src/main/java/uk/gov/hmcts/darts/arm/component/impl/ArmResponseFilesProcessSingleElementImpl.java
mario-paniccia: About those enums changing value I think Chris has explained perfectly here:
> Initialising the EOD values makes sense as they're basically constants in the sense they're populated by flyway - any changes to that data would drive a new deployment and therefore the application instance would be refreshed, with the value re-read during bean initialisation. Which all seems fine.

PR #2164, opened 2024-10-17
DMP-4104: When the ManualDeletion automated task runs deletedBy and deletedTimestamp should be set

Ben-Edwards-cgi:
### Change description ###
# Summary of Changes

The recent updates to the code primarily enhance the functionality of the `ManualDeletionProcessorImpl` class, allowing it to manage user-specific deletions more effectively. The class now tracks which user is performing deletions, ensuring a clearer audit trail and potentially improving accountability in the deletion process.

## Highlights

- **User Identity Integration**: The code now includes a reference to `UserIdentity`, allowing the system to identify which user is performing deletion actions.
- **User-Specific Deletion**: When media and transcription documents are deleted, the actions now log the user account responsible for the deletion instead of just marking the items as deleted.
- **Testing Enhancements**: The test cases have been updated to verify that the user account is correctly retrieved and used during the deletion process, ensuring that the functionality works as intended.
- **Improved Logging**: The logging of deletion actions now includes user information, which will help in tracking and auditing deletions in the future.

```
[ ] Yes
[X] No
```


src/test/java/uk/gov/hmcts/darts/task/service/impl/ManualDeletionProcessorImplTest.java
```diff
@@ -89,6 +98,10 @@ void processShouldDeleteMediaAndTranscription() {
         verify(unstructuredDeleter).delete(any(ExternalObjectDirectoryEntity.class));
         verify(logApi, times(1)).mediaDeleted(MEDIA_ID);
         verify(logApi, times(1)).transcriptionDeleted(TRANSCRIPTION_ID);
+        verify(userIdentity, times(1)).getUserAccount();
+
+        verify(mediaAction.getMedia(), times(1)).markAsDeleted(userAccount);
```
chrisbellingham-hmcts, line 103: As this test makes heavy use of mocks, it's not totally clear to me that the `deletedBy` and `deletedTimestamp` fields are being successfully updated on the associated records. 

Do you think it's worth testing this more explicitly, or is the `SoftDelete.markAsDeleted()` functionality tested elsewhere to give us confidence?

Ben-Edwards-cgi: that method auto sets the three values but agreed it maybe worth removing the mock for the SoftDelete entity to confirm exact values are set
image

Ben-Edwards-cgi: Test updated to verify raw values instead of method call

chrisbellingham-hmcts (approved):

cmagowan-hmcts (approved):

jackmaloney (approved):

PR #2168, opened 2024-10-18
DMP-4076: Debug logging for investigation of duplicates in media_link…

chrisbellingham-hmcts:
Despite my best efforts I've been unable to recreate the issue described by [DMP-4076](https://tools.hmcts.net/jira/browse/DMP-4076) locally.

The `addCase` method within `MediaLinkedCaseHelper` is the only place where we insert into the `media_linked_case` table.

One theoretical reason for why we could see duplicates (no I couldnt' recreate this either) - it's possible the provided `courtCase` has been modified by the calling code since it was retrieved from the DB, meaning that the object equality comparison fails at the `linkedCases.contains(courtCase)` stage, thereby allowing a new record to be inserted.

This PR adds debug logging which should help confirm/deny that, and otherwise may help should we see this issue again in the future.

This PR is dev-only change.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/helper/MediaLinkedCaseHelper.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/helper/MediaLinkedCaseHelper.java
@@ -1,6 +1,7 @@
 package uk.gov.hmcts.darts.common.helper;
 
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Component;
 import uk.gov.hmcts.darts.common.entity.CourtCaseEntity;
 import uk.gov.hmcts.darts.common.entity.MediaEntity;
@@ -11,6 +12,7 @@ import java.util.List;
 
 @Component
 @RequiredArgsConstructor
+@Slf4j
 public class MediaLinkedCaseHelper {
 
     private final MediaLinkedCaseRepository mediaLinkedCaseRepository;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/darts/common/helper/MediaLinkedCaseHelper.java
```diff
@@ -11,18 +12,30 @@
...
 @RequiredArgsConstructor
+@Slf4j
 public class MediaLinkedCaseHelper {
 
     private final MediaLinkedCaseRepository mediaLinkedCaseRepository;
 
     public void addCase(MediaEntity mediaEntity, CourtCaseEntity courtCase) {
         List<MediaLinkedCaseEntity> mediaLinkedCaseEntities = mediaLinkedCaseRepository.findByMedia(mediaEntity);
         List<CourtCaseEntity> linkedCases = mediaLinkedCaseEntities.stream().map(MediaLinkedCaseEntity::getCourtCase).toList();
+        log.debug("Handling med_id {} and cas_id {}. Currently the media has the following cas_ids linked: {}",
```
Ben-Edwards-cgi, line 23: Can we put this inside a if (log.isDebugEnabled()) statement just so loop over the linkedCase array when we do not need to

chrisbellingham-hmcts: Have now added this as a guard.

Ben-Edwards-cgi (approved):

hemantasharma1129 (approved):

PR #2171, opened 2024-10-18
DMP-4088: Requester receiving approval email for their own requests

Ben-Edwards-cgi:
### Change description ###
# Summary of Changes

The recent updates in the code involve modifications to how user accounts with specific roles are retrieved from courthouses. The changes allow for the exclusion of certain users when fetching this information, enhancing the functionality of the authorisation process.

## Highlights

- **Method Update**: The method `getUsersWithRoleAtCourthouse` now accepts an additional parameter to exclude specific users from the results.
- **Interface Changes**: The interfaces for both `AuthorisationApi` and `AuthorisationService` have been updated to include the new exclusion functionality.
- **Implementation Adjustments**: The implementation classes reflect the changes made in the interfaces, ensuring consistency.
- **Database Query Update**: The repository method for fetching users has been modified to support the exclusion of specified users.
- **Test Cases Updated**: Tests have been adjusted to accommodate the new method signature and verify the correct functionality of the changes.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/repository/UserAccountRepository.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/repository/UserAccountRepository.java
@@ -35,8 +35,9 @@ public interface UserAccountRepository extends
         WHERE courthouse = :courthouse
         AND securityRole.id = :securityRole
         AND userAccount.active = true
+        and userAccount not in :excludingUsers
         """)
-    List<UserAccountEntity> findByRoleAndCourthouse(int securityRole, CourthouseEntity courthouse);
+    List<UserAccountEntity> findByRoleAndCourthouse(int securityRole, CourthouseEntity courthouse, Set<UserAccountEntity> excludingUsers);
 
     @Query("""
         SELECT userAccount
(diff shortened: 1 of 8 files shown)


mestebanez (commented):
Can we update the integration test TranscriptionControllerRequestTranscriptionIntTest to prove the notification api excludes the request user

src/main/java/uk/gov/hmcts/darts/common/repository/UserAccountRepository.java
```diff
@@ -35,8 +35,9 @@ public interface UserAccountRepository extends
         WHERE courthouse = :courthouse
         AND securityRole.id = :securityRole
         AND userAccount.active = true
+        and userAccount not in :excludingUsers
```
mestebanez, line 38: Maybe check against null although maybe we this will work regardless

Ben-Edwards-cgi: I don't think this can ever be null just an empty set which will get ignored

jackmaloney (approved):

hemantasharma1129 (approved):

PR #2177, opened 2024-10-18
DMP-3830 - Add integration test for blob temp file download

mestebanez:

[[N/A](https://tools.hmcts.net/jira/browse/DMP-3830)](https://tools.hmcts.net/jira/browse/DMP-3830)

### Change description ###

An extension to the PR https://github.com/hmcts/darts-api/pull/2170. This PR includes an integration test for blob download and temp file removal

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/DataManagementServiceStubImpl.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/DataManagementServiceStubImpl.java
@@ -39,7 +39,7 @@ import static uk.gov.hmcts.darts.test.common.TestUtils.getFile;
 @Component
 @Slf4j
 @RequiredArgsConstructor
-@Profile("intTest")
+@Profile("intTest && !blobTest")
 @Deprecated
 public class DataManagementServiceStubImpl implements DataManagementService {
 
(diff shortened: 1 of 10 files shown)


src/integrationTest/java/uk/gov/hmcts/darts/testutils/stubs/DataManagementServiceStubImpl.java
```diff
@@ -39,7 +39,7 @@
 @Component
 @Slf4j
 @RequiredArgsConstructor
-@Profile("intTest")
+@Profile("intTest && !blobTest")
```
mestebanez, line 42: We need a better way of excluding/including beans when test code is run. Production code should not know about test profiles. Maybe this can be address through a test specific configuration class in future.

src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmServiceImpl.java
```diff
@@ -37,7 +36,6 @@
 
 @Service
 @Slf4j
-@Profile("!intTest")
```
mestebanez, line 40: Remove the need for code to know about the test profile

src/main/java/uk/gov/hmcts/darts/datamanagement/service/impl/DataManagementServiceImpl.java
```diff
@@ -42,7 +41,6 @@
 
 @Service
 @Slf4j
-@Profile("!intTest")
```
mestebanez, line 45: Remove the need for code to know about the test profile

cmagowan-hmcts (approved):

PR #2193, opened 2024-10-24
DMP-4113: Use uppercase for Courthouse and Courtroom in logs

Ben-Edwards-cgi:
### Change description ###
# Summary of Git Diff

This Git Diff represents a series of code changes made to various files in a software project. The primary focus of these changes is the introduction of a new utility class called `DataUtil` that helps standardize the processing of data by ensuring that certain fields (like courthouse and courtroom names) are always converted to uppercase. This utility is utilized in multiple controllers and services to enhance the consistency and reliability of the data being handled.

## Highlights

- **New Utility Class**: A new class named `DataUtil` is introduced to provide utility functions for common data processing tasks, such as converting strings to uppercase and preprocessing event and request data.
  
- **Standardized Data Processing**: Several existing methods in various controllers (like `AudioController`, `CaseController`, and `EventsController`) have been updated to use `DataUtil` for preprocessing incoming data, ensuring uniformity across the application.

- **Improved Logging**: The logging statements in services like `AudioLoggerServiceImpl` and `CasesLoggerServiceImpl` have been modified to use the uppercase conversion from `DataUtil`, which helps maintain consistency in log entries.

- **Code Cleanup**: The changes also include removing redundant code and improving readability by centralizing the data processing logic within the new utility class.

- **Unit Tests Added**: New tests are provided for the `DataUtil` class to ensure that the new functionality works as expected, including tests for handling null values gracefully.

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/darts/event/service/handler/DarStartHandler.java
```diff
@@ -33,6 +34,7 @@ public DarStartHandler(RetrieveCoreObjectService retrieveCoreObjectService,
     @Override
     @Transactional
     public void handle(DartsEvent dartsEvent, EventHandlerEntity eventHandler) {
+        DataUtil.preProcess(dartsEvent);
```
chrisbellingham-hmcts, line 37: Do we only wish to preProcess specific "types" of event? 

If not, could we preProcess within `EventDispatcherImpl.receive()`, which is closer to the point of entry of events into the system? That would avoid the need to preProcess in each individual handler.

src/main/java/uk/gov/hmcts/darts/log/service/impl/AudioLoggerServiceImpl.java
```diff
@@ -14,8 +15,8 @@ public class AudioLoggerServiceImpl implements AudioLoggerService {
     @Override
     public void audioUploaded(AddAudioMetadataRequest request) {
         log.info("Audio uploaded: courthouse={}, courtroom={}, started_at={}, ended_at={}",
-                 request.getCourthouse(),
-                 request.getCourtroom(),
+                 DataUtil.toUpperCase(request.getCourthouse()),
+                 DataUtil.toUpperCase(request.getCourtroom()),
```
chrisbellingham-hmcts, line 19: I think the `AddAudioMetadataRequest` has already been preProcessed at the controller level, so do we need this transformation here? Belt and braces?

src/main/java/uk/gov/hmcts/darts/cases/controller/CaseController.java
```diff
@@ -70,8 +70,8 @@ public ResponseEntity<List<ScheduledCase>> casesGet(
         LocalDate date
     ) {
         GetCasesRequest request = new GetCasesRequest();
-        request.setCourthouse(courthouse.toUpperCase(Locale.ROOT));
-        request.setCourtroom(courtroom.toUpperCase(Locale.ROOT));
+        request.setCourthouse(DataUtil.toUpperCase(courthouse));
+        request.setCourtroom(DataUtil.toUpperCase(courtroom));
```
chrisbellingham-hmcts, line 74: Any reason not to use the same pattern as you've used in `AudioController` and in `casesPost`? I.e create a static void method in DataUtil that preProcesses?

src/main/java/uk/gov/hmcts/darts/event/service/handler/DarStartHandler.java
Ben-Edwards-cgi: Good spot,

I think we already preProcess on both the Dispatcher service and the controller so I removed the redundant checks :)

src/main/java/uk/gov/hmcts/darts/log/service/impl/AudioLoggerServiceImpl.java
Ben-Edwards-cgi: This is creating a brand new object so we create it here with the correct values instead of consuming an already created object

src/main/java/uk/gov/hmcts/darts/cases/controller/CaseController.java
Ben-Edwards-cgi: This is creating a brand new object so we create it here with the correct values instead of consuming an already created object

src/main/java/uk/gov/hmcts/darts/log/service/impl/AudioLoggerServiceImpl.java
Ben-Edwards-cgi: Agreed :)

I left these here incase we re-use this logging anywhere that is not already covered by a preProcess method in the future as the format is important for DynaTrace it seemed sensible to leave it here too :)

As such I left these here for Belt and braces

chrisbellingham-hmcts (approved):

jackmaloney (approved):

PR #2204, opened 2024-10-28
DMP-4016: Anonymise text from all versions of event on case expiry

mario-paniccia:
https://tools.hmcts.net/jira/browse/DMP-4016

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImpl.java
```diff
@@ -62,6 +62,7 @@ public void anonymizeCourtCaseEntity(UserAccountEntity userAccount, CourtCaseEnt
         courtCase.getDefenceList().forEach(defenceEntity -> anonymizeDefenceEntity(userAccount, defenceEntity));
         courtCase.getProsecutorList().forEach(prosecutorEntity -> anonymizeProsecutorEntity(userAccount, prosecutorEntity));
         courtCase.getHearings().forEach(hearingEntity -> anonymizeHearingEntity(userAccount, hearingEntity));
+        eventService.getAllCourtCaseEventVersions(courtCase).forEach(eventEntity -> anonymizeEventEntity(userAccount, eventEntity));
```
Ben-Edwards-cgi, line 65: @jackmaloney  should this only take into account event versions where all the associated cases are anonymised or due to be? or is this okay just anonymising  them all regardless?

Ben-Edwards-cgi, line 65: Optional: but maybe worth to pull this out into its own method i.e

anonymizeAllEventsFromCase(...) just so we can re-use it if we need to in the future

Ben-Edwards-cgi, line 65: You may need to manually save these events as they may not all get caught up in the saveCase method

jackmaloney: Yeh, it's on my todo list to raise this as a separate ticket to cover this requirement. For now, we just anonymise everything.

mario-paniccia: no need, this service is marked as `@Transactional`. The events have been retrieved with a call to a JPA repo so they are managed by JPA which will flush changes to DB once the method completes

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
```diff
@@ -34,4 +41,19 @@ public EventEntity getEventByEveId(Integer eveId) {
     public EventEntity saveEvent(EventEntity eventEntity) {
         return eventRepository.save(eventEntity);
     }
+
+    @Override
+    public List<EventEntity> getAllCourtCaseEventVersions(CourtCaseEntity courtCase) {
+        List<EventEntity> allEvents = new ArrayList<>();
```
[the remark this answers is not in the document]
Ben-Edwards-cgi: We select the events from two separate places this is just to ensure that if the same event is in both locations we we ignore the second one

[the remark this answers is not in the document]
jackmaloney: Yes, you can get duplicates but these should be cleaned up every evening as part of the [Duplicated event cleanup task ](https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=1689786831#TaskautomationLLD-Removeduplicatedeventstask)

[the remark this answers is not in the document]
jackmaloney: For the example above, it is highly unlikely you would get duplicates because it would need to be the same event that was received once into legacy and then once into modernised. This is an extremely unlikely edge case. 
However, there is probably no harm in making this a set as suggested, just in case @mario-paniccia ?

Ben-Edwards-cgi (approved):

src/main/java/uk/gov/hmcts/darts/event/service/impl/EventServiceImpl.java
```diff
@@ -34,4 +42,19 @@ public EventEntity getEventByEveId(Integer eveId) {
     public EventEntity saveEvent(EventEntity eventEntity) {
         return eventRepository.save(eventEntity);
     }
+
+    @Override
+    public Set<EventEntity> getAllCourtCaseEventVersions(CourtCaseEntity courtCase) {
+        Set<EventEntity> allEvents = new HashSet<>();
+        List<EventEntity> eventsFromCaseId = eventLinkedCaseRepository
```
jackmaloney, line 49: Maybe worth adding a comment to this method explaining that the way this table is populated differs between legacy and modernised, hence the reason for 2 different queries to the eventLinkedCaseRepository.

davet1985: Added

davet1985 (approved):

PR #2205, opened 2024-10-28
DMP-3451: Handle migrated Daily Lists

Ben-Edwards-cgi:
### Change description ###
Daily Lists that are migrated will have the XML and JSON fields defaulted to null which means modernised DARTS won't recognise the Daily List and it won't be processed. The daily list XML will be placed in the DETS datastore and the external_location and elt_id fields used to stored the name and datastore of the blob.

It is only anticipated that the elt_id would be set to DETS (4).

*Scenario - No XML or JSON*

Both XML and JSON fields are null. The elt_id and external_location fields are not null.

The daily_list XML content is stored in the location specified by the elt_id (should be DETS) with the blob name specified by the external_location field.

Fetch the XML from the datastore, saving it in the daily_list table. Copy the XML to JSON mapping from the darts-gateway and generate JSON content. Save the JSON content into the daily_list table and proceed as normal with processing the daily list. Then proceed as normal with processing the daily list.

It's been decided to copy the mapping from the darts-gateway and maintain it in two places. This isn't an ideal solution, but it is temporary and once migrated daily lists have been processed, within a few weeks of go live, the duplicated mapping can be removed from the darts-api codebase.

*Lifecycle*

Once the task has identified which daily lists need to be processed using the already implemented logic then the new logic to carry out the fetching of XML from DETS and the XML to JSON mapping should be carried out.

We should not be fetching and translating XML unless it's required for a daily list to be processed.

*Log errors*

We need to log the following errors, and others if appropriate.
 * the daily_list external_location or elt_id is null
 * there is an issue fetching the blob from DETS
 * there is an issue translating to JSON

Ensure the dal_id is included within the log message.

h4.Acceptance Criteria
{*}AC1 - No XML or JSON content{*}

GIVEN a migrated daily list exists
AND the XML field is null
AND the JSON content field is null
WHEN the daily list processing runs
AND the daily list entry is required to be processed
THEN the XML content is fetched from DETS and translated into JSON
AND the JSON content is saved on the daily list
AND the daily list status is set to PROCESSED

*AC2 - Error translating XML to JSON*

GIVEN a migrated daily list exists
AND the JSON content field is null
WHEN the daily list processing runs
AND the daily list entry is required to be processed
AND the XML content is not able to be translating into JSON
THEN an error log can be seen including the dal_id
AND the daily list status is set to FAILED

*AC3 - Error fetching from DETS*

GIVEN a migrated daily list exists
AND the XML field is null
AND the JSON content field is null
WHEN the daily list processing runs
AND the daily list entry is required to be processed
AND the XML content is not able to be fetched from DETS
THEN an error log can be seen including the dal_id
AND the daily list status is set to FAILED

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListUpdater.java
```diff
@@ -60,70 +78,127 @@ class DailyListUpdater {
...
-                List<Sitting> sittings = courtList.getSittings();
-                for (Sitting sitting : sittings) {
-                    List<Hearing> hearings = sitting.getHearings();
-                    for (Hearing dailyListHearing : hearings) {
-
-                        String caseNumber = getCaseNumber(dailyListEntity, dailyListHearing);
-                        if (caseNumber == null) {
-                            statusType = JobStatusType.PARTIALLY_PROCESSED;
-                            continue;
+        JobStatusType statusType = validateJsonExistsElseUpdate(dailyListEntity) ? JobStatusType.PROCESSED : JobStatusType.FAILED;
```
jackmaloney, line 91: I've created a ticket to track reverting this, mind adding a reference to it as a comment above this line so it is clear this is a temporary change? DMP-4191

jackmaloney (approved):

src/main/java/uk/gov/hmcts/darts/dailylist/service/impl/DailyListUpdater.java
Ben-Edwards-cgi: Updated

davet1985 (approved):

PR #2215, opened 2024-10-29
Dmp 3986.media linked case enhance

mario-paniccia:
https://tools.hmcts.net/jira/browse/DMP-3986

```
[ ] Yes
[x ] No
```


cmagowan-hmcts:
Looks fine, but would feel better if a integration test had a check in it for the new columns being populated correctly

src/main/java/uk/gov/hmcts/darts/common/enums/MediaLinkedCaseSourceType.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@Getter
+@RequiredArgsConstructor
+public enum MediaLinkedCaseSourceType {
+    LEGACY(0),
+    ADD_AUDIO_METADATA(1),
+    ADD_AUDIO_EVENT_LINKING(2),
+    AUDIO_LINKING_TASK(3);
```
Ben-Edwards-cgi, line 12: Should update integration tests to verify this gets added to the source on DB

src/main/resources/db/migration/common/V1_401__add_user_for_automated_task.sql
```diff
@@ -0,0 +1 @@
+INSERT INTO user_account VALUES (-5, NULL, 'AudioLinkingAutomatedTask', NULL, 'AudioLinkingAutomatedTask', '2024-01-01 00:00:00+00', '2024-01-01 00:00:00+00', NULL, 0, 0, NULL, true, true, 'AudioLinkingAutomatedTask');
```
Ben-Edwards-cgi, line 1: Do we need a new user could we not use the standard 0 one?

cmagowan-hmcts: the idea is that each AutomatedTask has its own user account, in this way when a modification is made to a row in the database, you know what task did it. ATM, when we go live, 2nd line support will not know which of our 30+ tasks changed the row.

mario-paniccia: @cmagowan-hmcts suggested using a different user for each automated task to be able to distinguish the tasks who made DB changes

davet1985: Yeah @cmagowan-hmcts and I discussed this and it's a good suggestion and this is the starting point.

src/main/java/uk/gov/hmcts/darts/common/enums/MediaLinkedCaseSourceType.java
mario-paniccia: 0 is a value that will be set on DB by migration, Darts API never stores a value `0`

cmagowan-hmcts (approved):

src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/AudioControllerAddAudioMetadataMediaLinkIntTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+        List<HearingEntity> addAudioLinkedHearings = new ArrayList<>();
+        for (HearingEntity hearing : allHearings) {
+            if (hearing.getCourtCase().getCaseNumber().contains("case1")) {
+                addAudioLinkedHearings.add(hearing);
+            }
+        }
+        assertEquals(1, addAudioLinkedHearings.size());
+
+        for (HearingEntity hearing : addAudioLinkedHearings) {
```
Ben-Edwards-cgi, line 128: As size is always one you do not need to loop here

src/main/resources/db/migration/common/V1_403__add_user_for_automated_task.sql
```diff
@@ -0,0 +1 @@
+INSERT INTO user_account VALUES (-5, NULL, 'AudioLinkingAutomatedTask', NULL, 'AudioLinkingAutomatedTask', '2024-01-01 00:00:00+00', '2024-01-01 00:00:00+00', NULL, 0, 0, NULL, true, true, 'AudioLinkingAutomatedTask');
```
jackmaloney, line 1: This can be removed after Ben's change goes in https://github.com/hmcts/darts-api/pull/2226

davet1985: Removed

jackmaloney (approved):

PR #2226, opened 2024-11-01
DMP-4225: Flyway script to run automated tasks as different users

Ben-Edwards-cgi:
### Change description ###
In order to more easily identify the automated task that changed data using the "created_by" and "last_modified_by" columns we would like each automated task to run as a different user.

Create a flyway script to add the required number of users, giving hard-coded negative IDs.

```
[ ] Yes
[X] No
```


jackmaloney (approved):

src/main/resources/db/migration/common/V1_403__automated_task_users.sql
```diff
@@ -0,0 +1,94 @@
+INSERT INTO user_account
+VALUES (-6, NULL, 'System user for ProcessDailyList Job', NULL,
+        'systemProcessDailyListAutomatedTask', '2024-01-01 00:00:00+00', '2024-01-01 00:00:00+00', NULL, 0, 0, NULL, true,
```
mestebanez, line 3: Dont know why, but see 31 tasks but 28 in the code base

src/main/resources/db/migration/common/V1_403__automated_task_users.sql
```diff
@@ -0,0 +1,94 @@
...
+        true, 'systemProcessDailyListAutomatedTask'),
+       (-7, NULL, '', NULL,
+        'systemCloseOldUnfinishedTranscriptionsAutomatedTask', '2024-01-01 00:00:00+00', '2024-01-01 00:00:00+00', NULL, 0, 0, NULL, true,
```
davet1985, line 6: Any specific reason to prefix them all with `system`? Seems overkill as they are set as "system" users as well...

src/main/resources/db/migration/common/V1_403__automated_task_users.sql
jackmaloney: There were 3 new ones added to Flyway last week, so this is expected.

src/main/resources/db/migration/common/V1_403__automated_task_users.sql
jackmaloney: This was to follow the pattern with the other system users already in the database.

davet1985 (approved):

PR #2230, opened 2024-11-05
DMP-4133 (Dev Only) ARM RPO - Integrate getIndexesByMatterId

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-4133

### Change description ###

Added functionality to call the ARM RPO endpoint getIndexesByMatterId

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiImpl.java
@@ -107,7 +107,8 @@ public class ArmRpoApiImpl implements ArmRpoApi {
 
         if (isNull(indexesByMatterIdResponse)
             || CollectionUtils.isEmpty(indexesByMatterIdResponse.getIndexes())
-            || isNull(indexesByMatterIdResponse.getIndexes().getFirst())) {
+            || isNull(indexesByMatterIdResponse.getIndexes().getFirst())
+            || isNull(indexesByMatterIdResponse.getIndexes().getFirst().getIndex())) {
             throw handleFailureAndCreateException(errorMessage.append("Unable to find indexes by matter ID in response").toString(),
                                                   armRpoExecutionDetailEntity,
                                                   userAccount);
--- a/src/main/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiImpl.java
@@ -114,7 +114,7 @@ public class ArmRpoApiImpl implements ArmRpoApi {
                                                   userAccount);
         }
         if (indexesByMatterIdResponse.getIndexes().size() > 1) {
-            log.warn("More than one index found in response for matterId: {}", matterId);
+            log.warn("More than one index found in response for matterId: {} - response {}", matterId, indexesByMatterIdResponse);
         }
         armRpoExecutionDetailEntity.setIndexId(indexesByMatterIdResponse.getIndexes().getFirst().getIndex().getIndexId());
         armRpoService.updateArmRpoStatus(armRpoExecutionDetailEntity, ArmRpoHelper.completedRpoStatus(), userAccount);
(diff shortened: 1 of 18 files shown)


src/main/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiImpl.java
```diff
@@ -59,8 +63,37 @@ public void getRecordManagementMatter(String bearerToken, Integer executionId, U
...
+            indexesByMatterIdResponse = armRpoClient.getIndexesByMatterId(bearerToken, createIndexesByMatterIdRequest(matterId));
+        } catch (FeignException e) {
+            // this ensures the full error body containing the ARM error detail is logged rather than a truncated version
+            log.error(errorMessage.append("Unable to get ARM RPO response") + " {}", e.contentUTF8());
+            throw handleFailureAndCreateException(errorMessage.toString(), armRpoExecutionDetailEntity, userAccount);
+        }
+
+        if (isNull(indexesByMatterIdResponse)
+            || CollectionUtils.isEmpty(indexesByMatterIdResponse.getIndexes())
+            || isNull(indexesByMatterIdResponse.getIndexes().getFirst())) {
```
chrisbellingham-hmcts, line 83: Can we be sure ARM will always return a list with at most one item?

src/main/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiImpl.java
```diff
@@ -59,8 +63,37 @@ public void getRecordManagementMatter(String bearerToken, Integer executionId, U
...
+            throw handleFailureAndCreateException(errorMessage.append("Unable to find indexes by matter ID in response").toString(),
+                                                  armRpoExecutionDetailEntity,
+                                                  userAccount);
+        }
+
+        armRpoExecutionDetailEntity.setIndexId(indexesByMatterIdResponse.getIndexes().getFirst().getIndex().getIndexId());
```
chrisbellingham-hmcts, line 89: Worth checking if `index` is null, to guard against NPE when invoking `getIndexId()`?

karen-hedges: done in line 82

chrisbellingham-hmcts: I mean if we get a response like this:

```json
{
  "indexes": [
    {
      "index": null,
      "isMultiStream": false,
      "children": []
    }
  ]
}
```

The check on L82 will pass as `indexes` is not null or empty, but if the nested `index` is null (or omited from the response) then you'll get NPE when you try to get the `indexId` attribute of the null `index` object.

src/main/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiImpl.java
chrisbellingham-hmcts: Yep not worried about the null/empty check, but I'm wondering can we ever get, say, 2 items in this list? If that's possible, would it always be the first item we want to get? One for @hemantasharma1129 to clarify perhaps?

Even if we only ever expect one value here, it might be worth logging a warning if size > 1, just so we capture it.

hemantasharma1129 (approved):

src/main/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiImpl.java
hemantasharma1129: I agree with @chrisbellingham-hmcts. ARM has confirmed it will be alway 1 object in the collection but its worth adding a warning if its more than 2.

src/main/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiImpl.java
```diff
@@ -89,8 +92,42 @@ public void getRecordManagementMatter(String bearerToken, Integer executionId, U
...
+            indexesByMatterIdResponse = armRpoClient.getIndexesByMatterId(bearerToken, createIndexesByMatterIdRequest(matterId));
+        } catch (FeignException e) {
+            // this ensures the full error body containing the ARM error detail is logged rather than a truncated version
+            log.error(errorMessage.append("Unable to get ARM RPO response") + " {}", e.contentUTF8());
+            throw handleFailureAndCreateException(errorMessage.toString(), armRpoExecutionDetailEntity, userAccount);
+        }
+
+        handleResponseStatus(userAccount, indexesByMatterIdResponse, errorMessage, armRpoExecutionDetailEntity);
+
+        if (isNull(indexesByMatterIdResponse)
```
chrisbellingham-hmcts, line 112: Is this null check already performed by `handleResponseStatus()`? If so you can omit it here.

chrisbellingham-hmcts (approved):

PR #2235, opened 2024-11-07
DMP-4267 Find courthouse by display name

jackmaloney:
https://tools.hmcts.net/jira/browse/DMP-4267

### Change description ###
This PR fixes an issue whereby when a user searches for a case using a courthouse and courtroom where the courthouse ID does not match the courthouse display name, then the case is not found. 

```
[ ] Yes
[x ] No
```


karen-hedges (approved):

davet1985 (approved):
LGTM

PR #2238, opened 2024-11-07
DMP-4216: Allow inactive tasks to be run manually

Ben-Edwards-cgi:
### Change description ###
# Summary of Changes

Updated manual task execution to bypass enabled / disabled checks. Meaning the task will be executed when run manually even if the task is disabled in the database.

## Highlights

- **New Method for Running Tasks**: 
  - A new method `run(boolean isManualRun)` was added to the `AutomatedTask` interface, allowing tasks to indicate whether they are running manually or automatically.
  - A default method `run()` was also added, which defaults to calling the new method with `false` (indicating an automated run).

- **Modifications in Implementations**:
  - The `AbstractLockableAutomatedTask` class was updated to implement the new `run(boolean isManualRun)` method.
  - The `AutomatedTaskRunner` class now takes into account whether a task is being run manually or automatically, logging the type of execution.

- **Service Layer Updates**:
  - The `AdminAutomatedTasksServiceImpl` class now triggers the automated task with the new manual run parameter set to `true` when tasks are run manually.

- **Test Adjustments**:
  - Corresponding tests were updated to match the new method signatures and to ensure that the manual run functionality is tested appropriately. 

These changes collectively improve the control and monitoring of automated tasks, making it easier for developers to distinguish between different execution modes.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/task/runner/impl/AbstractLockableAutomatedTask.java
+++ b/src/main/java/uk/gov/hmcts/darts/task/runner/impl/AbstractLockableAutomatedTask.java
@@ -79,7 +79,7 @@ public abstract class AbstractLockableAutomatedTask implements AutomatedTask, Au
     }
 
     @Override
-    public void run() {
+    public void run(boolean isManualRun) {
         executionId = ThreadLocal.withInitial(UUID::randomUUID);
         preRunTask();
         try {
--- a/src/test/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceImplTest.java
+++ b/src/test/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceImplTest.java
@@ -71,8 +71,8 @@ class AdminAutomatedTasksServiceImplTest {
 
         adminAutomatedTaskService.runAutomatedTask(1);
 
-        verify(automatedTaskRunner, times(1)).run(someAutomatedTask);
-        verify(auditApi, times(1)).record(AuditActivity.RUN_JOB_MANUALLY,"some-task-name");
+        verify(automatedTaskRunner, times(1)).run(someAutomatedTask, true);
+        verify(auditApi, times(1)).record(AuditActivity.RUN_JOB_MANUALLY, "some-task-name");
     }
 
     @Test
@@ -95,7 +95,7 @@ class AdminAutomatedTasksServiceImplTest {
         assertFalse(automatedTaskEntity.getTaskEnabled());
         assertEquals(100, automatedTaskEntity.getBatchSize());
         assertEquals(expectedReturnTask, task);
-        verify(auditApi).record(AuditActivity.ENABLE_DISABLE_JOB,"some-task-name disabled");
+        verify(auditApi).record(AuditActivity.ENABLE_DISABLE_JOB, "some-task-name disabled");
         verifyNoMoreInteractions(auditApi);
     }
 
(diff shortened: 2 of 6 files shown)


src/test/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceImplTest.java
```diff
@@ -71,8 +71,8 @@ void invokesTaskWhenTaskIsNotLocked() {
 
         adminAutomatedTaskService.runAutomatedTask(1);
 
-        verify(automatedTaskRunner, times(1)).run(someAutomatedTask);
-        verify(auditApi, times(1)).record(AuditActivity.RUN_JOB_MANUALLY,"some-task-name");
+        verify(automatedTaskRunner, times(1)).run(someAutomatedTask, true);
```
karen-hedges, line 74: You have set isManual to true but is there a test where its set to false

davet1985 (changes requested):
Can you please add an info log as described in the ticket

```
Add an info log entry to state that an inactive task has been run, such as "Task: {} is inactive but has been run manually"
```

src/test/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceImplTest.java
Ben-Edwards-cgi: the is manual run false should be covered by the existing tests for automated tasks. As when it is not manually run we default this value to be false

davet1985 (approved):

src/main/java/uk/gov/hmcts/darts/task/runner/impl/AbstractLockableAutomatedTask.java
```diff
@@ -79,7 +79,7 @@ private void setupUserAuthentication() {
     }
 
     @Override
-    public void run() {
+    public void run(boolean isManualRun) {
```
jackmaloney, line 82: This code is making my head hurt. Why are we passing this `isManualRun` in as a parameter? Can't we just use the `isManualTask` variable that is already used in this class?

src/main/java/uk/gov/hmcts/darts/task/runner/impl/AbstractLockableAutomatedTask.java
```diff
@@ -90,7 +90,10 @@ public void run() {
                 String dbCronExpression = automatedTask.getCronExpression();
                 // Check the cron expression hasn't been changed in the database by another instance, if so skip this run
                 if (isManualTask || getLastCronExpression().equals(dbCronExpression)) {
-                    if (TRUE.equals(automatedTask.getTaskEnabled())) {
+                    if (isManualRun || TRUE.equals(automatedTask.getTaskEnabled())) {
+                        if (isManualRun && !TRUE.equals(automatedTask.getTaskEnabled())) {
```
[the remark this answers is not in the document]
Ben-Edwards-cgi: Agreed removed :)

src/main/java/uk/gov/hmcts/darts/task/runner/impl/AbstractLockableAutomatedTask.java
Ben-Edwards-cgi: I have removed the ManualTaskService as this no longer provided any value and updated the code to reflect this change

jackmaloney (approved):

PR #2270, opened 2024-11-14
DMP-4143: Implement ProcessE2EARMRPOPending automated task

chrisbellingham-hmcts:
# [DMP-4143](https://tools.hmcts.net/jira/browse/DMP-4143)

Implement ProcessE2EARMRPOPending automated task.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiImpl.java
@@ -286,7 +286,7 @@ public class ArmRpoApiImpl implements ArmRpoApi {
     }
 
     @Override
-    public void addAsyncSearch(String bearerToken, Integer executionId, UserAccountEntity userAccount) {
+    public String addAsyncSearch(String bearerToken, Integer executionId, UserAccountEntity userAccount) {
 
         final ArmRpoExecutionDetailEntity executionDetail = armRpoService.getArmRpoExecutionDetailEntity(executionId);
         armRpoService.updateArmRpoStateAndStatus(executionDetail,
@@ -336,6 +336,8 @@ public class ArmRpoApiImpl implements ArmRpoApi {
 
         executionDetail.setSearchId(searchId);
         armRpoService.updateArmRpoStatus(executionDetail, ArmRpoHelper.completedRpoStatus(), userAccount);
+
+        return searchName;
     }
 
 
--- a/src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmRpoServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmRpoServiceImpl.java
@@ -1,8 +1,10 @@
 package uk.gov.hmcts.darts.arm.service.impl;
 
+import jakarta.persistence.EntityManager;
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
 import uk.gov.hmcts.darts.arm.service.ArmRpoService;
 import uk.gov.hmcts.darts.common.entity.ArmRpoExecutionDetailEntity;
 import uk.gov.hmcts.darts.common.entity.ArmRpoStateEntity;
@@ -20,6 +22,19 @@ public class ArmRpoServiceImpl implements ArmRpoService {
 
     public static final String ARM_RPO_EXECUTION_DETAIL_NOT_FOUND = "ArmRpoExecutionDetail not found";
     private final ArmRpoExecutionDetailRepository armRpoExecutionDetailRepository;
+    private final EntityManager entityManager;
+
+    @Override
+    @Transactional
+    public ArmRpoExecutionDetailEntity createArmRpoExecutionDetailEntity(UserAccountEntity userAccount) {
+        ArmRpoExecutionDetailEntity armRpoExecutionDetailEntity = new ArmRpoExecutionDetailEntity();
+
+        UserAccountEntity mergedUserAccountEntity = entityManager.merge(userAccount);
+        armRpoExecutionDetailEntity.setCreatedBy(mergedUserAccountEntity);
+        armRpoExecutionDetailEntity.setLastModifiedBy(mergedUserAccountEntity);
+
+        return saveArmRpoExecutionDetailEntity(armRpoExecutionDetailEntity);
+    }
 
     @Override
     public ArmRpoExecutionDetailEntity getArmRpoExecutionDetailEntity(Integer executionId) {
@@ -53,4 +68,5 @@ public class ArmRpoServiceImpl implements ArmRpoService {
     public ArmRpoExecutionDetailEntity saveArmRpoExecutionDetailEntity(ArmRpoExecutionDetailEntity armRpoExecutionDetailEntity) {
         return armRpoExecutionDetailRepository.save(armRpoExecutionDetailEntity);
     }
+
 }
--- a/src/main/java/uk/gov/hmcts/darts/arm/service/impl/TriggerArmRpoSearchServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/arm/service/impl/TriggerArmRpoSearchServiceImpl.java
@@ -1,0 +1,79 @@
+package uk.gov.hmcts.darts.arm.service.impl;
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+import uk.gov.hmcts.darts.arm.helper.ArmRpoHelper;
+import uk.gov.hmcts.darts.arm.rpo.ArmRpoApi;
+import uk.gov.hmcts.darts.arm.service.ArmApiService;
+import uk.gov.hmcts.darts.arm.service.ArmRpoService;
+import uk.gov.hmcts.darts.authorisation.component.UserIdentity;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class TriggerArmRpoSearchServiceImpl implements uk.gov.hmcts.darts.arm.service.TriggerArmRpoSearchService {
+
+    private final ArmRpoApi armRpoApi;
+    private final ArmRpoService armRpoService;
+    private final ArmApiService armApiService;
+    private final UserIdentity userIdentity;
+
+    /**
+     * This method integrates various ARM RPO API calls to ultimately trigger a search. The results of that search are then processed by another automated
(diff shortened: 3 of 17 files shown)


src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmRpoServiceImpl.java
```diff
@@ -20,6 +22,19 @@ public class ArmRpoServiceImpl implements ArmRpoService {
...
     public static final String ARM_RPO_EXECUTION_DETAIL_NOT_FOUND = "ArmRpoExecutionDetail not found";
     private final ArmRpoExecutionDetailRepository armRpoExecutionDetailRepository;
+    private final EntityManager entityManager;
+
+    @Override
+    @Transactional
+    public ArmRpoExecutionDetailEntity createArmRpoExecutionDetailEntity(UserAccountEntity userAccount) {
+        ArmRpoExecutionDetailEntity armRpoExecutionDetailEntity = new ArmRpoExecutionDetailEntity();
+
+        UserAccountEntity mergedUserAccountEntity = entityManager.merge(userAccount);
```
Ben-Edwards-cgi, line 32: Is this needed?

src/main/java/uk/gov/hmcts/darts/arm/service/impl/TriggerArmRpoSearchServiceImpl.java
```diff
@@ -0,0 +1,79 @@
...
+import uk.gov.hmcts.darts.arm.helper.ArmRpoHelper;
+import uk.gov.hmcts.darts.arm.rpo.ArmRpoApi;
+import uk.gov.hmcts.darts.arm.service.ArmApiService;
+import uk.gov.hmcts.darts.arm.service.ArmRpoService;
+import uk.gov.hmcts.darts.authorisation.component.UserIdentity;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class TriggerArmRpoSearchServiceImpl implements uk.gov.hmcts.darts.arm.service.TriggerArmRpoSearchService {
```
Ben-Edwards-cgi, line 15: Does not need to be fully qualified

src/main/java/uk/gov/hmcts/darts/arm/service/impl/TriggerArmRpoSearchServiceImpl.java
```diff
@@ -0,0 +1,79 @@
...
+                                                         userAccountEntity);
+
+        String searchName = armRpoApi.addAsyncSearch(armBearerToken,
+                                                     executionId,
+                                                     userAccountEntity);
+
+        armRpoApi.saveBackgroundSearch(armBearerToken,
+                                       executionId,
+                                       searchName,
+                                       userAccountEntity);
```
Ben-Edwards-cgi, line 74: As these do not seem to depend on each other could they be made to run in parallel?

src/main/java/uk/gov/hmcts/darts/task/runner/impl/ProcessE2EArmRpoPendingAutomatedTask.java
```diff
@@ -0,0 +1,43 @@
...
+    havingValue = "true"
+)
+@Slf4j
+@Component
+public class ProcessE2EArmRpoPendingAutomatedTask extends AbstractLockableAutomatedTask implements AutoloadingManualTask {
+
+    private final TriggerArmRpoSearchService triggerArmRpoSearchService;
+
+    protected ProcessE2EArmRpoPendingAutomatedTask(AutomatedTaskRepository automatedTaskRepository,
+                                                   AutomatedTaskConfigurationProperties automatedTaskConfigurationProperties,
```
Ben-Edwards-cgi, line 25: This should be the config for this task instead of the default one (So locks / users get assigned correctly)

src/main/java/uk/gov/hmcts/darts/arm/service/impl/TriggerArmRpoSearchServiceImpl.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * This method integrates various ARM RPO API calls to ultimately trigger a search. The results of that search are then processed by another automated
+     * task.
+     * <p/>
+     * If any of the underlying ArmRpoApi methods fail, ArmRpoApi will explicitly set the arm_rpo_execution_detail status to FAILED for that
+     * particular stage. So if you're thinking about adding any transactionality at the level of TriggerArmRpoSearchServiceImpl please be careful to avoid
+     * unwanted arm_rpo_execution_detail rollbacks.
+     */
+    @Override
+    public void triggerArmRpoSearch() {
```
Ben-Edwards-cgi, line 31: Needs unit tests

src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmRpoServiceImpl.java
chrisbellingham-hmcts: I think it's needed, but there were some challenges in this area, so I'd be interested in your perspective here.

The `userIdentity.getUserAccount()` call that obtains the `UserAccountEntity` takes place within `TriggerArmRpoSearchServiceImpl.triggerArmRpoSearch()`, as we must pass the entity to a series of method calls.

There is no active transaction within `triggerArmRpoSearch()`, so upon completion of `userIdentity.getUserAccount()`, our entity is no longer attached to an entity manager. A new transaction is then started here in `createArmRpoExecutionDetailEntity()`, and we merge the `userAccount` to the entity manager prior to saving the `ArmRpoExecutionDetailEntity` to avoid detached entity exceptions.

I deliberately avoided making the higher level method `triggerArmRpoSearch()` `@Transactional` on the basis that the underlying method calls (`getRecordManagementMatter()`, `getIndexesByMatterId()` etc) explicitly handle exceptions by setting the state and a failure status on the corresponding `ArmRpoExecutionDetailEntity`. And so, if the `triggerArmRpoSearch()` is made transactional, in the event of an exception bubbling up we will get the unwanted side-effect of the transaction rolling back the DB state, undoing those statuses.

There were a couple of other approaches that I thought about:
* Explictly create a more limited transaction using `TransactionTemplate`, wrapping the `userIdentity.getUserAccount()` and `armRpoService.createArmRpoExecutionDetailEntity()` calls. This would have added a little more code complexity than the merge aprpoach, but I think is feasible.
* Make `triggerArmRpoSearch()` transactional, and wrap all method calls in a try/catch, and catch `Exception` to avoid any exceptions that bubble up from triggering a rollback. For some reason I don't understand this approach caused constraint violation issues whenever an exception was thrown, and so state wasn't properly recorded in the DB.
* Make `triggerArmRpoSearch()` transactional, with `@Transactional(noRollbackFor = Exception.class)` but this doesn't appear to include all subsclasses of exception.

So on balance, I went with the merge approach as the simplest solution. But to be honest I'm a little unsure why other automated tasks don't need to take this approach, as they aren't run inside a transaction either. So I might be missing something...

src/main/java/uk/gov/hmcts/darts/arm/service/impl/TriggerArmRpoSearchServiceImpl.java
chrisbellingham-hmcts: Each of these methods set various fields on the underlying `ArmRpoExecutionDetailEntity` record, including the `armRpoState()` and `armRpoStatus` fields. So at this time, sequential calls are required by design (https://tools.hmcts.net/confluence/display/DMP/ARM+RPO).

Also I understand this code will only be run once a day by its corresponding automated task, so in this case, I'm not sure the perf improvements are worth the extra complexity of making the calls async.

src/main/java/uk/gov/hmcts/darts/task/runner/impl/ProcessE2EArmRpoPendingAutomatedTask.java
chrisbellingham-hmcts: Good spot.

src/main/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiImpl.java
```diff
@@ -284,7 +284,7 @@ public List<MasterIndexFieldByRecordClassSchema> getMasterIndexFieldByRecordClas
     }
 
     @Override
-    public void addAsyncSearch(String bearerToken, Integer executionId, UserAccountEntity userAccount) {
+    public String addAsyncSearch(String bearerToken, Integer executionId, UserAccountEntity userAccount) {
```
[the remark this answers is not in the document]
chrisbellingham-hmcts: Added return value assertion in `addAsyncSearch_shouldSucceedWhenASuccessResponseIsObtainedFromArm()`

src/main/java/uk/gov/hmcts/darts/arm/service/impl/TriggerArmRpoSearchServiceImpl.java
chrisbellingham-hmcts: Added test class `TriggerArmRpoSearchServiceImplTest`.

karen-hedges (approved):

src/main/java/uk/gov/hmcts/darts/arm/service/impl/TriggerArmRpoSearchServiceImpl.java
Ben-Edwards-cgi: Following the above I'm happy with this approach :)

src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmRpoServiceImpl.java
```diff
@@ -20,6 +22,19 @@ public class ArmRpoServiceImpl implements ArmRpoService {
 
     public static final String ARM_RPO_EXECUTION_DETAIL_NOT_FOUND = "ArmRpoExecutionDetail not found";
     private final ArmRpoExecutionDetailRepository armRpoExecutionDetailRepository;
+    private final EntityManager entityManager;
+
+    @Override
+    @Transactional
+    public ArmRpoExecutionDetailEntity createArmRpoExecutionDetailEntity(UserAccountEntity userAccount) {
```
Ben-Edwards-cgi, line 29: Needs unit tests

chrisbellingham-hmcts: Added test `ArmRpoServiceImplTest.createArmRpoExecutionDetailEntity_shouldCreateAndSaveExpectedEntityState()`

hemantasharma1129 (approved):

Ben-Edwards-cgi (approved):

PR #2272, opened 2024-11-14
DMP-3485: Automated task for Deletion from ARM's response folder

Ben-Edwards-cgi:
### Change description ###
Once ARM produces the response files, DETS-ARM pull response automated job will start pulling the response files and process it.

As soon as the files associated with a transaction are processed, DARTS should delete the corresponding response files from ARM's response folder.

Along with Synchronous deletion an automated job will also be required to make sure any deletion which is missed during Synchronous deletion are handled in a asynchronous way. The purpose of the job is to clean up data from  ARM response folders Asynchronously.
 * Create flyway script for new automated task
 ** Task name : DETSCleanupArmResponseFiles
 ** Batchsize : 4000
 ** Cron : Runs daily at 23:27
 ** Make sure the job is disabled
 * Create Automated job in darts-api - DETSCleanupArmResponseFilesAutomatedTask
 * The job will use the exact logic to fetch the EOD objects from external_object_directory with only dofference is the manifest file prefix would be 'DETS'. Refer BatchCleanupArmResponseFilesServiceImpl.cleanupResponseFiles.
 ** Location Type = ARM

 * 
 ** Status in (
STORED, ARM_RPO_PENDING, ARM_RESPONSE_PROCESSING_FAILED, ARM_RESPONSE_MANIFEST_FAILED, ARM_RESPONSE_CHECKSUM_VERIFICATION_FAILED)
 ** isResponseCleaned = false
 ** prefix=DETS

+*Acceptance Criteria:*+

*GIVEN*

I am system user and there exist EOD objects which false in Delete criteria

*WHEN*

the automate job DETSCleanupArmResponseFiles runs successfully

*THEN*

the corresponding response files (IU, CR, UF & IL) will be deleted in ARM's response folder

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/arm/service/impl/BatchCleanupArmResponseFilesServiceCommon.java
+++ b/src/main/java/uk/gov/hmcts/darts/arm/service/impl/BatchCleanupArmResponseFilesServiceCommon.java
@@ -52,6 +52,7 @@ public class BatchCleanupArmResponseFilesServiceCommon implements BatchCleanupAr
 
     @Override
     public void cleanupResponseFiles(int batchsize) {
+        System.out.println("TMP: " + manifestFilePrefix);
         if (batchsize == 0) {
             log.warn("{}: Batch Cleanup ARM Response Files - Batch size is 0, so not running", manifestFilePrefix);
             return;
@@ -114,7 +115,7 @@ public class BatchCleanupArmResponseFilesServiceCommon implements BatchCleanupAr
 
             log.debug("{}: Found ARM manifest file {} for cleanup", manifestFilePrefix, manifestFilename);
             List<InputUploadAndAssociatedFilenames> inputUploadAndAssociatedList = armResponseFileHelper.getCorrespondingArmFilesForManifestFilename(
-                manifestFilename);
+                manifestFilePrefix, manifestFilename);
             //inputUploadAndAssociatedList should only contain 1 matching InputUpload file, but looping through it just in case.
             for (InputUploadAndAssociatedFilenames inputUploadAndAssociates : inputUploadAndAssociatedList) {
                 deleteResponseFiles(userAccount, inputUploadAndAssociates, eodEntriesWithManifestFilename);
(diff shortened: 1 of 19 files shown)


src/main/java/uk/gov/hmcts/darts/arm/service/impl/BatchCleanupArmResponseFilesServiceCommon.java
```diff
@@ -34,42 +33,33 @@
...
-
-    private UserAccountEntity userAccount;
+    private final String manifestFilePrefix;
 
 
     @Override
     public void cleanupResponseFiles(int batchsize) {
         if (batchsize == 0) {
-            log.warn("Batch Cleanup ARM Response Files - Batch size is 0, so not running");
+            log.warn("{}: {}: Batch Cleanup ARM Response Files - Batch size is 0, so not running", manifestFilePrefix);
```
karen-hedges, line 56: Changing the log messages around will impact the test teams automated tests. You will need to check and see if they will be impacted by your changes.

src/main/resources/db/migration/common/V1_413__DETSCleanupArmResponseFiles_automated_task.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO darts.automated_task (aut_id,task_name,task_description,cron_expression,cron_editable, batch_size,
+created_ts, created_by,last_modified_ts,last_modified_by, task_enabled)
+VALUES (nextval('aut_seq'),'DETSCleanupArmResponseFiles','Clean up data from ARM response folders Asynchronously. ','0 27 23 * * ?',true,4000,
```
karen-hedges, line 3: Automated jobs are now using hardcoded values so don't use nextVal. You may also need to add a matching user for the job but not sure if that code has yet been merged.

karen-hedges: Also mention in the description it is for dets as the admin returns the descriptions of tasks so they can be run

src/main/java/uk/gov/hmcts/darts/arm/service/impl/BatchCleanupArmResponseFilesServiceCommon.java
karen-hedges, line 56: You are expecting 2 params in your log message but you are only providing 1.

Ben-Edwards-cgi: Will let them know before we merge :)

src/main/resources/db/migration/common/V1_413__DETSCleanupArmResponseFiles_automated_task.sql
Ben-Edwards-cgi: Updated

hemantasharma1129 (approved):

karen-hedges (approved):

PR #2288, opened 2024-11-19
DMP-4251: Introduce and use event obfuscation feature flag

Ben-Edwards-cgi:
### Change description ###
*Add feature flag*

If this flag has not already been created, add the new config to DARTS API application.yaml, to act as a feature flag for event obfuscation
{code:java}
darts:
  event-obfuscation:
    enabled: ${EVENT_OBFUSCATION_ENABLED:false} {code}
Check the env var "EVENT_OBFUSCATION_ENABLED" exists in flux for the darts-api service staging environment, set to true. Ensure it is not set or set to false for all other environments.

The feature should also be enabled on the dev and staging pipeline deployments, in order to facilitate testing.

*Hide endpoint*

The _POST /admin/events/obfuscate_ endpoint should be hidden if the feature is not enabled. Any requests should received a 501 not implemented response.
h3. *Acceptance Criteria*

*AC1 - feature enabled*

GIVEN the event obfuscation feature is enabled
WHEN the POST /admin/events/obfuscate endpoint is called
THEN a 200 response is received
AND the event text is obfuscated

*AC2 - feature not enabled*

GIVEN the event obfuscation feature is not enabled
WHEN the POST /admin/events/obfuscate endpoint is called
THEN a 501 response is received
AND the event text is not obfuscated

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/event/controller/EventsController.java
+++ b/src/main/java/uk/gov/hmcts/darts/event/controller/EventsController.java
@@ -68,7 +68,7 @@ public class EventsController implements EventApi {
     private final UserIdentity userIdentity;
 
     @Value("${darts.event-obfuscation.enabled}")
-    private final boolean eveventObfuscationEnabled;
+    private final boolean eventObfuscationEnabled;
 
     @Override
     @SecurityRequirement(name = SECURITY_SCHEMES_BEARER_AUTH)
@@ -189,7 +189,7 @@ public class EventsController implements EventApi {
     @Authorisation(contextId = ANY_ENTITY_ID,
         globalAccessSecurityRoles = {SUPER_ADMIN})
     public ResponseEntity<Void> adminObfuscateEveByIds(AdminObfuscateEveByIdsRequest adminObfuscateEveByIdsRequest) {
-        if (!eveventObfuscationEnabled) {
+        if (!eventObfuscationEnabled) {
             throw new DartsApiException(CommonApiError.FEATURE_FLAG_NOT_ENABLED, "Event obfuscation is not enabled");
         }
         this.dataAnonymisationService.anonymiseEventByIds(userIdentity.getUserAccount(), adminObfuscateEveByIdsRequest.getEveIds());
--- a/src/main/java/uk/gov/hmcts/darts/event/controller/EventsController.java
+++ b/src/main/java/uk/gov/hmcts/darts/event/controller/EventsController.java
@@ -6,6 +6,7 @@ import io.swagger.v3.oas.annotations.enums.ParameterIn;
 import io.swagger.v3.oas.annotations.security.SecurityRequirement;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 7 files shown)


src/main/resources/application.yaml
```diff
@@ -113,6 +113,8 @@ spring:
             loggerLevel: ${FEIGN_LOG_LEVEL:none}
 
 darts:
+  event-obfuscation:
+    enabled: ${EVENT_OBFUSCATION_ENABLED:false}
```
chrisbellingham-hmcts, line 117: Can you confirm this point from the ticket?
> Check the env var "EVENT_OBFUSCATION_ENABLED" exists in flux for the darts-api service staging environment, set to true. Ensure it is not set or set to false for all other environments.

src/main/java/uk/gov/hmcts/darts/event/controller/EventsController.java
```diff
@@ -64,6 +67,9 @@ public class EventsController implements EventApi {
     private final DataAnonymisationService dataAnonymisationService;
     private final UserIdentity userIdentity;
 
+    @Value("${darts.event-obfuscation.enabled}")
+    private final boolean eveventObfuscationEnabled;
```
chrisbellingham-hmcts, line 71: Typo, should be `eventObfuscationEnabled`?

src/main/resources/application.yaml
chrisbellingham-hmcts, line 117: Probably also worth enabling this for local dev in `application-local.yaml`.

src/main/java/uk/gov/hmcts/darts/event/controller/EventsController.java
Ben-Edwards-cgi: Updated :)

src/main/resources/application.yaml
Ben-Edwards-cgi: For local development we use environment properties to override these values.
(This approach aligns to the one we use for manual deletion)

Ben-Edwards-cgi: This is already set :)
https://github.com/hmcts/sds-flux-config/blob/master/apps/darts-modernisation/darts-api/stg.yaml

chrisbellingham-hmcts: Our approach isn't totally consistent but the general vision (for local development) has been to minimise which env vars we need to set (generally limited to secrets), and have everything else set to some sensible default in `application-local.yaml`. The goal is to make getting stood up for local development easier.

I'd think (?) for local dev we'd want this (and probably any other feature flags?) enabled 99.9% of the time, so it would make the dev experience a bit nicer if we can default it to true when running with the local profile (so in `application-local.yaml`):
```
  event-obfuscation:
    enabled: ${EVENT_OBFUSCATION_ENABLED:true}
```

Then we can still use an env var to override it if we really want to? Just one less thing to think about...

chrisbellingham-hmcts (approved):

jackmaloney:
Is there a flux PR for this change? If so, can you add it to the ticket? Cheers

davet1985:
> Is there a flux PR for this change? If so, can you add it to the ticket? Cheers

@jackmaloney I believe the flag was added to flux during the DARTS portal work

PR #2292, opened 2024-11-20
Dmp 3997 arm response dynamic invalid line files

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-3997

### Change description ###

Added code to allow for multiple ARM IL response files

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/arm/service/AbstractArmBatchProcessResponseFilesIntTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/arm/service/AbstractArmBatchProcessResponseFilesIntTest.java
@@ -124,6 +124,8 @@ abstract class AbstractArmBatchProcessResponseFilesIntTest extends IntegrationBa
 
         MediaEntity media3 = createMediaEntity(hearing, startTime, endTime, 3);
 
+        MediaEntity media4 = createMediaEntity(hearing, startTime, endTime, 4);
+
         OffsetDateTime startTime2 = OffsetDateTime.parse("2023-06-10T14:00:00Z");
         OffsetDateTime endTime2 = OffsetDateTime.parse("2023-06-10T14:45:00Z");
         MediaEntity media5 = createMediaEntity(hearing, startTime2, endTime2, 1);
@@ -152,8 +154,7 @@ abstract class AbstractArmBatchProcessResponseFilesIntTest extends IntegrationBa
         armEod2 = dartsPersistence.save(armEod2);
 
         ExternalObjectDirectoryEntity armEod3 = PersistableFactory.getExternalObjectDirectoryTestData().someMinimalBuilder()
-            .media(media3).status(dartsDatabase
-                                      .getObjectRecordStatusEntity(ARM_DROP_ZONE))
+            .media(media3).status(dartsDatabase.getObjectRecordStatusEntity(ARM_DROP_ZONE))
             .verificationAttempts(1).transferAttempts(1)
             .externalLocationType(dartsDatabase.getExternalLocationTypeEntity(ARM)).externalLocation(UUID.randomUUID()).build();
         armEod3.setTransferAttempts(1);
@@ -161,13 +162,20 @@ abstract class AbstractArmBatchProcessResponseFilesIntTest extends IntegrationBa
         armEod3.setVerificationAttempts(1);
         armEod3 = dartsPersistence.save(armEod3);
 
+        ExternalObjectDirectoryEntity armEod4 = PersistableFactory.getExternalObjectDirectoryTestData().someMinimalBuilder()
+            .media(media4).status(dartsDatabase.getObjectRecordStatusEntity(ARM_DROP_ZONE))
+            .verificationAttempts(1).transferAttempts(1)
+            .externalLocationType(dartsDatabase.getExternalLocationTypeEntity(ARM)).externalLocation(UUID.randomUUID()).build();
+        armEod4.setTransferAttempts(1);
+        armEod4.setManifestFile(manifestFile1);
+        armEod4.setVerificationAttempts(1);
+        armEod4 = dartsPersistence.save(armEod4);
+
         String manifestFile2 = prefix() + "_" + manifest2Uuid + ".a360";
         ExternalObjectDirectoryEntity armEod5 = PersistableFactory
             .getExternalObjectDirectoryTestData().someMinimalBuilder()
-            .media(media5).status(dartsDatabase
-                                      .getObjectRecordStatusEntity(ARM_DROP_ZONE))
-            .externalLocationType(dartsDatabase
-                                      .getExternalLocationTypeEntity(ARM)).externalLocation(UUID.randomUUID()).build();
+            .media(media5).status(dartsDatabase.getObjectRecordStatusEntity(ARM_DROP_ZONE))
+            .externalLocationType(dartsDatabase.getExternalLocationTypeEntity(ARM)).externalLocation(UUID.randomUUID()).build();
         armEod5.setTransferAttempts(1);
         armEod5.setManifestFile(manifestFile2);
         armEod5.setVerificationAttempts(1);
@@ -191,10 +199,13 @@ abstract class AbstractArmBatchProcessResponseFilesIntTest extends IntegrationBa
         String invalidLineFileFilename2 = String.format("%s_a17b9015-e6ad-77c5-8d1e-13259aae1896_0_il.rsp", hashcode1);
         String uploadFileFilename3 = String.format("%s_04e6bc3b-952a-79b6-8362-13259aae1897_1_uf.rsp", hashcode1);
         String invalidLineFileFilename3 = String.format("%s_a17b9015-e6ad-77c5-8d1e-13259aae1897_0_il.rsp", hashcode1);
+        String invalidLineFileFilename4 = String.format("%s_a17b9015-e6ad-77c5-8d1e-13259aae1892_0_il.rsp", hashcode1);
+        String invalidLineFileFilename5 = String.format("%s_a17b9015-e6ad-77c5-8d1e-13259aae1893_0_il.rsp", hashcode1);
 
         List<String> hashcodeResponses = List.of(createRecordFilename1, uploadFileFilename1,
                                                  createRecordFilename2, invalidLineFileFilename2,
-                                                 uploadFileFilename3, invalidLineFileFilename3);
+                                                 uploadFileFilename3, invalidLineFileFilename3,
+                                                 invalidLineFileFilename4, invalidLineFileFilename5);
 
         when(armDataManagementApi.listResponseBlobs(hashcode1)).thenReturn(hashcodeResponses);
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
```diff
@@ -282,18 +282,26 @@ private void processResponseFileByHashcode(BatchInputUploadFileFilenameProcessor
...
                     && (nonNull(armResponseBatchData.getCreateRecordFilenameProcessor())
                     || nonNull(armResponseBatchData.getArmResponseUploadFileRecord()))) {
 
                     preProcessResponseFilesActions(armResponseBatchData.getExternalObjectDirectoryId());
 
                     processInvalidLineFile(armResponseBatchData.getExternalObjectDirectoryId(),
-                                           armResponseBatchData.getInvalidLineFileFilenameProcessor(),
-                                           armResponseBatchData.getArmResponseInvalidLineRecord(),
+                                           armResponseBatchData.getInvalidLineFileFilenameProcessors().getFirst(),
+                                           armResponseBatchData.getArmResponseInvalidLineRecords().getFirst(),
```
Ben-Edwards-cgi, line 294: Should this process all entries in getArmResponseInvalidLineRecords() ?

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
```diff
@@ -317,6 +325,73 @@ && nonNull(armResponseBatchData.getUploadFileFilenameProcessor())) {
         );
     }
 
+    private void processMultipleInvalidLineFiles(ArmResponseBatchData armResponseBatchData) {
+        try {
+            ExternalObjectDirectoryEntity externalObjectDirectory = getExternalObjectDirectoryEntity(armResponseBatchData.getExternalObjectDirectoryId());
+            var invalidLineRecord1 = armResponseBatchData.getArmResponseInvalidLineRecords().getFirst();
+            var invalidLineRecord2 = armResponseBatchData.getArmResponseInvalidLineRecords().getLast();
```
Ben-Edwards-cgi, line 332: This can be moved into one of the nested if conditions (just above if (if (nonNull(invalidLineRecord1) &&)

As we exclusively use it for that condition (So no point grabbing it if we do not need it)

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
```diff
@@ -317,6 +325,73 @@ && nonNull(armResponseBatchData.getUploadFileFilenameProcessor())) {
...
+
+        String errorCode;
+        if (CREATE_RECORD.equalsIgnoreCase(uploadNewFileRecord1.getOperation())) {
+            errorCode = invalidLineRecord1.getExceptionDescription() + " " + invalidLineRecord2.getExceptionDescription();
+        } else if (CREATE_RECORD.equalsIgnoreCase(uploadNewFileRecord2.getOperation())) {
+            errorCode = invalidLineRecord2.getExceptionDescription() + " " + invalidLineRecord1.getExceptionDescription();
+        } else {
+            errorCode = invalidLineRecord1.getExceptionDescription() + " " + invalidLineRecord2.getExceptionDescription();
+        }
+        return errorCode;
```
Ben-Edwards-cgi, line 392: Can be condensed
`if (CREATE_RECORD.equalsIgnoreCase(uploadNewFileRecord2.getOperation())) {
            return invalidLineRecord2.getExceptionDescription() + " " + invalidLineRecord1.getExceptionDescription();`
            }
            return errorCode = invalidLineRecord1.getExceptionDescription() + " " + invalidLineRecord2.getExceptionDescription();`

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
```diff
@@ -716,8 +791,9 @@ private static List<String> getResponseBlobsToBeDeleted(ArmResponseBatchData arm
         if (nonNull(armResponseBatchData.getUploadFileFilenameProcessor())) {
             responseBlobsToBeDeleted.add(armResponseBatchData.getUploadFileFilenameProcessor().getUploadFileFilenameAndPath());
         }
-        if (nonNull(armResponseBatchData.getInvalidLineFileFilenameProcessor())) {
-            responseBlobsToBeDeleted.add(armResponseBatchData.getInvalidLineFileFilenameProcessor().getInvalidLineFileFilenameAndPath());
+        if (CollectionUtils.isNotEmpty(armResponseBatchData.getInvalidLineFileFilenameProcessors())) {
```
Ben-Edwards-cgi, line 794: Guessing this collection will never contain null entries?

src/integrationTest/java/uk/gov/hmcts/darts/arm/service/AbstractArmBatchProcessResponseFilesIntTest.java
```diff
@@ -204,20 +215,47 @@ void batchProcessResponseFiles_WithMediaReturnsSuccess() throws IOException {
...
+        when(armDataManagementApi.getBlobData(invalidLineFileFilename5)).thenReturn(invalidLineFileBinaryDataTest5);
+
+        when(armDataManagementApi.deleteBlobData(createRecordFilename1)).thenReturn(true);
+        when(armDataManagementApi.deleteBlobData(uploadFileFilename1)).thenReturn(true);
+        when(armDataManagementApi.deleteBlobData(createRecordFilename2)).thenReturn(true);
+        when(armDataManagementApi.deleteBlobData(invalidLineFileFilename2)).thenReturn(true);
+        when(armDataManagementApi.deleteBlobData(uploadFileFilename3)).thenReturn(true);
+        when(armDataManagementApi.deleteBlobData(invalidLineFileFilename3)).thenReturn(true);
+        when(armDataManagementApi.deleteBlobData(invalidLineFileFilename4)).thenReturn(true);
+        when(armDataManagementApi.deleteBlobData(invalidLineFileFilename5)).thenReturn(true);
```
Ben-Edwards-cgi, line 247: If we always return true for this method we can simply do 
`when(armDataManagementApi.deleteBlobData(any())).thenReturn(true);`

src/integrationTest/java/uk/gov/hmcts/darts/arm/service/AbstractArmBatchProcessResponseFilesIntTest.java
```diff
@@ -234,6 +272,10 @@ void batchProcessResponseFiles_WithMediaReturnsSuccess() throws IOException {
         when(armDataManagementApi.deleteBlobData(uploadFileFilename3)).thenReturn(true);
         when(armDataManagementApi.deleteBlobData(invalidLineFileFilename3)).thenReturn(true);
         when(armDataManagementApi.deleteBlobData(blobNameAndPath1)).thenReturn(true);
+        when(armDataManagementApi.deleteBlobData(invalidLineFileFilename4)).thenReturn(true);
+        when(armDataManagementApi.deleteBlobData(invalidLineFileFilename5)).thenReturn(true);
+
+        when(armDataManagementApi.deleteBlobData(createRecordFilename5)).thenReturn(true);
```
Ben-Edwards-cgi, line 278: See above

src/integrationTest/java/uk/gov/hmcts/darts/arm/service/AbstractArmBatchProcessResponseFilesIntTest.java
```diff
@@ -312,9 +354,104 @@ void batchProcessResponseFiles_WithMediaReturnsSuccess() throws IOException {
...
+        BinaryData invalidLineFileBinaryDataTest2 = convertStringToBinaryData(getInvalidLineFileContents(invalidLineFileTest2, armEod1.getId()));
+        BinaryData invalidLineFileBinaryDataTest3 = convertStringToBinaryData(getInvalidLineFileContents(invalidLineFileTest3, armEod1.getId()));
+
+        when(armDataManagementApi.getBlobData(invalidLineFileFilename1)).thenReturn(invalidLineFileBinaryDataTest1);
+        when(armDataManagementApi.getBlobData(invalidLineFileFilename2)).thenReturn(invalidLineFileBinaryDataTest2);
+        when(armDataManagementApi.getBlobData(invalidLineFileFilename3)).thenReturn(invalidLineFileBinaryDataTest3);
+
+        when(armDataManagementApi.deleteBlobData(invalidLineFileFilename1)).thenReturn(true);
+        when(armDataManagementApi.deleteBlobData(invalidLineFileFilename2)).thenReturn(true);
+        when(armDataManagementApi.deleteBlobData(invalidLineFileFilename3)).thenReturn(true);
```
Ben-Edwards-cgi, line 418: If we always return true for this method we can condense this to
when(armDataManagementApi.deleteBlobData(any())).thenReturn(true);

src/integrationTest/java/uk/gov/hmcts/darts/arm/service/AbstractArmBatchProcessResponseFilesIntTest.java
```diff
@@ -312,9 +354,104 @@ void batchProcessResponseFiles_WithMediaReturnsSuccess() throws IOException {
...
+
+        when(armDataManagementApi.getBlobData(invalidLineFileFilename1)).thenReturn(invalidLineFileBinaryDataTest1);
+        when(armDataManagementApi.getBlobData(invalidLineFileFilename2)).thenReturn(invalidLineFileBinaryDataTest2);
+        when(armDataManagementApi.getBlobData(invalidLineFileFilename3)).thenReturn(invalidLineFileBinaryDataTest3);
+
+        when(armDataManagementApi.deleteBlobData(invalidLineFileFilename1)).thenReturn(true);
+        when(armDataManagementApi.deleteBlobData(invalidLineFileFilename2)).thenReturn(true);
+        when(armDataManagementApi.deleteBlobData(invalidLineFileFilename3)).thenReturn(true);
```
Ben-Edwards-cgi, line 426: If we always return true for this method we can condense this to
when(armDataManagementApi.deleteBlobData(any())).thenReturn(true);

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
karen-hedges: Its a list and is always created. Items are only added to the list.

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
karen-hedges: Sorry that was a type in the else if statement.

hemantasharma1129 (approved):

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
Ben-Edwards-cgi: Is it possible to add null as an item to the list?

karen-hedges: Line 673 is the only place the IL responses are added and line 672 has a check for null before adding to the list so no, nulls wont get added.

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
Ben-Edwards-cgi: This checks that the getInvalidLineFileFilenameProcessors has size of 1 but not the getArmResponseInvalidLineRecords()

karen-hedges: There can only be one as the addition of the objects add the 2 types of objects at the same time. Its completely unnecessary to add another check for the size of the other list.

Ben-Edwards-cgi: As a general rule I think its better to protect against the edge cases, to ensure data integrity, and to allow for easier re-use. However, I'm happy if there is just a comment in both sections of code (This section and where the items get added) to ensure if we edit the code in the future we know we must add to both

chrisbellingham-hmcts: If we were to codify a second `size()` check, it's not clear if the second negative case would be reachable code (from a either a test or practical perspective), based on my reading of the calling code.

So on that basis I agree that a comment sounds reasonable, stating the invariant that _both_ `armResponseInvalidLineRecords` and `invalidLineFileFilenameProcessors` are expected to contain one (and only one) item.

chrisbellingham-hmcts (approved):

PR #2293, opened 2024-11-20
DMP-4199: Only expire event when all linked cases have expired

Ben-Edwards-cgi:
### Change description ###
When the case expiry job runs, the event text of all associated events are anonymised. However, events can be linked to multiple cases. Therefore, the event text should only be removed once the is_data_anonymised flag for all linked cases has been set to true. 
h4. Acceptance criteria

*AC1: Event linked to multiple cases, with all* *is_data_anonymised flags set to true for linked cases*

GIVEN

a single event is linked to multiple cases

AND 

the is_data_anonymised flag is set to true for all linked cases 

WHEN 

the case expiry job runs

THEN

the event text is anonymised

 

*AC2: Event linked to multiple cases, with at least one is_data_anonymised flag is set to false*

GIVEN

a single event is linked to multiple cases

AND 

the is_data_anonymised flag is set to false for at least one linked cases that has a retain_until_ts in the future

WHEN 

the case expiry job runs

THEN

the event text is not anonymised

```
[ ] Yes
[X] No
```


src/integrationTest/java/uk/gov/hmcts/darts/task/runner/impl/CaseExpiryDeletionAutomatedTaskITest.java
```diff
@@ -66,6 +71,36 @@ void positiveRetentionDatePassed() {
...
+        final int caseId1 = courtCase1.getId();
+
+        caseExpiryDeletionAutomatedTask.preRunTask();
+        caseExpiryDeletionAutomatedTask.runTask();
+
+        assertCase(caseId1, true, event.getId());
+    }
+
+    @Test
+    void positiveRetentionDatePassedForBothCaseLinkedEventsAnoymised() {
```
jackmaloney, line 90: I'm finding some of these integration test names difficult to read. There was some work done by Mario a while back to try and standardise and make clearer test names. See here for examples https://github.com/mario-paniccia/darts-api-fork/pull/1 Any change you could try and incorporate that? e.g. given when then style?

src/main/java/uk/gov/hmcts/darts/common/repository/EventLinkedCaseRepository.java
```diff
@@ -13,4 +15,16 @@ public interface EventLinkedCaseRepository extends JpaRepository<EventLinkedCase
...
+        SELECT COUNT(DISTINCT cc) = (COUNT(cc.isDataAnonymised) filter (where cc.isDataAnonymised = true))
+                    FROM EventLinkedCaseEntity elc 
+                    LEFT JOIN CourtCaseEntity cc 
+                    ON (elc.courtCase = cc 
+                    or (cc.courthouse.courthouseName = elc.courthouseName and cc.caseNumber = elc.caseNumber)) 
+                    WHERE elc.event = :eventEntity
+                    group by elc.event
+        """
+    )
+    boolean allAssociatedCasesAnonymised(EventEntity eventEntity);
```
jackmaloney, line 29: Maybe prepend with 'count' and append with 'byEventEntity'?

src/main/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImpl.java
```diff
@@ -65,14 +65,13 @@ void anonymiseCourtCaseEntity(UserAccountEntity userAccount, CourtCaseEntity cou
...
         courtCase.getHearings().forEach(hearingEntity -> anonymiseHearingEntity(userAccount, hearingEntity));
-        anonymiseAllEventsFromCase(userAccount, courtCase);
-
         courtCase.markAsExpired(userAccount);
         //This also saves defendant, defence and prosecutor entities
         tidyUpTransformedMediaEntities(userAccount, courtCase);
         auditApi.record(AuditActivity.CASE_EXPIRED, userAccount, courtCase);
         anonymiseCreatedModifiedBaseEntity(userAccount, courtCase);
         caseService.saveCase(courtCase);
+        anonymiseAllEventsFromCase(userAccount, courtCase);
```
jackmaloney, line 74: When does it actually save the anonymised event back to the database?

src/test/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImplTest.java
```diff
@@ -87,16 +87,58 @@ private void assertLastModifiedByAndAt(CreatedModifiedBaseEntity entity, UserAcc
 
 
     @Test
-    void positiveEventEntityAnonymise() {
+    void positiveEventEntityAnonymiseNotUpdatedAsNotAllCasesExpiredAndOnlyAnonymiseIfAllCasesExpiredIsTrue() {
```
jackmaloney, line 90: Why do you need the word 'positive' in front of every test? What purpose is that serving?

src/integrationTest/java/uk/gov/hmcts/darts/task/runner/impl/CaseExpiryDeletionAutomatedTaskITest.java
Ben-Edwards-cgi: My understanding is JUnit5 test names can't have underscores in them as it fails HMCTS style rules.
However I can add a @DislayName to the tests to provide more human readable context

src/main/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImpl.java
Ben-Edwards-cgi: Spring should auto save any modifications when we exit the transaction.
But for the avoidance of any doubt I have added a save call into the anonymise method

src/integrationTest/java/uk/gov/hmcts/darts/task/runner/impl/CaseExpiryDeletionAutomatedTaskITest.java
chrisbellingham-hmcts: I think there's a rule exclusion somewhere to allow underscores in test names. It's how I tend to name my tests nowadays.

src/test/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImplTest.java
Ben-Edwards-cgi: Its a test term:

positive = Indicates the test is verifying things we expect the application to do
negative = Indicates the test is verifying things we do not expect the application to do (error scenarios and the such)

src/integrationTest/java/uk/gov/hmcts/darts/task/runner/impl/CaseExpiryDeletionAutomatedTaskITest.java
Ben-Edwards-cgi: Is that an exclusion we added as the HMCTS base spring template prevents its? 
(My understanding is we should be following the HMCTs styles)

src/test/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImplTest.java
jackmaloney: Ok I still don't get the benefit of this and seems odd that this has been introduced as a pattern when it hasn't been done by anyone else in the code base. 
I keep re-reading these test names and really struggling to get my head round what you are trying to say. For someone new coming in years down the line they're not going to have a clue. Please can you try and use some of the patterns we already have in place for test names?

src/integrationTest/java/uk/gov/hmcts/darts/task/runner/impl/CaseExpiryDeletionAutomatedTaskITest.java
chrisbellingham-hmcts: I assume it's this one `<exclude name="MethodNamingConventions"/>` in `config/pmd/test-ruleset.xml`, which was added by us a year ago.

src/test/java/uk/gov/hmcts/darts/common/service/impl/DataAnonymisationServiceImplTest.java
chrisbellingham-hmcts: Worth a chat in best-practices I think to refine some common naming pattern. I know we tried this a while back and didn't get very far, but now that we're fewer in numbers it's worth revisiting. I'll initiate something.

Ben-Edwards-cgi: Its probs also worth using nested classes as it helps structure the test cases

src/integrationTest/java/uk/gov/hmcts/darts/task/runner/impl/CaseExpiryDeletionAutomatedTaskITest.java
Ben-Edwards-cgi: If we added this exclusion it will be worth checking with the spring template and HMCTs standards to see why this was added in the first place. As we should be aligned to both the standards and templates

jackmaloney: That exclusion is only in the test ruleset so in my opinion that is fine if we think it is benficial for our project, and on this occasion I believe that to be the case as it allows us to write much clearer test names than what we have in this PR. 
There are many examples of other HMCTS projects using underscores within test names, we are not the first team to do this. 

For the purpose of this PR, if you don't want to align with what is being suggested, can you at least make the test name clearer? Start by removing this word 'positive' in front of every test and say what the outcome of the test should be, as currently that is not clear.

Ben-Edwards-cgi: I'm happy to change it but this pattern has been used on almost 100 PRs I have raised (and merged) on darts, why is this only being mentioned now?

jackmaloney (approved):

karen-hedges (approved):

PR #2307, opened 2024-11-22
DMP-4382: (DevOnly) PmdMain / PmdTestCommon do not get applied correctly in the build pipeline

Ben-Edwards-cgi:
### Change description ###
# Summary of Git Diff

This Git diff modifies the `build.gradle` file to enhance the configuration of the PMD (Programming Mistake Detector) tool. The changes primarily involve defining specific configurations for different source sets, ensuring better management of PMD rules and reporting.

## Highlights
- **Tool Version**: The PMD tool version remains at `7.7.0`.
- **Reports Directory**: The reports directory is set to `$project.buildDir/reports/pmd`.
- **Rule Sets**: The previous configuration that used a single source set for PMD checks has been replaced with multiple source sets for different test categories.
- **New PMD Configurations**:
  - `pmdMain`: Configured to use `main-ruleset.xml` with a maximum of 1308 allowed failures.
  - `pmdTestCommon`: Configured to use `test-ruleset.xml` with a maximum of 407 allowed failures.
  - `pmdTest`, `pmdFunctionalTest`, `pmdIntegrationTest`, and `pmdSmokeTest` are all set to use `test-ruleset.xml`.
- **Commentary**: Added comments to emphasize the goal of reducing PMD errors rather than allowing them to increase.

```
[ ] Yes
[X] No
```


build.gradle
```diff
@@ -221,21 +221,35 @@ checkstyle {
 
 var sourceSetsCopy = sourceSets
```
chrisbellingham-hmcts, line 222: Seems to be unused now, can we remove?

build.gradle
```diff
@@ -221,21 +221,35 @@ checkstyle {
...
-  sourceSets = [sourceSetsCopy.test, sourceSetsCopy.functionalTest, sourceSetsCopy.integrationTest, sourceSetsCopy.smokeTest]
-  reportsDir = file("$project.buildDir/reports/pmd")
-  // https://github.com/pmd/pmd/issues/876
-  ruleSets = []
-  ruleSetFiles = files("config/pmd/test-ruleset.xml")
+
+pmdMain {
+    ruleSetFiles = files("config/pmd/main-ruleset.xml")
+    //Ensures we don't passively increase pmd errors but we should aim to reduce this
+    maxFailures = 1308
```
chrisbellingham-hmcts, line 234: Ouch

chrisbellingham-hmcts (approved):

build.gradle
Ben-Edwards-cgi: Updated :)

davet1985 (approved):

hemantasharma1129 (approved):

jackmaloney:
Do we need all that whitespace? Makes reviewing PR's difficult

Ben-Edwards-cgi:
> Do we need all that whitespace? Makes reviewing PR's difficult

The only real changes are here: 
image


The white space is added by the InteliJ style rules I imported from HMCTs I'm happy to update these rules if you have a latest spec / import file for them?

Ben-Edwards-cgi:
@davet1985  build will constantly fail for now as the PMD limits will be increasing with each PR that gets merged.
I will update these limits to match master once we have two reviews and the build should pass and we can merge. Doing this before hand would mean we would have to update these every time master has an update

davet1985 (approved):

jackmaloney (approved):

PR #2310, opened 2024-11-25
Dmp 4373 dets to arm darts to arm invalid line files failing

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-4373

### Change description ###

Added code to parse the input field if the invalid line file response input field has invalid json

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/darts/arm/util/ArmResponseFilesUtil.java
```diff
@@ -37,4 +41,28 @@ public static String getObjectTypeId(ExternalObjectDirectoryEntity externalObjec
...
+
+    public static Integer getRelationIdFromArmResponseFilesInputField(String input) {
+        try {
+            String relationId = findFirstPatternMatch("\"relation_id\".*?:.*?\"([^\"]+)\"", input);
+            return Integer.parseInt(relationId);
+        } catch (Exception e) {
+            log.error("Error: {}", e.getMessage());
+        }
+        return null;
+    }
```
Ben-Edwards-cgi, line 53: @hemantasharma1129  @jackmaloney  if we get invalid JSON returned to us should we still be trying to to get a relation_id or should we log this for live support to work out?

Conscious this could cause us issues if we grab the incorrect relation_id

src/main/java/uk/gov/hmcts/darts/arm/util/ArmResponseFilesUtil.java
```diff
@@ -37,4 +41,28 @@ public static String getObjectTypeId(ExternalObjectDirectoryEntity externalObjec
...
+
+    public static String getOperationFromArmResponseFilesInputField(String input) {
+        return findFirstPatternMatch("\"operation_id\".*?:.*?\"([^\"]+)\"", input);
```
Ben-Edwards-cgi, line 56: Dependent on the response from the previous comment it maybe better to assume we always get valid json return then grab the operation_id from this valid json.

If the response from previous comment is that we still want to grab these details from invalid JSON then this comment can be ignored :)

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
```diff
@@ -387,16 +389,35 @@ private void processMultipleInvalidLineFiles(ArmResponseBatchData armResponseBat
...
+    private String getOperation(ArmResponseInvalidLineRecord record) {
+        UploadNewFileRecord uploadRecord = readInputJson(record.getInput());
+        if (uploadRecord != null) {
+            return uploadRecord.getOperation();
+        }
+        String operation = getOperationFromArmResponseFilesInputField(record.getInput());
+        return operation != null ? operation : "UNKNOWN";
+    }
+
+    private void appendErrorDescription(StringBuilder errorDescription, String operation, ArmResponseInvalidLineRecord record) {
```
Ben-Edwards-cgi, line 418: Entire class could benefit from unit testing

src/main/java/uk/gov/hmcts/darts/arm/util/ArmResponseFilesUtil.java
hemantasharma1129: We must extract the relationId from IL file for error handling. Thats how ARM will respond back so DARTS should be smart enough to extract it and update the DB with correct error message.

hemantasharma1129 (approved):

src/main/java/uk/gov/hmcts/darts/arm/util/ArmResponseFilesUtil.java
Ben-Edwards-cgi: Does this include when the JSON given to us is invalid?
It feels very risky to try and pull out the value from invalid json? as we could get the incorrect value and process it as if it is valid causing data integrity issues

hemantasharma1129: If the question is why reading the JSON, then answer is we must read it to make sure we update the DARTS system in case DARTS sends an INVALID manifest to ARM. Even if its invalid, we should be able to extract the relation_id to update DARTS db.
If the concern is around the Regex solution, happy to hear any other solution which is much safer to allow us to extract the relation_id from the IL file.

Ben-Edwards-cgi: I think any solution we have if we get invalid json (As in json structure is invalid not just wrong fields) something more fundamental has gone wrong that we need to fix first then we can replay these fields. It feels very risky to try and process invalid json to pull out the values, as we have 0 control or validation against this invalid json response, so it could cause security vulnerabilities as we are accepting what every they give us as well as data integrity issues as we are marking fields as invalid from the invalid json we get but the invalid Json could have returned us the wrong data (Which is likely if it is invalid json in the first place) so we could be marking the incorrect records as invalid. 

I believe the safest approach is to log the invalid json and skip processing then raise an alert on DynaTrace for live support team to investigate and if they want to manually replay or mark the file as failed.

src/main/java/uk/gov/hmcts/darts/arm/util/ArmResponseFilesUtil.java
karen-hedges: See Hemanta's comments

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
karen-hedges: Updated integration tests to account for new code

src/main/java/uk/gov/hmcts/darts/arm/util/ArmResponseFilesUtil.java
Ben-Edwards-cgi: I'm happy for all other failures but for invalid json as it is a critical failure I don't think we should try to derive the values but instead force the logging route for it.

But for all other failures we can convert the object into json and grab the value (instead of using regex).

@hemantasharma1129  is this what you mean? If not it maybe worth having a quick 5-10 min chat on this

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
karen-hedges: added some unit tests

src/main/java/uk/gov/hmcts/darts/arm/util/ArmResponseFilesUtil.java
karen-hedges: this has been removed

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
```diff
@@ -678,13 +713,23 @@ private void readInvalidLineFile(BinaryData invalidLineFileBinary, InvalidLineFi
...
+                            armBatchResponses.addResponseBatchData(Integer.valueOf(uploadNewFileRecord.getRelationId()),
+                                                                   armResponseInvalidLineRecord, invalidLineFileFilenameProcessor);
+
+                        } else {
+                            log.warn("Unable to get EOD id (relation id) from uploadNewFileRecord {} invalid line file {}",
+                                     armResponseInvalidLineRecord.getInput(), invalidLineFileFilenameProcessor.getInvalidLineFilename());
+                        }
+                    } else {
+                        log.warn("Failed to obtain EOD id (relation id) from invalid line record {} from file {}", input,
+                                 invalidLineFileFilenameProcessor.getInvalidLineFilename());
```
Ben-Edwards-cgi, line 729: This code seems to be repeated a fair few times wonder if this could be pulled out into its own method?

src/main/java/uk/gov/hmcts/darts/arm/util/ArmResponseFilesUtil.java
```diff
@@ -9,13 +10,13 @@
 
 @UtilityClass
 @SuppressWarnings({"HideUtilityClassConstructor"})
+@Slf4j
```
Ben-Edwards-cgi, line 13: This is added but does not seem to be used?

src/test/java/uk/gov/hmcts/darts/common/service/impl/EodHelperMocks.java
```diff
@@ -37,6 +42,7 @@
 /**
  * Helper test class that mocks {@link EodHelper} entities and methods so that they can be used in unit tests.
  */
+@SuppressWarnings("PMD.NcssCount")
```
Ben-Edwards-cgi, line 45: Should be actioned instead of supressed

src/test/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchProcessResponseFilesImplTest.java
```diff
@@ -0,0 +1,256 @@
...
+    @Mock
+    private ObjectStateRecordRepository osrRepository;
+    @Mock
+    private CurrentTimeHelper timeHelper;
+    @Mock
+    private UserIdentity userIdentity;
+
+    private DetsToArmBatchProcessResponseFilesImpl detsToArmBatchProcessResponseFilesImpl;
+
+    private static final EodHelperMocks EOD_HELPER_MOCKS = new EodHelperMocks();
```
Ben-Edwards-cgi, line 70: Should this be non-static as we would want the mocks to reset after each test?

src/test/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchProcessResponseFilesImplTest.java
```diff
@@ -0,0 +1,256 @@
...
+    @Test
+    void preProcessResponseFilesActions() {
+        // given
+        when(osrRepository.findByArmEodId(String.valueOf(externalObjectDirectoryEntity.getId()))).thenReturn(Optional.of(objectStateRecordEntity));
+
+        // when
+        detsToArmBatchProcessResponseFilesImpl.preProcessResponseFilesActions(externalObjectDirectoryEntity.getId());
+
+        // then
+        verify(osrRepository, times(1)).findByArmEodId(String.valueOf(externalObjectDirectoryEntity.getId()));
```
Ben-Edwards-cgi, line 141: Verifications are usually better when they use hard coded values instead of method calls.

I.e getId() could return null but we actually want it to be 123. So checking 123 gives us more assurance

src/test/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchProcessResponseFilesImplTest.java
```diff
@@ -0,0 +1,256 @@
...
+
+        // when
+        detsToArmBatchProcessResponseFilesImpl.onUploadFileChecksumValidationSuccess(batchInputUploadFileFilenameProcessor,
+                                                                                     batchData,
+                                                                                     uploadFileRecord,
+                                                                                     externalObjectDirectoryEntity,
+                                                                                     checksum);
+
+        // then
+        verify(osrRepository, times(1)).findByArmEodId(anyString());
```
Ben-Edwards-cgi, line 171: Verify should use expected values instead of any just to make sure it was hit with the right things

src/test/java/uk/gov/hmcts/darts/common/service/impl/EodHelperMocks.java
karen-hedges: This is on a test class so its not important

src/test/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchProcessResponseFilesImplTest.java
```diff
@@ -0,0 +1,256 @@
...
+    void processInvalidLineFileActions() {
+        // given
+        ArmResponseInvalidLineRecord invalidLineRecord = new ArmResponseInvalidLineRecord();
+        invalidLineRecord.setErrorStatus("errorStatus");
+        invalidLineRecord.setExceptionDescription("exceptionDescription");
+        invalidLineRecord.setInput(
+            "{\\\"operation\\\":\\\"create_record\\\",\\\"relation_id\\\":\\\"<EODID>\\\","
+                + "\\\"record_metadata\\\":{\\\"record_class\\\":\\\"A360TEST\\\",\\\"publisher\\\":\\\"A360\\\","
+                + "\\\"region\\\":\\\"GBR\\\",\\\"title\\\":\\\"CGITestFilesMalformedManifest_1.00MB_100.00MB_001\\\","
+                + "\\\"recordDate\\\":\\\"2023-12-21T10:03:53Z\\\"}}");
```
Ben-Edwards-cgi, line 217: Optional: use """ instead of " to allow for easier reading of the text (would prevent the need for double escapes, etc

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
karen-hedges: The pattern is similar but the objects retrieved are different and each type has there own methods. Its simple enough to just follow the logic.

src/main/java/uk/gov/hmcts/darts/arm/util/ArmResponseFilesUtil.java
karen-hedges: removed

src/test/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchProcessResponseFilesImplTest.java
```diff
@@ -0,0 +1,256 @@
...
+    void preProcessResponseFilesActions() {
+        // given
+        when(osrRepository.findByArmEodId(String.valueOf(externalObjectDirectoryEntity.getId()))).thenReturn(Optional.of(objectStateRecordEntity));
+
+        // when
+        detsToArmBatchProcessResponseFilesImpl.preProcessResponseFilesActions(externalObjectDirectoryEntity.getId());
+
+        // then
+        verify(osrRepository, times(1)).findByArmEodId(String.valueOf(externalObjectDirectoryEntity.getId()));
+        verify(osrRepository, times(1)).save(objectStateRecordEntity);
```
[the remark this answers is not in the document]
karen-hedges: done

src/test/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchProcessResponseFilesImplTest.java
karen-hedges: done

src/test/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchProcessResponseFilesImplTest.java
karen-hedges: done

src/test/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchProcessResponseFilesImplTest.java
karen-hedges: done

src/test/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchProcessResponseFilesImplTest.java
Ben-Edwards-cgi: Its more for when we use verify as it will use the saved version for the counts, in this cause the static version which will never change.

so verify(EOD_HELPER_MOCKS.getMockedArmLocation(),times(1)).getId(), in this test and the same getId was called in a previous test the times called will be (2) but we only want it to be (1) as it has not reset the call count from the previous test as it is statically defined

src/test/java/uk/gov/hmcts/darts/common/service/impl/EodHelperMocks.java
Ben-Edwards-cgi: It is still important for test classes as it improves test maintainability, readability and quality.

karen-hedges: ITs been removed

src/test/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchProcessResponseFilesImplTest.java
karen-hedges: I don't need to verify the EOD_HELPER_MOCKS

Ben-Edwards-cgi: Its more if we do verify on other tests, if this runs first it will break the verify

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
Ben-Edwards-cgi: Its more from a maintainability and testability view point having the logic centralised means updates are centralised and changes do not get missed

karen-hedges: Its unnecessary testing.

Ben-Edwards-cgi: Sorry I mean having centralised logic instead of duplicate code means we only need to test it once instead of in every place it is used so we gain much high quality tests.

and by having centralised logic in production code it reduces the risk of defects and makes changes easier as you only need to change / test one place instead of all instances which means it becomes easier to miss stuff

jackmaloney: @Ben-Edwards-cgi do you mean to merge with other methods e.g. `readUploadFile` in the same class which looks very similar to this one? If so, I tend to agree that would be a good idea. This class is nearly a 1000 lines long now which is way too many. Anything to make things cleaner is probably good from a maintainability point, especially for another developer picking this up in future.

Ben-Edwards-cgi: Yeah exactly that just so the logic is in a method we can re-use to prevent us from duplicating

karen-hedges: This is a bug ticket. To optimise here is going to take time and 3 other tickets are awaiting this fix. If this is so necessary raise another to optimise but this needs to go in ASAP.

src/test/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchProcessResponseFilesImplTest.java
karen-hedges: Again we should not be verifying the static test data.

Ben-Edwards-cgi: We may need to verify the returned result was called correctly to ensure we pass the correct values to other methods. For instance it had its getId called which we then pass to a repository, as these are all stored within your statically defined  EodHelperMocks the times it was called will be incorrect across tests

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
hemantasharma1129: Dont think code optimisation/refactoring should be done part of this ticket. There should be a separate ticket to refactor the class and shouldnt be done part of a bug ticket which is blocking 6 other tickets.

jackmaloney: Yeh happy for a ticket to be raised to track this and get this one merged

karen-hedges: Created ticket to refactor: https://tools.hmcts.net/jira/browse/DMP-4414

src/test/java/uk/gov/hmcts/darts/arm/service/impl/DetsToArmBatchProcessResponseFilesImplTest.java
Ben-Edwards-cgi: So we can get this ticket merged in I have raised: https://tools.hmcts.net/jira/browse/DMP-4415 to look at this with a wider audience to gain a consensus on what the view is around static mocks

Ben-Edwards-cgi (approved):
Approved with exception to the comments that will be addressed in:
https://tools.hmcts.net/jira/browse/DMP-4415
https://tools.hmcts.net/jira/browse/DMP-4414

PR #2320, opened 2024-11-26
DMP-2724: Inform ARM to delete unstructured data for a manual deletion request

Ben-Edwards-cgi:
### Change description ###
Following a user selected an audio/transcript file for manual deletion, DARTS will need to inform ARM so that the unstructured data is requested for manual deletion.

This will involve updating metadata and passing it to ARM.  Essentially resetting the retention date for the files to be deleted so that they have "expired" and providing a reason of "Manual Deletion".

In DMP-2925, the object from INBOUND and UNSTRUCTURED location gets deleted. Part of this ticket the same logic needs to be extended for ARM as well.

Since part of DMP-2925 most of the logic will be implemented already, we just have to implement the part highlighted in {color:#ff8b00}*ORANGE.*{color}

The task should identify files for manual deletion using the following logic
 * find all entries in the object_admin_action table where "marked_for_manual_deletion_ts" is greater than 24 hours ago.
 * the associated media record does not have "is_deleted" = true
 * the associated transcription document record does not have "is_deleted" = true

Once the media and transcription document records are identified the following needs to done
 * find associated object via external_object_directory entries for the med_id or trd_id

----------------------------------------------- ONLY IMPLEMENT THE FOLLOWING 4 STEPS --------------------------------------------------------
 * *{color:#ff8b00}check if the associated object is successfully saved in ARM i.e. check the corresponding record in EOD with status :  "STORED" and externale_location_type(elt_id) of "ARM"{color}*
 * *{color:#ff8b00}Call ARM's UpdateMetadata for the object chosen for manual deletion.{color}*
 * *{color:#ff8b00}In Updatemetadata, pass the following parameter:{color}*
 ** *{color:#ff8b00}event_date should be calculated as follows:{color}*
 *** *{color:#ff8b00}For media :- media.deleted_ts minus 100 Years (event-date-adjustment-years from application.yaml){color}*
 *** *{color:#ff8b00}For Transcription Document :- transcription_document.deleted_ts minus 100 Years (event-date-adjustment-years from application.yaml){color}*
 ** *{color:#ff8b00}retn_conf_reason should be a JSON object as shown below (e.g. of Media is shown but the same logic need to be implemented for transcription_doc, annotation_doc or case_document):{color}*
 *** 
{code:java}
{

  "manual_deletion_ts": "2024-06-26 12:10:00", (Media -> Object_Admin_action -> mark_for_manual_del_ts)

  "ret_conf_reason": "manual deletion reason", (Media -> Object_Admin_action -> Object_hidden_reason)   

  "ticket_reference":"INC12345", (Media -> Object_Admin_action -> ticket_reference)
  
  "comments":"test" (Media -> Object_Admin_action -> comments)
} {code}
 

 * *{color:#ff8b00}delete the external_object_directory record for "ARM" record{color}*

-------------------------------------------------------------------------------------------------------------------------------------------------------
 * {color:#172b4d}delete the external_object_directory record for "Inbound" and "Unstructured" record{color}
 * set is_deleted = true on the media or transcription_document entry

 
h4. User Story

AS A

DARTS System

I WANT TO

notify ARM to delete a particular file following a file being marked for manual deletion in DARTS

SO THAT

any copy of the file being requested for manual deletion can be deleted in ARM if present. 
h4. Acceptance Criteria

GIVEN

all copies of an audio/transcript file requested for manual deletion must be deleted from the whole DARTS system

WHEN

a user manually requests for an audio/transcript file to be deleted

THEN

the metadata of the file must update the retention value of that file to today's date and forwarded to ARM, allowing ARM to believe that the retain until date of that file has now expired

AND

ARM will delete any copies that it holds of that file

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/darts/common/repository/ObjectAdminActionRepository.java
```diff
@@ -34,6 +34,8 @@ Optional<ObjectAdminActionEntity> findByTranscriptionDocument_IdAndObjectHiddenR
     @Query("""
         SELECT o FROM ObjectAdminActionEntity o
                 WHERE o.markedForManualDelDateTime < :deletionThreshold
+                AND ((o.media is not null and o.media.isDeleted = false)
```
karen-hedges, line 37: There are 2 other documents case document and annotation document. Should these also have the same conditions or are they not deleted?

src/main/java/uk/gov/hmcts/darts/task/service/impl/ManualDeletionProcessorImpl.java
```diff
@@ -100,15 +149,26 @@ private void deleteFromExternalDataStore(ExternalObjectDirectoryEntity externalO
...
         return action.getMedia() != null && !action.getMedia().isDeleted();
     }
 
-    public OffsetDateTime getDeletionThreshold() {
-        return OffsetDateTime.now().minus(Duration.parse("PT" + gracePeriod));
+    OffsetDateTime getDeletionThreshold() {
+        return OffsetDateTime.now().minus(gracePeriod);
+    }
+
+    static class RetConfReason {
```
hemantasharma1129, line 164: Any reason why the JSON structure is defined as inner class? Cant we have it as a separate class similar to CaseRetentionConfidenceReason? The both (old and new) JSON gets passed into the Updatemetadata;s confidence reason field. One is associated with Case retention and the new one is for Manual Deletion.

Ben-Edwards-cgi: Yeah this class is exclusively used within the ManualDeletionProcessor as we convert it into a string before we send it to other classes. As such it makes sense to keep the logic centralised as the RetConfReason inner class only has any meaning when combined with the ManualDeletionProcessor.

If this changes in the future I think we can pull it out but I do not see any value in doing so now as the class is meaningless without the ManualDeletion context. (So improves maintainability)

src/main/java/uk/gov/hmcts/darts/task/service/impl/ManualDeletionProcessorImpl.java
```diff
@@ -83,13 +97,48 @@ private void deleteTranscriptionDocument(UserAccountEntity userAccount, Transcri
...
+        try {
+            RetConfReason retConfReason = new RetConfReason();
+            retConfReason.manualDeletionTs = deletedTs;
+            retConfReason.manualDeletionReason = objectAdminAction.getObjectHiddenReason().getReason();
+            retConfReason.ticketReference = objectAdminAction.getTicketReference();
+            retConfReason.comments = objectAdminAction.getComments();
+            return objectMapper.writeValueAsString(retConfReason);
+        } catch (Exception e) {
+            log.error("Error while creating RetConfReason", e);
+            return "Error while creating RetConfReason";
```
karen-hedges, line 137: If there is an error here should this error be sent to ARM? Isn't this an internal issue?

src/main/java/uk/gov/hmcts/darts/task/service/impl/ManualDeletionProcessorImpl.java
karen-hedges: We already have a pattern of having the json objects separate from the calling classes. You are breaking this pattern by using an inner class

Ben-Edwards-cgi: The purpose of nested classes is to group classes that belong together, which makes your code more readable and maintainable and is standard practice in Java development.

As this class is exclusively used within the ManualDeletionProcessor and has no use / context outside of this class there no value in moving it into its own class as we lose the benefits outlined above.

If the existing pattern we are using goes against these standards the solution should be to refactor those classes instead of enforcing them to ensure we produce quality code

src/main/java/uk/gov/hmcts/darts/common/repository/ExternalObjectDirectoryRepository.java
```diff
@@ -65,6 +65,12 @@ List<ExternalObjectDirectoryEntity> findByMediaStatusAndType(MediaEntity media,
     List<ExternalObjectDirectoryEntity> findByTranscriptionDocumentEntityAndExternalLocationType(TranscriptionDocumentEntity transcriptionDocument,
                                                                                                  ExternalLocationTypeEntity externalLocationType);
 
+
+    List<ExternalObjectDirectoryEntity> findByTranscriptionDocumentEntityAndExternalLocationTypeAndStatus(
```
karen-hedges, line 69: I forgot to mention that you need to add integration tests for all new repository methods. I think its in the best practises page.

Ben-Edwards-cgi: This is spring native query so it would have been tested apart of the spring framework, but for custom queries I would agree

karen-hedges: So you will need to add repository integration tests for your changes to ObjectAdminActionRepository

Ben-Edwards-cgi: I must have missed that but yes agreed will add now :)

Ben-Edwards-cgi: There were existing tests that cover this but i have added additional ones to cover isDeleted = true scenarios + where we only have media or transcription documents not both

src/main/java/uk/gov/hmcts/darts/task/service/impl/ManualDeletionProcessorImpl.java
hemantasharma1129: We should not pass any random error message to ARM if the agreed contract is a JSON.
The entire process should be Transactional i.e. If it fails in any of the step, the entire manual deletion process should fail for that object. This way the isDeleted should stay FALSE and will be picked up in the next execution. 
The update to Media isDeleted should happen at the end.

src/main/java/uk/gov/hmcts/darts/task/service/impl/ManualDeletionProcessorImpl.java
hemantasharma1129: Does it not increase the line of Code of a main class by adding a inner class for an object which defines a Json structure?

Ben-Edwards-cgi: The class it is added in is a relatively small in nature anyway so the additional  9 lines of code wont have any impact but it will make maintainability easier as the inner class has no meaning outside of the ManualDeletion class so having this logic in the same class helps provide context

src/main/java/uk/gov/hmcts/darts/task/service/impl/ManualDeletionProcessorImpl.java
Ben-Edwards-cgi: I have updated this code to chuck an exception so any existing error handling logic can kick in

src/main/java/uk/gov/hmcts/darts/task/service/impl/ManualDeletionProcessorImpl.java
hemantasharma1129: I think this is deviating from what we have done already and introducing a new pattern. If we are changing the code i.e. maintaining the JSON template as inner class rather than a separate class, then we will end up introducing 2 different patter in the code. I dont see any issue with the existing pattern.

hemantasharma1129: How come maintaining a JSON template in a separate class, which makes it reusable in future and reduces the line of code in main class, is a TechDebt? I do agree, from encapsulation point of view, you want to maintain everything in the same class but its more of a trade off. Considering Easy maintenance, More readable (if the Json template grows), Reusable etc we should have a separate class.

src/main/java/uk/gov/hmcts/darts/task/service/impl/ManualDeletionProcessorImpl.java
hemantasharma1129: Does the existing error handling rollback the transaction? If the ARM call fails for any reason, the isdeleted flag in Media should be rollback so that it gets picked up again in next iteration.

Ben-Edwards-cgi: Looks like there was no exception handling in this class before I have added one and added some transactions to ensure this gets handled correctly

hemantasharma1129 (approved):

karen-hedges (approved):

PR #2342, opened 2024-12-04
DMP-4256 CR and IL files are not getting deleted

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-4256

### Change description ###

Added more tests and updated the bash script to set the status in ARM response files. Also added code to delete the response files if they are invalid json

```
[ ] Yes
[ x] No
```

### DIFF
--- a/bin/generateBatchArmResponses.sh
+++ b/bin/generateBatchArmResponses.sh
@@ -46,7 +46,7 @@ addNewResponses () {
   fi
 
   #${parameter//pattern/string}
-  uiFileContentsParam="{\"operation\": \"upload_file\", \"transaction_id\": \"2d1c7f6f-224e-768e-a274-41af570e6502\", \"relation_id\": \"EODID\", \"a360_record_id\": \"1cf976c7-cedd-703f-ab70-01588bd56d50\", \"process_time\": \"2023-07-11T11:39:26.790000\", \"status\": 1, \"input\": \"{\\\"operation\\\": \\\"create_record\\\",\\\"relation_id\\\": \\\"EODID\",\\\"record_metadata\\\": {\\\"record_class\\\": \\\"A360TEST\",\\\"publisher\\\": \\\"A360\\\",\\\"recordDate\\\": \\\"2016-11-22T11:39:30Z\\\",\\\"region\\\": \\\"GBR\",\\\"title\\\": \\\"A360230711_TestIngestion_2\\\"}}\", \"exception_description\": null, \"errorStatus1\": null}"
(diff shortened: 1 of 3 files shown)


hemantasharma1129 (approved):

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
```diff
@@ -534,6 +568,16 @@ private void readUploadFile(BinaryData uploadFileBinary, UploadFileFilenameProce
         }
     }
 
+    private ArmResponseUploadFileRecord getResponseUploadFileRecordOrDelete(Path jsonPath) throws IOException {
```
chrisbellingham-hmcts, line 571: Just a thought, but given there are 3 very similar versions of this method, consider genericising it to avoid the duplication. I.e is there enough commonality for all 3 to be replaced with a single method with the following signature?
```java
private <T> T getRecordOrDelete(Path jsonPath, Class<T> clazz, String errorMessage) throws IOException {
```

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
```diff
@@ -328,6 +329,29 @@ && nonNull(armResponseBatchData.getUploadFileFilenameProcessor())) {
...
+                     armResponseBatchData.getUploadFileFilenameProcessor().getUploadFileFilenameAndPath());
+        }
+        if (CollectionUtils.isNotEmpty(armResponseBatchData.getInvalidLineFileFilenameProcessors())) {
+            armResponseBatchData.getInvalidLineFileFilenameProcessors().forEach(
+                invalidLineFileFilenameProcessor -> log.info("Found eod {} with invalid line file: {}",
+                                                             armResponseBatchData.getExternalObjectDirectoryId(),
+                                                             invalidLineFileFilenameProcessor.getInvalidLineFileFilenameAndPath())
+            );
+        }
+    }
```
chrisbellingham-hmcts, line 353: Could some of the `armResponseBatchData` attributes be assigned to variables, to avoid the repeated calls to their getters? I think it would help reduce the verbosity of the code somewhat.

bin/generateBatchArmResponses.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
 echo "This script generates the ARM pull response files. These can be used for testing where ARM returns a response for an object such as media, transcription document, annotation document or case document to be stored in the ARM storage"
 
-uuid1=$(uuidgen)
+uuid1=$(uuidgen | tr -d '-')
+##uuid1=${uuid1//'-'/''}
```
chrisbellingham-hmcts, line 5: Intentional to leave the commented line here?

karen-hedges: doh

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
karen-hedges: There is a ticket to refactor this class as its nearly 1000 lines long so might be best left for that?

src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
karen-hedges: There is a ticket to refactor this class. It might be bests its done then.

chrisbellingham-hmcts (approved):

PR #2352, opened 2024-12-05
DMP-4461: ArmMissingResponseFile Task not processing any EOD records

Ben-Edwards-cgi:
### Change description ###

Introduces several updates primarily focused on refining logging practices within the codebase. Key changes include adjustments to the logging prefix, modifications to the constructor parameters, and consistent application of the new logging prefix in various logging statements.

## Highlights

- **Logging Prefix Update**: 
  - A new variable  `loggingPrefix` was added to replace instances where `manifestFilePrefix`  was used for log prefixs.

- **Constructor Modifications**: 
  - Constructors in several classes have been updated to accept an additional `loggingPrefix` parameter, streamlining the logging process.

- **Log Statement Changes**: 
  - Numerous logging statements have been adjusted to utilize the new `loggingPrefix` for consistency and clarity.

- **Test Adjustments**: 
  - Corresponding updates were made in test classes to ensure they align with changes in constructors and logging practices.

- **String Formatting Fix**: 
  - One specific change in `ArmResponseFileHelper.java` corrected the string formatting in a log warning statement.

These modifications aim to enhance the maintainability of the code and improve the clarity of log outputs across the application.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmMissingResponseCleanupImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmMissingResponseCleanupImpl.java
@@ -36,7 +36,7 @@ public class ArmMissingResponseCleanupImpl extends BatchCleanupArmResponseFilesS
                                          ArmResponseFileHelper armResponseFileHelper) {
         super(externalObjectDirectoryRepository, objectRecordStatusRepository, externalLocationTypeRepository, armDataManagementApi, userIdentity,
               batchCleanupConfiguration, armDataManagementConfiguration, currentTimeHelper, armResponseFileHelper,
-              "ArmMissingResponseCleanup");
+              "", "ArmMissingResponseCleanup");
         this.armRawDataFailed = objectRecordStatusRepository.getReferenceById(ARM_RAW_DATA_FAILED.getId());
     }
 
--- a/src/main/java/uk/gov/hmcts/darts/arm/service/impl/BatchCleanupArmResponseFilesServiceCommon.java
+++ b/src/main/java/uk/gov/hmcts/darts/arm/service/impl/BatchCleanupArmResponseFilesServiceCommon.java
@@ -50,6 +50,21 @@ public class BatchCleanupArmResponseFilesServiceCommon implements BatchCleanupAr
     protected final String manifestFilePrefix;
     protected final String loggingPrefix;
 
+    public BatchCleanupArmResponseFilesServiceCommon(ExternalObjectDirectoryRepository externalObjectDirectoryRepository,
+                                                     ObjectRecordStatusRepository objectRecordStatusRepository,
+                                                     ExternalLocationTypeRepository externalLocationTypeRepository,
+                                                     ArmDataManagementApi armDataManagementApi,
+                                                     UserIdentity userIdentity,
+                                                     ArmBatchCleanupConfiguration batchCleanupConfiguration,
+                                                     ArmDataManagementConfiguration armDataManagementConfiguration,
+                                                     CurrentTimeHelper currentTimeHelper,
+                                                     ArmResponseFileHelper armResponseFileHelper,
+                                                     String manifestFilePrefix) {
+        this(externalObjectDirectoryRepository, objectRecordStatusRepository, externalLocationTypeRepository, armDataManagementApi, userIdentity,
+             batchCleanupConfiguration, armDataManagementConfiguration, currentTimeHelper, armResponseFileHelper,
+             manifestFilePrefix, manifestFilePrefix);
+    }
+
 
     @Override
     public void cleanupResponseFiles(int batchsize) {
--- a/src/main/java/uk/gov/hmcts/darts/arm/service/impl/BatchCleanupArmResponseFilesServiceCommon.java
+++ b/src/main/java/uk/gov/hmcts/darts/arm/service/impl/BatchCleanupArmResponseFilesServiceCommon.java
@@ -48,18 +48,19 @@ public class BatchCleanupArmResponseFilesServiceCommon implements BatchCleanupAr
     protected final CurrentTimeHelper currentTimeHelper;
     protected final ArmResponseFileHelper armResponseFileHelper;
     protected final String manifestFilePrefix;
+    protected final String loggingPrefix;
 
 
     @Override
     public void cleanupResponseFiles(int batchsize) {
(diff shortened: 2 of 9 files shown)


src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmMissingResponseCleanupImpl.java
```diff
@@ -36,7 +36,7 @@ public ArmMissingResponseCleanupImpl(ExternalObjectDirectoryRepository externalO
                                          ArmResponseFileHelper armResponseFileHelper) {
         super(externalObjectDirectoryRepository, objectRecordStatusRepository, externalLocationTypeRepository, armDataManagementApi, userIdentity,
               batchCleanupConfiguration, armDataManagementConfiguration, currentTimeHelper, armResponseFileHelper,
-              "ArmMissingResponseCleanup");
+              "", "ArmMissingResponseCleanup");
```
hemantasharma1129, line 39: By Passing a empty prefix, will not fail the validation in ArmResponseFileHelper.getCorrespondingArmFilesForManifestFilename? Cleanup logic during ARM Missing response replay flow should be independent of DETS or DARTS i.e. any manifest files corresponding to the EOD should get deleted irrespective of whether its DETS or DARTs.

Ben-Edwards-cgi: Lets have a chat around this :).

chrisbellingham-hmcts (commented):
Hey Ben. No comments from me on the content here, but as this appears to be a non-functional (logging) change, is this only intended to be an intermediate PR on the road to resolving https://tools.hmcts.net/jira/browse/DMP-4461?

Asking because it's not obvious how these changes correspond to the ticket which reports an issue with EOD records not being processed.

Suggest refocussing the PR description on the intent rather than a summary of what's changed, as I think the former is more useful to reviewers.

chrisbellingham-hmcts (approved):

src/main/java/uk/gov/hmcts/darts/arm/service/impl/DetsBatchCleanupArmResponseFilesServiceImpl.java
```diff
@@ -17,16 +17,16 @@ public class DetsBatchCleanupArmResponseFilesServiceImpl extends BatchCleanupArm
...
+                                                       ExternalLocationTypeRepository externalLocationTypeRepository,
+                                                       ArmDataManagementApi armDataManagementApi,
+                                                       UserIdentity userIdentity, ArmBatchCleanupConfiguration batchCleanupConfiguration,
+                                                       ArmDataManagementConfiguration armDataManagementConfiguration, CurrentTimeHelper currentTimeHelper,
+                                                       ArmResponseFileHelper armResponseFileHelper,
+                                                       @Value("${darts.storage.dets.dets-manifest-file-prefix}") String manifestFilePrefix) {
         super(externalObjectDirectoryRepository, objectRecordStatusRepository, externalLocationTypeRepository, armDataManagementApi, userIdentity,
               batchCleanupConfiguration, armDataManagementConfiguration, currentTimeHelper, armResponseFileHelper,
-              manifestFilePrefix
+              manifestFilePrefix, manifestFilePrefix
```
karen-hedges, line 29: Instead of passing the same object twice to hack around the code so you can reuse, it would be better to remove the manifest file prefix into an abstract method so each inherited class can pass the correct value

Ben-Edwards-cgi: A method feels wrong for this, as we want it to be a constant value which we can not guarantee with a method, but perhaps a overloaded constructor would be better

Ben-Edwards-cgi:
> Hey Ben. No comments from me on the content here, but as this appears to be a non-functional (logging) change, is this only intended to be an intermediate PR on the road to resolving https://tools.hmcts.net/jira/browse/DMP-4461?
> 
> Asking because it's not obvious how these changes correspond to the ticket which reports an issue with EOD records not being processed.
> 
> Suggest refocussing the PR description on the intent rather than a summary of what's changed, as I think the former is more useful to reviewers.

This change fixes the root issue as the manifestFilePrefix is used as both a logging prefix and in some internal methods.
The fix essentially sets this prefix to be "" for this task (As it needs to process all file name prefix's not just DARTs or DETs) so that we can hit the logic that is wrapped in a fileName.startsWith(prefix) correctly for this method which was incorrectly being skipped before

src/main/java/uk/gov/hmcts/darts/arm/service/impl/BatchCleanupArmResponseFilesServiceCommon.java
```diff
@@ -48,18 +48,34 @@ public class BatchCleanupArmResponseFilesServiceCommon implements BatchCleanupAr
...
+                                                     ArmDataManagementApi armDataManagementApi,
+                                                     UserIdentity userIdentity,
+                                                     ArmBatchCleanupConfiguration batchCleanupConfiguration,
+                                                     ArmDataManagementConfiguration armDataManagementConfiguration,
+                                                     CurrentTimeHelper currentTimeHelper,
+                                                     ArmResponseFileHelper armResponseFileHelper,
+                                                     String manifestFilePrefix) {
+        this(externalObjectDirectoryRepository, objectRecordStatusRepository, externalLocationTypeRepository, armDataManagementApi, userIdentity,
+             batchCleanupConfiguration, armDataManagementConfiguration, currentTimeHelper, armResponseFileHelper,
+             manifestFilePrefix, manifestFilePrefix);
```
karen-hedges, line 65: If you make the manifest prefix an abstract method and in the parent classes they pass the appropriate prefix you wont need to pass them in the constructor

Ben-Edwards-cgi: A method feels wrong for this, as we want it to be a constant value which we can not guarantee with a method. I have updated the code to use overloaded constructors to bypass the need to pass in the value twice for methods this

Ben-Edwards-cgi: @hemantasharma1129  @jackmaloney  what are your views?

PR #2357, opened 2024-12-06
DMP-4351 Do not change is_current status on legacy events

jackmaloney:
https://tools.hmcts.net/jira/browse/DMP-4351

### Change description ###
Legacy events should remain with the is_current flag set to the same status as they were migrated as. This change skips events where the original created date is earlier than the modernised go live date. 

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/event/service/impl/CleanupCurrentFlagEventProcessorImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/event/service/impl/CleanupCurrentFlagEventProcessorImpl.java
@@ -1,9 +1,9 @@
 package uk.gov.hmcts.darts.event.service.impl;
 
 import lombok.Getter;
-import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections4.CollectionUtils;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.scheduling.annotation.Async;
 import org.springframework.stereotype.Component;
 import uk.gov.hmcts.darts.common.entity.EventEntity;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/darts/event/service/impl/CleanupCurrentFlagEventProcessorImpl.java
```diff
@@ -52,6 +68,13 @@ private void supersedeOldEvents(EventRepository.EventIdAndHearingIds eventIdAndH
         if (CollectionUtils.isNotEmpty(eventsToBeSuperseded)) {
             List<Integer> eveIdsThatHaveBeenSuperseded = new ArrayList<>();
             for (EventEntity eventToBeSuperseded : eventsToBeSuperseded) {
+                if (eventToBeSuperseded.getCreatedDateTime().isBefore(earliestIsCurrentClearUpDate)) {
+                    log.info("Event version with event id {} received into modernised DARTS is for a legacy event. Skipping event is_current cleanup " +
+                                 "for eve_id {}",
+                         eventIdAndHearingIds.getEventId(),
+                         eventIdAndHearingIds.getEveId());
+                    continue;
```
Ben-Edwards-cgi, line 76: I think we also need to exclude these from the SQL calls else we always pull back the same data and skip it in code

davet1985: @jackmaloney as discussed, I think there's no point trying to put that logic into the query for the tiny amount of times this is going to run, as we get the benefit of being able to log doing it this way. Happy with this how it is 👍

Ben-Edwards-cgi: Looking at this again as this only runs for event_ids that we get into the system and not in bulk I am also happy with the above :) as after the first few days of go live we should not get any events that hit legacy data

Ben-Edwards-cgi (approved):

PR #2376, opened 2024-12-12
Using pre-built images in docker-compose-local.yml

davet1985:

N/A

### Change description ###

- removing part of `bin/dcup` script that builds local images

```
[ ] Yes
[ ] No
```

### DIFF
--- a/docker-compose-local.yml
+++ b/docker-compose-local.yml
@@ -67,16 +67,14 @@ services:
       - darts-network
 
   darts-gateway:
-    container_name: darts-gateway
-    image: darts-gateway:master
+    image: sdshmctspublic.azurecr.io/darts/gateway:latest
     ports:
       - "8070:8070"
     networks:
       - darts-network
 
   darts-stub-services:
-    container_name: darts-stub-services
-    image: darts-stub-services:master
+    image: sdshmctspublic.azurecr.io/darts/stub-services:latest
     ports:
       - "4551:4551"
     networks:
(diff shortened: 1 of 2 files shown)


docker-compose-local.yml
```diff
@@ -118,7 +116,7 @@ services:
     environment:
       ALLOW_CONFIG_MUTATIONS: true
       DARTS_PORTAL_URL: http://localhost:3000
-      DARTS_API_URL: http://localhost:4550
+      DARTS_API_URL: http://darts-api:4550
```
chrisbellingham-hmcts, line 119: Yeah this probably makes more sense considering `darts-api` is part of the stack here. I guess we can override the value on the command line somehow if we wish to run against localhost darts-api (i.e when running darts-api in the IDE)?

chrisbellingham-hmcts (approved):

docker-compose-local.yml
davet1985: Due to the way docker works with networks, it's not really possible to get a running container to communicate with a service running on the host machine. Unless you have some ideas here?

chrisbellingham-hmcts: Ah yes you're right. No worries, we can just run darts-portal with yarn as a process on the host if we need to run against darts-api on the host.

davet1985: Yeah indeed, that's how I do it. Or possibly if you run the darts-portal in docker directly it might be work because the networking is different.

jackmaloney (approved):

PR #2382, opened 2024-12-13
DMP-3534: Retention Confidence

chrisbellingham-hmcts:
## [DMP-3534](https://tools.hmcts.net/jira/browse/DMP-3534)

Two fundamental changes in this PR:

1. The numerical values associated with retn_conf_score have been altered per the ticket requirement. See `src/main/java/uk/gov/hmcts/darts/retention/enums/RetentionConfidenceScoreEnum.java`
2. The retention confidence reason and score that are set on a case are now driven by the provided confidence category. A DB lookup is performed on `retention_confidence_category_mapper` to obtain these values. See `src/main/java/uk/gov/hmcts/darts/retention/service/impl/RetentionServiceImpl.java`

Some additional work has been done to facilitate these changes:
* `RetentionConfidenceCategoryMapperEntity` has been updated to utilise enums rather than ints, improving code readability and aligning with the approach taken for `CaseRetentionEntity` and `CourtCaseEntity`. 
* The logic that performs the category lookup and that sets the confidence attributes on a court case has been centralised in `RetentionServiceImpl`, and calling code has been adjusted to share this common code.
* Test data classes have been created to allow for easy creation of `RetentionConfidenceCategoryMapperEntity` objects for both unit and integration tests going forward.

The changes must be accompanied by a manual data fix upon merge, which should be applied to staging, test, demo and ITHC.

```sql
UPDATE annotation_document SET ret_conf_score=0 WHERE ret_conf_score=1;
UPDATE annotation_document SET ret_conf_score=1 WHERE ret_conf_score=2;
UPDATE case_document SET ret_conf_score=0 WHERE ret_conf_score=1;
UPDATE case_document SET ret_conf_score=1 WHERE ret_conf_score=2;
UPDATE court_case SET ret_conf_score=0 WHERE ret_conf_score=1;
UPDATE court_case SET ret_conf_score=1 WHERE ret_conf_score=2;
UPDATE media SET ret_conf_score=0 WHERE ret_conf_score=1;
UPDATE media SET ret_conf_score=1 WHERE ret_conf_score=2;
UPDATE retention_confidence_category_mapper SET ret_conf_score=0 WHERE ret_conf_score=1;
UPDATE retention_confidence_category_mapper SET ret_conf_score=1 WHERE ret_conf_score=2;
UPDATE transcription_document SET ret_conf_score=0 WHERE ret_conf_score=1;
UPDATE transcription_document SET ret_conf_score=1 WHERE ret_conf_score=2; 
```

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/darts/cases/service/impl/CloseOldCasesProcessorImplTest.java
```diff
@@ -94,16 +92,17 @@ void closeCases() {
...
+
+        when(retentionApi.updateCourtCaseConfidenceAttributesForRetention(any(), eq(RetentionConfidenceCategoryEnum.AGED_CASE)))
+            .thenReturn(courtCase);
 
         // when
         closeOldCasesProcessor.closeCases(2);
 
         // then
         assertTrue(courtCase.getClosed());
-        assertEquals(CURRENT_DATE_TIME, courtCase.getRetConfUpdatedTs());
```
chrisbellingham-hmcts, line 106: Author note: The class under test no longer has responsibility for setting this attribute, therefore the assertion is removed here. Dedicated unit tests exist in `RetentionServiceImplTest` to verify this attribute is set correctly.

src/test/java/uk/gov/hmcts/darts/retention/service/impl/RetentionPostServiceImplTest.java
```diff
@@ -305,11 +303,6 @@ void ok_BeforeCurrentRetentionDate_Judge() {
         assertEquals(10, savedRetention.getCreatedBy().getId());
         assertEquals(RetentionPolicyEnum.MANUAL.getPolicyKey(), savedRetention.getRetentionPolicyType().getFixedPolicyKey());
         assertEquals(MANUAL_OVERRIDE, savedRetention.getConfidenceCategory());
-        verify(caseRepository).save(courtCaseEntityArgumentCaptor.capture());
-        CourtCaseEntity savedCourtCase = courtCaseEntityArgumentCaptor.getValue();
-        assertEquals(RetentionConfidenceReasonEnum.MANUAL_OVERRIDE, savedCourtCase.getRetConfReason());
-        assertEquals(RetentionConfidenceScoreEnum.CASE_PERFECTLY_CLOSED, savedCourtCase.getRetConfScore());
-        assertEquals(CURRENT_DATE_TIME, savedCourtCase.getRetConfUpdatedTs());
```
chrisbellingham-hmcts, line 312: Author note: The class under test no longer has responsibility for setting these attributes, therefore the assertions are removed here. Dedicated unit tests exist in `RetentionServiceImplTest` to verify these attributes are set correctly.

Similar comment applies to the other tests in this test class.

src/main/java/uk/gov/hmcts/darts/retention/enums/RetentionConfidenceScoreEnum.java
```diff
@@ -7,7 +7,7 @@
 @RequiredArgsConstructor
 public enum RetentionConfidenceScoreEnum {
 
-    CASE_PERFECTLY_CLOSED(1), CASE_NOT_PERFECTLY_CLOSED(2);
+    CASE_PERFECTLY_CLOSED(0), CASE_NOT_PERFECTLY_CLOSED(1);
```
chrisbellingham-hmcts, line 10: Should this PR include a flyway script to migrate `1` -> `0`, and `2` -> `1` for `court_case.ret_conf_score`, otherwise we'll have problems with the existing case data being wrongly interpreted?

src/integrationTest/java/uk/gov/hmcts/darts/cases/service/CloseOldCasesProcessorTest.java
```diff
@@ -95,8 +120,50 @@ void givenClosedEventsUseDateAsClosedDate() {
...
+        // And
+        HearingEntity hearing = dartsDatabase.createHearing("a_courthouse", "1", "1078", LocalDateTime.now().minusYears(7).plusMonths(3));
+
+        OffsetDateTime closeDate = OffsetDateTime.now().minusYears(7);
+
+        EventEntity eventEntity1 = dartsDatabase.getEventStub().createEvent(hearing, 8);//Re-examination
+        eventEntity1.setCreatedDateTime(OffsetDateTime.now().minusYears(7).plusDays(10));
+        EventEntity eventEntity2 = dartsDatabase.getEventStub().createEvent(hearing, 214);//case closed
+        eventEntity2.setCreatedDateTime(closeDate);
+        EventEntity eventEntity3 = dartsDatabase.getEventStub().createEvent(hearing, 23);//Application: No case to answer
```
jackmaloney, line 136: What does comment on this line mean?

src/integrationTest/java/uk/gov/hmcts/darts/cases/service/CloseOldCasesProcessorTest.java
```diff
@@ -95,8 +120,50 @@ void givenClosedEventsUseDateAsClosedDate() {
...
+    @Test
+    void closeCases_shouldCloseCases_andUseDateAsClosedDate_andSetNullConfidenceReasonAndScore_whenNoConfidenceMappingExists() {
+        // Given createAndSaveRetentionConfidenceCategoryMappings() is not invoked, so no confidence mappings exist in the DB
+
+        // And
+        HearingEntity hearing = dartsDatabase.createHearing("a_courthouse", "1", "1078", LocalDateTime.now().minusYears(7).plusMonths(3));
+
+        OffsetDateTime closeDate = OffsetDateTime.now().minusYears(7);
+
+        EventEntity eventEntity1 = dartsDatabase.getEventStub().createEvent(hearing, 8);//Re-examination
```
jackmaloney, line 132: What does comment on this line mean?

src/main/java/uk/gov/hmcts/darts/retention/enums/RetentionConfidenceScoreEnum.java
jackmaloney: Were these missed previously? I can't see any entries in the database in the retention_confidence_category_mapper table. Don't think what you have done will work without those?

chrisbellingham-hmcts: `retention_confidence_category_mapper` is indeed empty, feedback from @hemantasharma1129 is that the mapping table will be populated as a separate task. If a mapping from a category doesn't exist (e.g. because the `retention_confidence_category_mapper` table is empty), then `null` values will get populated for score and reason against the case.

`court_case.ret_conf_score` however has numerical values already populated for various cases in the DB, and my understanding is we're changing the meaning of those numerical values in this PR. So any existing case records that have `ret_conf_score=1` would have intended that to mean `CASE_PERFECTLY_CLOSED`. However when this PR goes in, they will be interpreted as `CASE_NOT_PERFECTLY_CLOSED`, which seems dangerous?

src/integrationTest/java/uk/gov/hmcts/darts/cases/service/CloseOldCasesProcessorTest.java
chrisbellingham-hmcts: I will review the comments in this test method, I have copy/pasted an existing test setup to test a slightly different scenario here.

src/integrationTest/java/uk/gov/hmcts/darts/cases/service/CloseOldCasesProcessorTest.java
```diff
@@ -95,8 +120,50 @@ void givenClosedEventsUseDateAsClosedDate() {
         assertEquals(RetentionConfidenceCategoryEnum.AGED_CASE, caseRetentionEntity.getConfidenceCategory());
     }
 
+    @Test
+    void closeCases_shouldCloseCases_andUseDateAsClosedDate_andSetNullConfidenceReasonAndScore_whenNoConfidenceMappingExists() {
+        // Given createAndSaveRetentionConfidenceCategoryMappings() is not invoked, so no confidence mappings exist in the DB
```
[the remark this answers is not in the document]
chrisbellingham-hmcts: `createAndSaveRetentionConfidenceCategoryMappings()` is a private method in the test class, which is invoked as a data setup step in most of the other tests to create some mappings in `retention_confidence_category_mapper`. For this case however, we explicitly don't want to insert any mappings, so the comment just highlights the lack of mappings in the DB.

src/main/java/uk/gov/hmcts/darts/retention/enums/RetentionConfidenceScoreEnum.java
hemantasharma1129: @chrisbellingham-hmcts yes we can introduce a flyway script to update 1 -> 0 and 2 -> 1 any existing data in DB.

@jackmaloney retention_confidence_category_mapper is part of Retention migration Path to Live item 36. This is agreed that RCC development will be completed from our end so that nothing is pending  and test RCC once the mapper table is populated by migration. As there is no data in the mapper table, no reasons will be populated. I will check if a  Test ticket exist, if not I will create one.

chrisbellingham-hmcts: Flyway script added in commit https://github.com/hmcts/darts-api/pull/2382/commits/7546b060012ca5ed516a776f2b61b2f8ea26e62d for all impacted tables including:

- `annotation_document`
- `case_document`
- `court_case`
- `media`
- `retention_confidence_category_mapper` (this table should be empty, but it's good form to include the migration anyway)
- `transcription_document`

jackmaloney: Are we saying that this ticket cannot be tested util migration do that @hemantasharma1129 ? Why do we need migration team to populate this table? We have other reference data tables that application team are the owners of that migration team won't migrate, can we not do this for this table?

hemantasharma1129: This ticket is tested with hardcoded value already. Those hardcoded values need to be persisted so that migration can also refer to it and be in sync. The reason we cant populate these values because we dont know what the ret_conf_reason values would be. Only key we have is the Confidence Category (1,2 & 3). Judith did an analysis which Jonathan will continue. This table will be populated part of Retention migration which is due in early Jan 2025.

jackmaloney: Confirmed this will need to be done in new year by migration so i'll close this one.

jackmaloney (approved):

PR #2391, opened 2024-12-18
DMP-2542 Unlink expired cases from media

jackmaloney:
https://tools.hmcts.net/jira/browse/DMP-2542

### Change description ###
When a case is expired, the link to the media should be removed but only if all other associated cases have also expired. 

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/task/runner/impl/CaseExpiryDeletionAutomatedTask.java
```diff
@@ -30,6 +31,7 @@ public class CaseExpiryDeletionAutomatedTask
 
     private final CurrentTimeHelper currentTimeHelper;
     private final DataAnonymisationService dataAnonymisationService;
+    private final HearingsService hearingsService;
```
chrisbellingham-hmcts, line 34: It's preferable to avoid directly injecting services from other modules.  Suggest creating a `HearingApi` interface (and impl) in `src/main/java/uk/gov/hmcts/darts/hearings/api`, and a `removeMediaLinkToHearing(int courtCaseId)` method that simply delegates to the needed functionality in `HearingService`.

You can then inject `HearingApi` here , and call `hearingApi.removeMediaLinkToHearing(courtCaseId)` on L70.

src/test/java/uk/gov/hmcts/darts/hearings/service/impl/HearingsServiceImplTest.java
```diff
@@ -145,6 +152,54 @@ void testGetEventsResponseCaseIsExpired() {
         assertThat(exception.getMessage()).isEqualTo("Case has expired.");
     }
 
+    @Test
+    void testRemoveMediaLinkToHearing() {
```
chrisbellingham-hmcts, line 156: Preferably go with the `methodName_expectations_conditions()` naming convention, as documented in [Best Practices](https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=1694706043#BestPractices(Backend)-TestNamingShallClearlyCommunicatetheTestIntent).

Comment applies to all newly added tests in this PR.

src/main/java/uk/gov/hmcts/darts/common/repository/HearingRepository.java
```diff
@@ -39,6 +39,14 @@ public interface HearingRepository extends JpaRepository<HearingEntity, Integer>
     )
     List<Integer> findHearingIdsByEventId(Integer eventId);
 
+    @Query("""
+        SELECT h FROM HearingEntity h
+        JOIN h.mediaList media
+        WHERE media.id = :mediaId
+        """
+    )
+    Optional<List<HearingEntity>> findHearingIdsByMediaId(Integer mediaId);
```
chrisbellingham-hmcts, line 48: A non-null result should always be returned as the collection would simply be empty if the query doesn't match any records. So I don't think the Optional wrapper adds any value/safety? 

See [Best Practices](https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=1694706043#BestPractices(Backend)-RepositoryMethodsShouldReturnOptional<>ForAllNon-CollectionReturnTypes) for more details.

chrisbellingham-hmcts (approved):

hemantasharma1129 (approved):

PR #2398, opened 2024-12-20
DMP-4386 Handle zero second audio on preview

jackmaloney:
https://tools.hmcts.net/jira/browse/DMP-4386

### Change description ###
On audio preview, if the audio start and end date is equal (or the start date is after the end date) then throw a 422 error. 

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/audio/service/AudioPreviewTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/audio/service/AudioPreviewTest.java
@@ -18,7 +18,6 @@ import java.time.LocalDate;
 import static java.util.Objects.nonNull;
 import static org.assertj.core.api.Assertions.assertThat;
 import static uk.gov.hmcts.darts.audio.enums.AudioPreviewStatus.ENCODING;
-import static uk.gov.hmcts.darts.audio.enums.AudioPreviewStatus.FAILED;
 import static uk.gov.hmcts.darts.audio.enums.AudioPreviewStatus.READY;
 import static uk.gov.hmcts.darts.test.common.AwaitabilityUtil.waitForMax10SecondsWithOneSecondPoll;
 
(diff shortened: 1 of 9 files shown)


src/integrationTest/java/uk/gov/hmcts/darts/audio/service/AudioPreviewTest.java
```diff
@@ -75,17 +75,6 @@ void audioPreviewEventuallyBecomesReady() {
         });
     }
 
-    @Test
-    void audioPreviewEventuallyFailsWhenMediaDoesntExist() {
```
jackmaloney, line 79: Author note: This test was removed as it's not actually possible to hit this scenario. A previous change carried out in DMP-4304 checks media exists and if not throws an error. This was done in the controller layer, and therefore this service layer test was not picked up as part of that change. I've moved the validation to the service layer (AudioPreview) as believe controllers should be kept as skinny as possible and any real logic should not be in controllers. This caused this test to fail as this class now checks that media exists. As we have a test covering this scenario in the AudioControllerPreviewIntTest, this can be safely removed.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioPreviewServiceImpl.java
```diff
@@ -15,9 +16,15 @@
 public class AudioPreviewServiceImpl implements AudioPreviewService {
 
     private final CachingAudioPreviewEncoder cachingAudioPreviewEncoder;
+    private final MediaIdValidator mediaIdValidator;
 
     @Override
     public AudioPreview getOrCreateAudioPreview(Integer mediaId) {
+        // return 404 if the media is hidden
```
chrisbellingham-hmcts, line 23: Consider removing these comments are they are susceptible to becoming out of sync with the actual behaviour - if the underlying validator logic (or returned HTTP code) is ever changed there's a good chance the dev won't update these comments in step.

chrisbellingham-hmcts (approved):

davet1985 (approved):

hemantasharma1129 (approved):

PR #2408, opened 2025-01-02
DMP-4385 Filter out 0 second audios from ATS processing

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-4385

### Change description ###

Updated the ATS filtering to filter out audio files that have less than a second start and end time, start or end time is null or start time is after end time

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
@@ -101,10 +101,13 @@ public class OutboundFileProcessorImpl implements OutboundFileProcessor {
         List<AudioFileInfo> concatenatedAndMergedAudioFileInfos = new ArrayList<>();
         if (isNotEmpty(audioFileInfos)) {
             // Used for logging only
-            String audioFilenames = audioFileInfos.stream().map(AudioFileInfo::getMediaFile).collect(Collectors.joining(", "));
+            String audioFilenames = null;
+            if (log.isDebugEnabled()) {
+                audioFileInfos.stream().map(AudioFileInfo::getMediaFile).collect(Collectors.joining(", "));
+            }
 
             List<ChannelAudio> concatenationsList = new ArrayList<>();
-            
+
             if (isWellFormedAudio(audioFileInfos)) {
                 log.debug("Audio files {} are well formed", audioFilenames);
                 List<ChannelAudio> concatenatedAudios = concatenateByChannelWithGaps(audioFileInfos);
--- a/src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
@@ -103,7 +103,7 @@ public class OutboundFileProcessorImpl implements OutboundFileProcessor {
             // Used for logging only
             String audioFilenames = null;
             if (log.isDebugEnabled()) {
-                audioFileInfos.stream().map(AudioFileInfo::getMediaFile).collect(Collectors.joining(", "));
+                audioFilenames = audioFileInfos.stream().map(AudioFileInfo::getMediaFile).collect(Collectors.joining(", "));
             }
 
             List<ChannelAudio> concatenationsList = new ArrayList<>();
--- a/src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
@@ -104,7 +104,6 @@ public class OutboundFileProcessorImpl implements OutboundFileProcessor {
             String audioFilenames = audioFileInfos.stream().map(AudioFileInfo::getMediaFile).collect(Collectors.joining(", "));
 
             List<ChannelAudio> concatenationsList = new ArrayList<>();
-
             
             if (isWellFormedAudio(audioFileInfos)) {
                 log.debug("Audio files {} are well formed", audioFilenames);
--- a/src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
@@ -100,10 +100,12 @@ public class OutboundFileProcessorImpl implements OutboundFileProcessor {
 
         List<AudioFileInfo> concatenatedAndMergedAudioFileInfos = new ArrayList<>();
         if (isNotEmpty(audioFileInfos)) {
+            // Used for logging only
             String audioFilenames = audioFileInfos.stream().map(AudioFileInfo::getMediaFile).collect(Collectors.joining(", "));
 
             List<ChannelAudio> concatenationsList = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
```diff
@@ -100,10 +100,11 @@ public List<AudioFileInfo> processAudioForPlaybacks(Map<MediaEntity, Path> media
 
         List<AudioFileInfo> concatenatedAndMergedAudioFileInfos = new ArrayList<>();
         if (isNotEmpty(audioFileInfos)) {
+            // Used for logging only
             String audioFilenames = audioFileInfos.stream().map(AudioFileInfo::getMediaFile).collect(Collectors.joining(", "));
```
Ben-Edwards-cgi, line 104: If this is used for logging only it will be good to wrap it in a if(log.isInfoEnabled) / if(log.isDebugEnabled) etc
 to save on performance when that logging level disabled

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
```diff
@@ -223,7 +225,15 @@ private void processAudioRequest(MediaRequestEntity mediaRequestEntity) {
     List<MediaEntity> filterMediaByMediaRequestTimeframeAndSortByStartTimeAndChannel(List<MediaEntity> mediaEntitiesForRequest,
                                                                                      MediaRequestEntity mediaRequestEntity) {
         return mediaEntitiesForRequest.stream()
+            .filter(media -> nonNull(media.getStart()))
+            .filter(media -> nonNull(media.getEnd()))
+            // Filter out media where the media start and media end times are the same
+            .filter(media -> !media.getStart().truncatedTo(ChronoUnit.SECONDS).isEqual(media.getEnd().truncatedTo(ChronoUnit.SECONDS)))
```
Ben-Edwards-cgi, line 231: Assuming we want to exclude any audios that are less then 1 second long?

I.e a 0.9 second audio should be excluded?

karen-hedges: The ffmpeg command is truncated to seconds so the filtering also needs to do the same

karen-hedges: The media request can only be in seconds so any media that is less than 1 second will also need to be filtered out. The ffmpeg trim operation is performed at second level so this will fail on any files that are less than 1 second long.

src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
karen-hedges: done

src/main/java/uk/gov/hmcts/darts/audio/component/impl/OutboundFileProcessorImpl.java
```diff
@@ -100,7 +100,11 @@ public List<AudioFileInfo> processAudioForPlaybacks(Map<MediaEntity, Path> media
 
         List<AudioFileInfo> concatenatedAndMergedAudioFileInfos = new ArrayList<>();
         if (isNotEmpty(audioFileInfos)) {
-            String audioFilenames = audioFileInfos.stream().map(AudioFileInfo::getMediaFile).collect(Collectors.joining(", "));
+            // Used for logging only
+            String audioFilenames = null;
+            if (log.isDebugEnabled()) {
+                audioFileInfos.stream().map(AudioFileInfo::getMediaFile).collect(Collectors.joining(", "));
```
jackmaloney, line 106: Does this line do anything? It's not actually logging anything so seems redundant to me?

karen-hedges: Ben suggested I wrap the audioFilenames in a log.isDebugEnabled() but because its used in the if and else statement, I had to create a variable to store it in first and forgot to add it back in.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
```diff
@@ -223,7 +225,15 @@ private void processAudioRequest(MediaRequestEntity mediaRequestEntity) {
     List<MediaEntity> filterMediaByMediaRequestTimeframeAndSortByStartTimeAndChannel(List<MediaEntity> mediaEntitiesForRequest,
                                                                                      MediaRequestEntity mediaRequestEntity) {
         return mediaEntitiesForRequest.stream()
+            .filter(media -> nonNull(media.getStart()))
+            .filter(media -> nonNull(media.getEnd()))
+            // Filter out media where the media start and media end times are the same and not less than a second apart
```
jackmaloney, line 230: Do you need a comment before every line? Isn't obvious from the code what each line is doing?

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
```diff
@@ -223,7 +225,15 @@ private void processAudioRequest(MediaRequestEntity mediaRequestEntity) {
...
+            .filter(media -> !media.getStart().truncatedTo(ChronoUnit.SECONDS).isEqual(media.getEnd().truncatedTo(ChronoUnit.SECONDS)))
+            // Filter out media where the media start time is after the media end time
+            .filter(media -> media.getStart().isBefore(media.getEnd()))
```
jackmaloney, line 233: Is this filter and the one below valid use cases? What AC on the ticket do these relate to?

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
```diff
@@ -223,7 +225,15 @@ private void processAudioRequest(MediaRequestEntity mediaRequestEntity) {
...
+            // Filter out media where the media end time is after the media request start time
             .filter(media -> mediaRequestEntity.getStartTime().isBefore(media.getEnd()))
```
jackmaloney, line 235: Is this filter and the one below valid use cases? What AC on the ticket do these relate to?

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
karen-hedges: I added them to reinforce the intention in case of changes in the future.

src/main/java/uk/gov/hmcts/darts/audio/service/impl/AudioTransformationServiceImpl.java
karen-hedges: line 231 is for AC1 where the start and end times are the same but also added where less than 1 second as its not possible to trim a file that is less than one second long. 
line 233 is for AC2 where the start time is after the end time
line 228 is for AC3
line 229 is for AC4

jackmaloney (approved):

Ben-Edwards-cgi (approved):

hemantasharma1129 (approved):

PR #2413, opened 2025-01-03
DMP-4538 ARM RPO - Endpoint changes

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-4538

### Change description ###

- Added code to sleep between addAsyncSearch & SaveBackgroundSearch RPO calls.
- Added check if isSaved is true in getExtendedSearchesByMatter.
- Added checks if the getExtendedSearchesByMatter and in progress is current state for reprocessing

```
[ ] Yes
[ x] No
```

### DIFF
--- a/charts/darts-api/values.dev.template.yaml
+++ b/charts/darts-api/values.dev.template.yaml
@@ -114,7 +114,6 @@ java:
     AUTOMATED_TASK_MODE: true
     API_MODE: true
     IS_MOCK_ARM_RPO_DOWNLOAD_CSV: false
-    ARM_RPO_THREAD_SLEEP_DURATION: 5s
 
   secrets:
     DARTS_API_DB_PASSWORD:
--- a/charts/darts-api/values.dev.template.yaml
+++ b/charts/darts-api/values.dev.template.yaml
@@ -114,6 +114,7 @@ java:
     AUTOMATED_TASK_MODE: true
     API_MODE: true
     IS_MOCK_ARM_RPO_DOWNLOAD_CSV: false
+    ARM_RPO_THREAD_SLEEP_DURATION: 5s
 
   secrets:
     DARTS_API_DB_PASSWORD:
--- a/src/integrationTest/java/uk/gov/hmcts/darts/arm/service/ArmRpoPollServiceIntTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/arm/service/ArmRpoPollServiceIntTest.java
@@ -139,6 +139,41 @@ class ArmRpoPollServiceIntTest extends PostgresIntegrationBase {
     }
 
     @Test
+    void pollArmRpo_shouldPollSuccessfullyWithSaveBackgroundCompletedGetExtendedProductionsByMatterInProgress() throws IOException {
+        // given
+        armRpoExecutionDetailEntity.setArmRpoStatus(ArmRpoHelper.completedRpoStatus());
+        armRpoExecutionDetailEntity.setArmRpoState(ArmRpoHelper.getExtendedProductionsByMatterRpoState());
+        armRpoExecutionDetailEntity.setMatterId("MatterId");
+        armRpoExecutionDetailEntity.setSearchId("SearchId");
+        armRpoExecutionDetailEntity.setStorageAccountId("StorageAccountId");
+        armRpoExecutionDetailEntity.setProductionId(PRODUCTION_ID);
+        armRpoExecutionDetailEntity = dartsPersistence.save(armRpoExecutionDetailEntity);
+
+        when(armApiService.getArmBearerToken()).thenReturn(BEARER_TOKEN);
+        when(armRpoClient.getExtendedSearchesByMatter(any(), any()))
+            .thenReturn(getExtendedSearchesByMatterResponse());
+        when(armRpoClient.getMasterIndexFieldByRecordClassSchema(any(), any()))
+            .thenReturn(getMasterIndexFieldByRecordClassSchemaResponse("propertyName", "ingestionDate"));
+        when(armRpoClient.createExportBasedOnSearchResultsTable(anyString(), any()))
+            .thenReturn(getCreateExportBasedOnSearchResultsTableResponseInProgress());
+
+        // when
+        armRpoPollService.pollArmRpo(false);
+
+        // then
+        var updatedArmRpoExecutionDetailEntity = dartsPersistence.getArmRpoExecutionDetailRepository().findById(armRpoExecutionDetailEntity.getId());
+        assertNotNull(updatedArmRpoExecutionDetailEntity);
+        assertEquals(ArmRpoHelper.createExportBasedOnSearchResultsTableRpoState().getId(), updatedArmRpoExecutionDetailEntity.get().getArmRpoState().getId());
+        assertEquals(ArmRpoHelper.inProgressRpoStatus().getId(), updatedArmRpoExecutionDetailEntity.get().getArmRpoStatus().getId());
+
+        verify(armRpoClient).getExtendedSearchesByMatter(any(), any());
+        verify(armRpoClient).getMasterIndexFieldByRecordClassSchema(any(), any());
+        verify(armRpoClient).createExportBasedOnSearchResultsTable(anyString(), any());
+        verifyNoMoreInteractions(armRpoClient);
+
+    }
+
+    @Test
     void pollArmRpo_shouldPollSuccessfullyWithSaveBackgroundCompletedCreateExportInProgress() throws IOException {
         // given
         armRpoExecutionDetailEntity.setArmRpoStatus(ArmRpoHelper.completedRpoStatus());
@@ -341,6 +376,7 @@ class ArmRpoPollServiceIntTest extends PostgresIntegrationBase {
         ExtendedSearchesByMatterResponse.Search search = new ExtendedSearchesByMatterResponse.Search();
         search.setTotalCount(4);
         search.setName("DARTS_RPO_2024-08-13");
+        search.setIsSaved(true);
         ExtendedSearchesByMatterResponse.SearchDetail searchDetail = new ExtendedSearchesByMatterResponse.SearchDetail();
         searchDetail.setSearch(search);
         response.setSearches(List.of(searchDetail));
--- a/src/integrationTest/java/uk/gov/hmcts/darts/arm/service/ArmRpoPollServiceIntTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/arm/service/ArmRpoPollServiceIntTest.java
@@ -139,10 +139,10 @@ class ArmRpoPollServiceIntTest extends PostgresIntegrationBase {
     }
 
     @Test
-    void pollArmRpo_shouldPollSuccessfullyWithSaveBackgroundCompletedGetExtendedProductionsByMatterInProgress() throws IOException {
+    void pollArmRpo_shouldPollSuccessfullyWithGetExtendedSearchesByMatterInProgress() throws IOException {
         // given
-        armRpoExecutionDetailEntity.setArmRpoStatus(ArmRpoHelper.completedRpoStatus());
-        armRpoExecutionDetailEntity.setArmRpoState(ArmRpoHelper.getExtendedProductionsByMatterRpoState());
+        armRpoExecutionDetailEntity.setArmRpoStatus(ArmRpoHelper.inProgressRpoStatus());
+        armRpoExecutionDetailEntity.setArmRpoState(ArmRpoHelper.getExtendedSearchesByMatterRpoState());
         armRpoExecutionDetailEntity.setMatterId("MatterId");
         armRpoExecutionDetailEntity.setSearchId("SearchId");
         armRpoExecutionDetailEntity.setStorageAccountId("StorageAccountId");
@@ -155,7 +155,17 @@ class ArmRpoPollServiceIntTest extends PostgresIntegrationBase {
         when(armRpoClient.getMasterIndexFieldByRecordClassSchema(any(), any()))
             .thenReturn(getMasterIndexFieldByRecordClassSchemaResponse("propertyName", "ingestionDate"));
         when(armRpoClient.createExportBasedOnSearchResultsTable(anyString(), any()))
-            .thenReturn(getCreateExportBasedOnSearchResultsTableResponseInProgress());
+            .thenReturn(getCreateExportBasedOnSearchResultsTableResponse());
+        when(armRpoClient.getExtendedProductionsByMatter(anyString(), any()))
+            .thenReturn(getExtendedProductionsByMatterResponse());
+        when(armRpoClient.getProductionOutputFiles(any(), any()))
+            .thenReturn(getProductionOutputFilesResponse(PRODUCTION_ID));
+
+        when(armRpoDownloadProduction.downloadProduction(any(), any(), any()))
+            .thenReturn(getFeignResponse(200));
+
+        when(armRpoClient.removeProduction(any(), any()))
+            .thenReturn(getRemoveProductionResponse());
 
         // when
         armRpoPollService.pollArmRpo(false);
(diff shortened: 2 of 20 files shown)


hemantasharma1129 (approved):

charts/darts-api/values.dev.template.yaml
```diff
@@ -114,6 +114,7 @@ java:
     AUTOMATED_TASK_MODE: true
     API_MODE: true
     IS_MOCK_ARM_RPO_DOWNLOAD_CSV: false
+    ARM_RPO_THREAD_SLEEP_DURATION: 5s
```
Ben-Edwards-cgi, line 117: Is this needed if we default to 5s if it is not provided?

src/main/java/uk/gov/hmcts/darts/arm/exception/ArmRpoGetExtendedSearchesByMatterIdException.java
```diff
@@ -0,0 +1,8 @@
+package uk.gov.hmcts.darts.arm.exception;
+
+@SuppressWarnings("java:S110")
```
Ben-Edwards-cgi, line 3: Should action instead of supressing

src/main/java/uk/gov/hmcts/darts/arm/service/impl/TriggerArmRpoSearchServiceImpl.java
```diff
@@ -84,4 +89,14 @@ public void triggerArmRpoSearch() {
...
 
+    // Added method to fix sonar complaint
+    void sleep(Duration threadSleepDuration) {
+        try {
+            Thread.sleep(threadSleepDuration);
+        } catch (InterruptedException e) {
+            log.error("Trigger ARM RPO search thread sleep interrupted", e);
+            Thread.currentThread().interrupt();
+        }
+    }
```
Ben-Edwards-cgi, line 100: Would be good to stick this into a Util class.
This way if we ever change how the sleep logic works it is all centralised across the application

src/test/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiGetExtendedSearchesByMatterTest.java
```diff
@@ -87,6 +89,39 @@ void getExtendedSearchesByMatter_Success() {
...
+        when(armRpoClient.getExtendedSearchesByMatter(anyString(), any())).thenReturn(extendedSearchesByMatterResponse);
+
+        // when
+        ArmRpoGetExtendedSearchesByMatterIdException armRpoGetExtendedSearchesByMatterIdException = assertThrows(
+            ArmRpoGetExtendedSearchesByMatterIdException.class, () -> armRpoApi.getExtendedSearchesByMatter("token", 1, userAccount));
+
+        // then
+        assertThat(armRpoGetExtendedSearchesByMatterIdException.getMessage(),
+                   containsString("The extendedSearchesByMatterResponse is not saved"));
+        verify(armRpoService).updateArmRpoStateAndStatus(any(),
```
Ben-Edwards-cgi, line 117: any() should be avoided in verify statements to ensure the expected values are being passed

src/test/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiGetExtendedSearchesByMatterTest.java
```diff
@@ -87,6 +89,39 @@ void getExtendedSearchesByMatter_Success() {
...
+                                                         eq(ARM_RPO_HELPER_MOCKS.getGetExtendedSearchesByMatterRpoState()),
+                                                         eq(ARM_RPO_HELPER_MOCKS.getInProgressRpoStatus()),
+                                                         any());
```
Ben-Edwards-cgi, line 120: any() should be avoided in verify statements to ensure the expected values are being passed

src/test/java/uk/gov/hmcts/darts/arm/service/impl/ArmRpoPollServiceImplTest.java
```diff
@@ -169,6 +170,47 @@ void pollArmRpo_shouldPollSuccessfully_whenSaveBackgroundCompletedForManualRun()
...
+            .thenReturn(createHeaderColumns());
+        when(armRpoApi.createExportBasedOnSearchResultsTable(anyString(), anyInt(), any(), any(), any())).thenReturn(true);
+        when(armRpoApi.getProductionOutputFiles(anyString(), anyInt(), any())).thenReturn(List.of("fileId"));
+        InputStream resource = IOUtils.toInputStream("dummy input stream", "UTF-8");
+        when(armRpoApi.downloadProduction(anyString(), anyInt(), anyString(), any())).thenReturn(resource);
+        doNothing().when(armRpoApi).removeProduction(anyString(), anyInt(), any());
+
+        String fileName = "fileId.csv";
+        Path filePath = Path.of(tempDirectory.getAbsolutePath()).resolve(fileName);
+        when(fileOperationService.saveFileToTempWorkspace(any(InputStream.class), anyString(), any(), anyBoolean())).thenReturn(filePath);
```
Ben-Edwards-cgi, line 191: Optional (and maybe worth a separate ticket) but it would be good to get away from the temp workspace and use in-memory processing where applicable. (Unless the file size is very large)

src/test/java/uk/gov/hmcts/darts/arm/service/impl/ArmRpoPollServiceImplTest.java
```diff
@@ -169,6 +170,47 @@ void pollArmRpo_shouldPollSuccessfully_whenSaveBackgroundCompletedForManualRun()
...
+        verify(armRpoApi).downloadProduction(anyString(), anyInt(), any(), any());
+        verify(armRpoApi).removeProduction(anyString(), anyInt(), any());
+
+        verify(userIdentity).getUserAccount();
+
+        verify(fileOperationService).saveFileToTempWorkspace(any(InputStream.class), anyString(), any(), anyBoolean());
+
+        verify(logApi).armRpoPollingSuccessful(any());
+
+        verifyNoMoreInteractions(armRpoApi, userIdentity, fileOperationService, logApi);
```
Ben-Edwards-cgi, line 211: We should avoid using any() in verify statements to ensure we pass the correct values to the methods.

src/test/java/uk/gov/hmcts/darts/arm/service/impl/ArmRpoPollServiceImplTest.java
```diff
@@ -376,7 +419,26 @@ void pollArmRpo_shouldHandleCreateExportInProgress() {
...
+        when(armApiService.getArmBearerToken()).thenReturn("bearerToken");
+        when(armRpoApi.getExtendedSearchesByMatter(anyString(), anyInt(), any())).thenThrow(ArmRpoGetExtendedSearchesByMatterIdException.class);
+
+        // when
+        armRpoPollService.pollArmRpo(false);
+
+        // then
+        verify(armRpoService).getLatestArmRpoExecutionDetailEntity();
+        verify(armApiService).getArmBearerToken();
+        verify(armRpoApi).getExtendedSearchesByMatter(anyString(), anyInt(), any());
```
Ben-Edwards-cgi, line 440: We should avoid using any() in verify statements to ensure we pass the correct values to the methods.

src/test/java/uk/gov/hmcts/darts/arm/service/impl/TriggerArmRpoSearchServiceImplTest.java
```diff
@@ -101,23 +101,36 @@ void triggerArmRpoSearch_shouldCallExpectedApis() {
...
+        // Given
+        Duration threadSleepDuration = Duration.ofMillis(1);
+
+        // When
+        Thread.currentThread().interrupt(); // Simulate an interrupt
+        triggerArmRpoSearchServiceImpl.sleep(threadSleepDuration);
+
+        // Then
+        assertTrue(Thread.interrupted(), "Trigger ARM RPO search thread sleep interrupted");
+    }
```
Ben-Edwards-cgi, line 135: This does not simulate an thread interruption correctly as you interrupt the thread before you call the sleep method as such the sleep interrupt logic never gets hit.

To test this you will need to run the sleep method on a new thread (outside the test thread) and interupt after you call the sleep method to ensure you get the correct exception thrown.

src/integrationTest/java/uk/gov/hmcts/darts/arm/service/ArmRpoPollServiceIntTest.java
```diff
@@ -138,6 +138,56 @@ void pollArmRpo_shouldPollSuccessfullyWithSaveBackgroundCompleted() throws IOExc
...
+        assertEquals(ArmRpoHelper.removeProductionRpoState().getId(), updatedArmRpoExecutionDetailEntity.get().getArmRpoState().getId());
+        assertEquals(ArmRpoHelper.completedRpoStatus().getId(), updatedArmRpoExecutionDetailEntity.get().getArmRpoStatus().getId());
+
+        verify(armRpoClient).getExtendedSearchesByMatter(any(), any());
+        verify(armRpoClient).getMasterIndexFieldByRecordClassSchema(any(), any());
+        verify(armRpoClient).createExportBasedOnSearchResultsTable(anyString(), any());
+        verify(armRpoClient).getExtendedProductionsByMatter(anyString(), any());
+        verify(armRpoClient).getProductionOutputFiles(any(), any());
+        verify(armRpoDownloadProduction).downloadProduction(any(), any(), any());
+        verify(armRpoClient).removeProduction(any(), any());
```
Ben-Edwards-cgi, line 185: any() should be avoided in verify statements to ensure the mocked method gets called with the correct values

karen-hedges: We are verifying the methods are being called not the variable values

src/test/java/uk/gov/hmcts/darts/arm/service/impl/ArmRpoPollServiceImplTest.java
karen-hedges: We are verifying the methods are being called not the variable values

src/test/java/uk/gov/hmcts/darts/arm/service/impl/ArmRpoPollServiceImplTest.java
karen-hedges: This can be a million rows of an excel file that has to be read, so no in memory is not an option.

src/test/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiGetExtendedSearchesByMatterTest.java
karen-hedges: We are verifying the methods are being called not the variable values

src/test/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiGetExtendedSearchesByMatterTest.java
karen-hedges: We are verifying the methods are being called not the variable values

Ben-Edwards-cgi: To ensure the method is being called correctly we also need to verify the values passed to the method.
As we could be passing invalid values or values for the wrong object which means the method behave unexpectedly

karen-hedges: Where we have an expectation of it taking a specific value and not another then it makes sense to use exact values but for this we don't care what the value is, just that it gets called.

src/main/java/uk/gov/hmcts/darts/arm/exception/ArmRpoGetExtendedSearchesByMatterIdException.java
karen-hedges: This is a restriction on the number of classes that can be inherited from. This is why it is being suppressed.

src/main/java/uk/gov/hmcts/darts/arm/client/model/rpo/ExtendedSearchesByMatterResponse.java
```diff
@@ -31,6 +31,7 @@ public static class Search {
 
         private String name;
         private Integer totalCount;
+        private Boolean isSaved;
```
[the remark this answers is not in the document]
karen-hedges: Json should use objects not primitives because values can be null.

charts/darts-api/values.dev.template.yaml
karen-hedges: ARM have only just come up with this random value and we suspect this will change again in the future

src/test/java/uk/gov/hmcts/darts/arm/service/impl/TriggerArmRpoSearchServiceImplTest.java
karen-hedges: As you can see by the assertion the error message that is thrown is my specific one. Although I do understand it looks odd it seems to work.

charts/darts-api/values.dev.template.yaml
Ben-Edwards-cgi: I think we can still configure it in the yaml using the defaults as we don't need it to be different on dev environments

src/main/java/uk/gov/hmcts/darts/arm/client/model/rpo/ExtendedSearchesByMatterResponse.java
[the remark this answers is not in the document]
Ben-Edwards-cgi: Should we treat null as being false in this scenario?
If not then I am happy with this :)

But if null is essentially false using primitives makes life easier

src/test/java/uk/gov/hmcts/darts/arm/service/impl/ArmRpoPollServiceImplTest.java
Ben-Edwards-cgi: Do we know what the max we would get in darts is?

src/test/java/uk/gov/hmcts/darts/arm/service/impl/TriggerArmRpoSearchServiceImplTest.java
Ben-Edwards-cgi: This assertion just verifies that the Thread.interrupted() is true.

The message is not the exception message but rather a message that gets put into console if the assertion fails to help with debugging. (So this is just a test debug message)

src/test/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiGetExtendedSearchesByMatterTest.java
Ben-Edwards-cgi: As the methods are stubbed the values we pass are especially important to ensure we are giving them the correct values.

As if it is called but with the wrong data then the test should fail.

src/main/java/uk/gov/hmcts/darts/arm/service/impl/TriggerArmRpoSearchServiceImpl.java
karen-hedges: Sleep shouldn't really be used but unfortunately is what is required here. We don't want to be promoting the use of sleep really.

charts/darts-api/values.dev.template.yaml
karen-hedges: removed

src/test/java/uk/gov/hmcts/darts/arm/service/impl/ArmRpoPollServiceImplTest.java
karen-hedges: 1 million rows is the max number of rows you can get in excel so that is the max we can be sent

src/test/java/uk/gov/hmcts/darts/arm/service/impl/TriggerArmRpoSearchServiceImplTest.java
karen-hedges: Ah I am misreading my assert there

karen-hedges: I have changed it to be called within a thread

src/test/java/uk/gov/hmcts/darts/arm/rpo/impl/ArmRpoApiGetExtendedSearchesByMatterTest.java
karen-hedges: If a value could be different or changes then it makes sense to verify the value but if its the same the whole time you are adding no value to verify the actual values

src/integrationTest/java/uk/gov/hmcts/darts/arm/service/ArmRpoPollServiceIntTest.java
jackmaloney: Isn't the answer to this is 'it depends'? Not sure there is a hard and fast rule that we shouldn't be using any() in verify statements but acknowledge that it can make the test harder to read with any's all over the place and could lead to false positives in certain situations, especially if someone makes a change in future. 
I guess the question is how important do we think the above is? If you're confident this isn't an issue then go with what you have @karen-hedges.

jackmaloney (approved):

src/integrationTest/java/uk/gov/hmcts/darts/arm/service/ArmRpoPollServiceIntTest.java
karen-hedges: These are all the calls the client so putting values in will mean generating all the json for them which is covered in another test that covers the client request/responses, hence unnecessary to do here.

jackmaloney: Just to clarify, you wouldn't need to generate all the json for this @karen-hedges . It's as simple as doing something like this:

when(armRpoClient.removeProduction(anyString(), any(RemoveProductionRequest.class)))

PR #2418, opened 2025-01-06
DMP-4471 ARM IU file content parsing error

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-4471

### Change description ###

Added code to delete the ARM IU file and referenced files if the file cannot be parsed

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFiles.java
```diff
@@ -120,14 +121,16 @@ public void processResponseFiles(int batchSize) {
 
     private void processInputUploadBlob(String inputUploadBlob, UserAccountEntity userAccount) {
         log.debug("Found ARM Input Upload file {}", inputUploadBlob);
+        BatchInputUploadFileFilenameProcessor batchUploadFileFilenameProcessor = null;
```
Ben-Edwards-cgi, line 124: I don't think the = null; is required here

karen-hedges: It is required for the exception as it passed in

src/integrationTest/java/uk/gov/hmcts/darts/arm/service/impl/DeleteArmResponseFilesHelperIntTest.java
```diff
@@ -0,0 +1,148 @@
...
+    private static final String RESPONSE_FILE_PREFIX = "04e6bc3b-952a-79b6-8362-13259aae1895";
+
+    @Autowired
+    private ExternalObjectDirectoryRepository externalObjectDirectoryRepository;
+
+    @MockBean
+    private ArmDataManagementApi armDataManagementApi;
+
+    @MockBean
+    private UserIdentity userIdentity;
```
Ben-Edwards-cgi, line 46: We should avoid using mocks for integration tests where we can

Ben-Edwards-cgi: We can set the Security user context which will auto log the user in.
I believe other tests use this approach :)

private ArmDataManagementApi armDataManagementApi; is also mocked would it be possible to mock this at a client level instead of service level?

(Just integration tests should cover the full component integration so mocks should be avoided to ensure all of our components can talk to each other correctly the exception to this is external clients which make sense to mock as integration does not cover E2E flows)

hemantasharma1129 (approved):

chrisbellingham-hmcts (approved):

PR #2461, opened 2025-01-21
DMP-4616: Flyway Updates : Add new role MEDIA_IN_PERPETUITY

Ben-Edwards-cgi:
### Change description ###
1. Add a new role as MEDIA_IN_PERPETUITY with display_name as "Media in Perpetuity" and display_state as TRUE.

2. Map the newly created MEDIA_IN_PERPETUITY role with existing group MEDIA_IN_PERPETUITY (grp_id=7).

3. Hide MEDIA_ACCESSOR Role by setting display_state = false.

```
[ ] Yes
[X] No
```


src/main/resources/db/migration/common/V1_429__added_role_MEDIA_IN_PERPETUITY.sql
```diff
@@ -0,0 +1,16 @@
+insert into security_role (rol_id, role_name, display_name, display_state)
+values (nextval('rol_seq'), 'MEDIA_IN_PERPETUITY', 'Media in Perpetuity', TRUE);
+
+
+update security_group
+set rol_id = (select rol_id from security_role where role_name = 'MEDIA_IN_PERPETUITY')
+where grp_id = 7;
```
chrisbellingham-hmcts, line 7: Just to make the intent of the `UPDATE` statement clearer, is it worth using a subquery as the where clause that selects the `MEDIA_IN_PERPETUITY` group by group_name? Otherwise the reader needs to do some work to understand which group `7` is.

Ben-Edwards-cgi: Updated

chrisbellingham-hmcts (approved):

jackmaloney (approved):

PR #2464, opened 2025-01-21
DMP-4630: Flyway Updates : Disable Automated jobs

Ben-Edwards-cgi:
### Change description ###
# Summary of Git Diff

This Git Diff introduces a new SQL migration file that disables specific automated tasks in the database.

## Highlights
- New file created: `V1_430__disabled_some_automated_jobs.sql`
- SQL command updates the `automated_task` table.
- The `task_enabled` flag is set to `false` for the following `aut_id` values: 22, 25, 28, 29, 30, and 31.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/resources/db/migration/postgres/V1_430__disabled_some_automated_jobs.sql
+++ b/src/main/resources/db/migration/postgres/V1_430__disabled_some_automated_jobs.sql
@@ -1,0 +1,3 @@
+update automated_task
+set task_enabled = false
+where aut_id in (22, 25, 28, 29, 30, 31);


hemantasharma1129 (approved):

src/main/resources/db/migration/postgres/V1_430__disabled_some_automated_jobs.sql
```diff
@@ -0,0 +1,3 @@
+update automated_task
+set task_enabled = false
+where aut_id in (22, 28, 29, 30, 31);
```
jackmaloney, line 3: This is probably fine, but there are some Flyway scripts that don't hardcode the id and use 'nextval('aut_seq')' to generate. Therefore, there is a chance that if production DB is wiped (which we know migration team will do) and then the flyway scripts are regenerated, the ID's could be different. It might be safer to use the 'task_name' field as that will be consistent?

Ben-Edwards-cgi: Agreed will update

jackmaloney (approved):

chrisbellingham-hmcts (approved):

PR #2517, opened 2025-01-30
DMP-4674: Log courthouse not found error on addAudio for Dyntrace

Ben-Edwards-cgi:
### Change description ###
# Summary of Git Diff

This Git diff introduces enhancements to the `AddAudioMetaDataValidator` and related classes in a Java application. The primary change is the addition of error handling and logging for cases where a specified courthouse does not exist. The updates also include modifications to the `AudioLoggerService` to support this new logging functionality and the creation of corresponding unit tests to verify the behavior.

## Highlights

- **Error Handling**: 
  - A try-catch block has been added in the `validate` method of the `AddAudioMetaDataValidator` class to catch `DartsApiException` when retrieving the courthouse.
  - If the exception indicates that the courthouse does not exist, a log entry is created using the `AudioLoggerService`.

- **New Logging Method**: 
  - A new method, `missingCourthouse`, has been added to the `AudioLoggerService` interface and its implementation to log warnings when a courthouse is not found.

- **Unit Tests**: 
  - A new test class `AddAudioMetaDataValidatorTest` is introduced to validate the behavior of the `AddAudioMetaDataValidator`.
  - Tests are included to ensure that when a nonexistent courthouse is specified, the correct exception is thrown and the logging method is invoked.
  - Additional tests confirm that valid courthouse requests do not trigger logging.

- **Code Structure**: 
  - The `AudioLoggerServiceImpl` is modified to include a `Clock` for timestamping log entries.

These changes improve the robustness of the audio metadata validation process by ensuring that invalid input is logged appropriately, aiding both debugging and system monitoring.

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/darts/log/service/impl/AudioLoggerServiceImpl.java
```diff
@@ -1,17 +1,24 @@
...
+import java.time.OffsetDateTime;
+
 import static uk.gov.hmcts.darts.util.DateTimeHelper.getDateTimeIsoFormatted;
 
 @Service
 @Slf4j
+@AllArgsConstructor
 public class AudioLoggerServiceImpl implements AudioLoggerService {
 
+    private final Clock clock;
```
karen-hedges, line 20: I thought it was agreed we wouldn't introduce clock as it creates inconsistent coding standards?

src/test/java/uk/gov/hmcts/darts/log/service/impl/AudioLoggerServiceImplTest.java
```diff
@@ -44,7 +45,7 @@ public static void tearDown() {
 
     @BeforeEach
     void setUp() {
-        audioLoggerService = new AudioLoggerServiceImpl();
+        audioLoggerService = new AudioLoggerServiceImpl(Clock.fixed(STARTED_AT.toInstant(), STARTED_AT.toZonedDateTime().getZone()));
```
karen-hedges, line 48: See comment about use of clock

src/main/java/uk/gov/hmcts/darts/log/service/impl/AudioLoggerServiceImpl.java
Ben-Edwards-cgi: My understanding was we needed a follow up call as I took an action to confirm some things. I have updated the logic so the clock stuff is simpler.

We already have inconsistencies as some places use clock some use currentTimeHelper

karen-hedges: There are only 2 production classes using clock. The rest all use the CurrentTimeHelper.

jackmaloney: Both instances where they have been introduced have been within the last 2 weeks. Can we try and use one pattern until a decision is actually made to use one over the other?

jackmaloney (changes requested):
The ticket mentions changing the message in the EventLoggerServiceImpl which doesn't look like has been done as part of this, please can you add?

src/main/java/uk/gov/hmcts/darts/log/service/impl/AudioLoggerServiceImpl.java
Ben-Edwards-cgi: I have updated the logic to use currentTimeHelper

src/main/java/uk/gov/hmcts/darts/common/helper/CurrentTimeHelper.java
```diff
@@ -14,12 +15,14 @@
 @RequiredArgsConstructor
 public class CurrentTimeHelper {
 
+    private final Clock clock;
```
jackmaloney, line 18: What does this change actually give us and why is it needed?

Ben-Edwards-cgi: It grants us much greater flexibility in adjusting dates/times without needing to utilise mocks in integration tests. (and ensures we are following industry standards for date/time operations)

src/main/java/uk/gov/hmcts/darts/log/service/impl/AudioLoggerServiceImpl.java
chrisbellingham-hmcts: I think the outcome of the call was to document the proposed usage of `Clock` (including how to use it in prod code and in unit/int tests), and then we can get a better sense of the impact to code (and test) complexity. But I'm still not sure switching is worth it, as `currentTimeHelper` does everything we need (I think) with minimal apparent drawbacks.

If we think `Clock` would allow us to more robustly test our application, that would be a compelling argument. But is that really the case? Maybe some examples would help sell it.

Ben-Edwards-cgi: In the interest of time I have removed clock and reverted to curentTimeHelper.

But clock does give us significant benefits, it mocks on a global level so you don't need to mock individual method calls. It is industry standard, and allows you easier manipulation of time objects, as you can pass in your custom TestClock so you can easily iterate based on the needs of your test where as using mocks such as currentTimeHelper this becomes much more painful to do. As you have to either set return orders, or override onAnswer (which is not type safe). It also works better with multi-threading and final classes as you are not overriding any class logic

chrisbellingham-hmcts (approved):

PR #2536, opened 2025-02-03
DMP-4653 Fixing courtroom name unique constraint violation

davet1985:

https://tools.hmcts.net/jira/browse/DMP-4653

### Change description ###

Primarily fixing courtroom name unique constraint violations.

- plus hearing, case and judge constraint violations
- increasing the attempts made by the Retryable annotation, up to 10 attempts retying every 50ms, spending a total of 500ms before failing
- most of these are used by requests via the SOAP service and if a failure did happen it would be retried by the external system
- creating a new transaction so that the retried method has the latest data, before it was stuck with a stable view when retrying and never doing anything
- adding javadoc for all methods in `RetrieveCoreObjectService` to help explain how they are used
- deprecating some methods as they are only used in tests
- removing the Retryable annotation from some methods because it's not necessary

The impact of this change, specifically the new transactions, is that it's possible that courtrooms, cases, hearings and judges could be persisted in the database when the whole request fails. I don't see this as an issue, since it's likely the request would be retried and the likelihood of this happening and the request failing is tiny, therefore any data persisted would be minimal.

These items are central to many of the requests DARTS receives and need to persisted quickly to allow other transactions to be aware of them.

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/service/RetrieveCoreObjectService.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/service/RetrieveCoreObjectService.java
@@ -43,7 +43,7 @@ public interface RetrieveCoreObjectService {
      * Retrieve or create a case and link to media.
      * @deprecated This method is only used by tests. Tests should be refactored and this method should be removed.
      */
-    @Deprecated
+    @Deprecated(since = "04/02/2025")
     HearingEntity retrieveOrCreateHearingWithMedia(String courthouseName, String courtroomName, String caseNumber, LocalDateTime hearingDate,
                                                    UserAccountEntity userAccount, MediaEntity mediaEntity);
 
@@ -52,7 +52,7 @@ public interface RetrieveCoreObjectService {
      * @deprecated This method is only used by tests.
      *     Tests should be refactored to use the other `retrieveOrCreateCourtroom` method, and this method should be removed.
      */
-    @Deprecated
+    @Deprecated(since = "04/02/2025")
     CourtroomEntity retrieveOrCreateCourtroom(CourthouseEntity courthouse, String courtroomName, UserAccountEntity userAccount);
 
     /**
@@ -87,7 +87,7 @@ public interface RetrieveCoreObjectService {
      * @deprecated This method is only used by tests.
      *     Tests should be refactored to use the other `retrieveOrCreateCase` method, and this method should be removed.
      */
-    @Deprecated
+    @Deprecated(since = "04/02/2025")
     CourtCaseEntity retrieveOrCreateCase(String courthouseName, String caseNumber, UserAccountEntity userAccount);
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/darts/common/service/RetrieveCoreObjectService.java
```diff
@@ -16,39 +18,109 @@
...
 public interface RetrieveCoreObjectService {
 
-    @Retryable(backoff = @Backoff(delay = 50), retryFor = {DataIntegrityViolationException.class, PSQLException.class})
+    /**
+     * This method is used when events are received and processed.
+     * By creating a new transaction, along with using the @Retryable annotation, we can ensure that concurrent requests to create the
+     * same hearing can be processed without causing a constraint violation.
+     */
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    @Retryable(backoff = @Backoff(delay = 50), retryFor = {DataIntegrityViolationException.class, PSQLException.class}, maxAttempts = 10)
```
Ben-Edwards-cgi, line 27: Is Retryable the best solution here? It feels like a quick patch rather then fix.

src/main/java/uk/gov/hmcts/darts/common/service/RetrieveCoreObjectService.java
```diff
@@ -16,39 +18,109 @@
...
     HearingEntity retrieveOrCreateHearingWithMedia(String courthouseName, String courtroomName, String caseNumber, LocalDateTime hearingDate,
                                                    UserAccountEntity userAccount, MediaEntity mediaEntity);
 
-    @Retryable(backoff = @Backoff(delay = 50), retryFor = {DataIntegrityViolationException.class, PSQLException.class})
+    /**
+     * Retrieve or create a courtroom.
+     * @deprecated This method is only used by tests.
+     *     Tests should be refactored to use the other `retrieveOrCreateCourtroom` method, and this method should be removed.
+     */
+    @SuppressWarnings("java:S1133")
```
Ben-Edwards-cgi, line 56: Should be actioned instead of suppressed. (Or have a reason to why it was supressed)

src/main/java/uk/gov/hmcts/darts/common/service/RetrieveCoreObjectService.java
davet1985: Retryable was already in place, I've just made adjustments and added the use of a new transaction. I did this because it wasn't working properly, it was retrying but the current transaction could not see the changes committed by a separate thread/transaction.

I'm not sure it is the best approach but it's what we are currently doing and a different one would be considerably more involved.

karen-hedges (approved):

src/main/java/uk/gov/hmcts/darts/common/service/RetrieveCoreObjectService.java
Ben-Edwards-cgi: Please can you add just a comment next to the suppression just to indicate such :) 
As S1133 means nothing to me :D

chrisbellingham-hmcts (approved):

src/main/java/uk/gov/hmcts/darts/common/service/RetrieveCoreObjectService.java
davet1985: Done.

Ben-Edwards-cgi (approved):

hemantasharma1129 (approved):

PR #2565, opened 2025-02-10
DMP-4735 Correct cron expression for tasks that run multiple times per second

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-4735

### Change description ###

Updated cron expressions that result in jobs attempting to run multiple times every second

```
[ ] Yes
[x ] No
```


hemantasharma1129 (approved):

src/main/resources/db/migration/common/V1_448__update_automated_tasks_cron_expressions.sql
```diff
@@ -0,0 +1,2 @@
+update darts.automated_task set cron_expression = '0 0/5 * ? * *' where task_name = 'InboundToUnstructuredDataStore';
+update darts.automated_task set cron_expression = '0 44 * ? * *' where task_name = 'ProcessArmResponseFiles';
```
chrisbellingham-hmcts, line 2: `darts.automated_task` - can you remove the `darts.` schema prefix from both statements please?

Having the prefix in the flyway script makes it hard to run this migration against other schemas, which has been a requirement for some of the DETS databases in the past.

Without the prefix, the migration will target the schema configured in the `flyway{}` block of `build.gradle`, which gives us more flexibility.

chrisbellingham-hmcts (approved):

src/main/resources/db/migration/common/V1_448__update_automated_tasks_cron_expressions.sql
karen-hedges: done

PR #2558, opened 2025-02-06
Data model 72.5

barry-jones50410:
**Before creating a pull request make sure that:**

- [ ] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [ ] tests have been updated / new tests has been added (if needed)

Please remove this line and everything above and fill the following sections:

### Change description ###

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[ ] No
```

PR #2566, opened 2025-02-10
DMP-4728: Apply version v72.5 datamodel updates

chrisbellingham-hmcts:
[DMP-4728](https://tools.hmcts.net/jira/browse/DMP-4728)

Apply datamodel updates per v72.5 reference script: https://github.com/hmcts/darts-api/pull/2558/

Migration tested locally vs cloned staging database.

```
[ ] Yes
[x] No
```


src/main/resources/db/migration/common/V1_449__v8_object_state_record.sql
```diff
@@ -0,0 +1,41 @@
+ALTER TABLE object_state_record
+    ADD COLUMN storage_id CHARACTER VARYING;
+
+ALTER TABLE object_state_record
+    ADD COLUMN data_ticket INTEGER;
+
```
jackmaloney, line 6: Do you not need to add the 'dets_location' column here as well?

chrisbellingham-hmcts: `dets_location` already exists, the change in [Barry's diff](https://github.com/hmcts/darts-api/pull/2558/files#diff-fc356bbdbbd5df2e53c7988a92f1274ea2b955ee9377a4a249c022cfce4a0507R22) is only to remove the dangling comment

jackmaloney (approved):

PR #2575, opened 2025-02-11
DMP-4743 Improve automated task - GenerateCaseDocument

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-4743

### Change description ###

Changed the job to get the batched ids instead of the batched objects

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/casedocument/service/impl/GenerateCaseDocumentBatchProcessorImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/casedocument/service/impl/GenerateCaseDocumentBatchProcessorImpl.java
@@ -27,13 +27,14 @@ public class GenerateCaseDocumentBatchProcessorImpl implements GenerateCaseDocum
     public void processGenerateCaseDocument(int batchSize) {
 
         OffsetDateTime caseClosedBeforeTimestamp = currentTimeHelper.currentOffsetDateTime().minusDays(caseDocumentGenerationDays);
-        var cases = caseRepository.findCasesNeedingCaseDocumentGenerated(caseClosedBeforeTimestamp, Limit.of(batchSize));
-        log.debug("Found {} cases needing case document out of a batch size {}", cases.size(), batchSize);
-        for (var courtCase : cases) {
+        var casesIdsNeedingCaseDocumentGenerated = caseRepository.findCasesIdsNeedingCaseDocumentGenerated(caseClosedBeforeTimestamp, Limit.of(batchSize));
+        log.debug("Found {} casesIdsNeedingCaseDocumentGenerated needing case document out of a batch size {}", casesIdsNeedingCaseDocumentGenerated.size(),
+                  batchSize);
+        for (var courtCaseId : casesIdsNeedingCaseDocumentGenerated) {
             try {
-                singleCaseProcessor.processGenerateCaseDocument(courtCase.getId());
+                singleCaseProcessor.processGenerateCaseDocument(courtCaseId);
             } catch (Exception exc) {
-                log.error("Error generating case document for case id '{}'", courtCase.getId(), exc);
+                log.error("Error generating case document for case id '{}'", courtCaseId, exc);
             }
         }
     }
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/darts/casedocument/service/impl/GenerateCaseDocumentBatchProcessorImpl.java
```diff
@@ -27,13 +27,14 @@ public class GenerateCaseDocumentBatchProcessorImpl implements GenerateCaseDocum
...
-        var cases = caseRepository.findCasesNeedingCaseDocumentGenerated(caseClosedBeforeTimestamp, Limit.of(batchSize));
-        log.debug("Found {} cases needing case document out of a batch size {}", cases.size(), batchSize);
-        for (var courtCase : cases) {
+        var casesIdsNeedingCaseDocumentGenerated = caseRepository.findCasesIdsNeedingCaseDocumentGenerated(caseClosedBeforeTimestamp, Limit.of(batchSize));
+        log.debug("Found {} casesIdsNeedingCaseDocumentGenerated needing case document out of a batch size {}", casesIdsNeedingCaseDocumentGenerated.size(),
+                  batchSize);
+        for (var courtCaseId : casesIdsNeedingCaseDocumentGenerated) {
             try {
-                singleCaseProcessor.processGenerateCaseDocument(courtCase.getId());
+                singleCaseProcessor.processGenerateCaseDocument(courtCaseId);
```
chrisbellingham-hmcts, line 35: For my understanding, how does this approach resolve potential race conditions? Isn't it the case that the list of ids (integers) that exist in `casesIdsNeedingCaseDocumentGenerated` would be the same as the ids (integers) that existed in the entities contained in the old `cases` list?

karen-hedges: No but it does improve performance by not getting the full object and its associated objects and the suggested ticket was related to moving away from getting the full objects to ids which is why its more an improvement than to fix a race condition.

chrisbellingham-hmcts: Ok thanks, that makes more sense. The Jira ticket talks about resolving a race condition so it might be worth updating the ticket description to avoid confusion on the purpose of this change.

chrisbellingham-hmcts (approved):

src/main/java/uk/gov/hmcts/darts/casedocument/service/impl/GenerateCaseDocumentBatchProcessorImpl.java
```diff
@@ -27,13 +27,14 @@ public class GenerateCaseDocumentBatchProcessorImpl implements GenerateCaseDocum
     public void processGenerateCaseDocument(int batchSize) {
 
         OffsetDateTime caseClosedBeforeTimestamp = currentTimeHelper.currentOffsetDateTime().minusDays(caseDocumentGenerationDays);
-        var cases = caseRepository.findCasesNeedingCaseDocumentGenerated(caseClosedBeforeTimestamp, Limit.of(batchSize));
-        log.debug("Found {} cases needing case document out of a batch size {}", cases.size(), batchSize);
-        for (var courtCase : cases) {
+        var casesIdsNeedingCaseDocumentGenerated = caseRepository.findCasesIdsNeedingCaseDocumentGenerated(caseClosedBeforeTimestamp, Limit.of(batchSize));
+        log.debug("Found {} casesIdsNeedingCaseDocumentGenerated needing case document out of a batch size {}", casesIdsNeedingCaseDocumentGenerated.size(),
```
jackmaloney, line 31: Should this log have been updated or was it a find and replace mistake?

karen-hedges: Yes it was find and replace mistake. Will correct.

karen-hedges: done

jackmaloney (approved):

PR #2611, opened 2025-02-18
Dmp 4642 refactor arm rpo api4

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-4642

### Change description ###

Refactored the ARM RPO API's 13 endpoints into individual services.

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/arm/rpo/ArmRpoApi.java
```diff
@@ -1,42 +0,0 @@
...
-import uk.gov.hmcts.darts.arm.model.rpo.MasterIndexFieldByRecordClassSchema;
-import uk.gov.hmcts.darts.common.entity.ArmRpoStateEntity;
-import uk.gov.hmcts.darts.common.entity.UserAccountEntity;
-
-import java.io.IOException;
-import java.io.InputStream;
-import java.time.Duration;
-import java.util.List;
-
-public interface ArmRpoApi {
```
chrisbellingham-hmcts, line 12: Just as a different design perspective, another approach could be to keep this interface. It's implementation would then inject all 13 of the new sub-components. Each method implementation would simply delegate down to the appropriate subcomponent.

The advantage would be that all 13 new components would be encapulated within `ArmRpoApiImpl`, meaning that any consuming components that require a broad swathe of RPO functionality (like `ArmRpoPollServiceImpl`) would only need to inject `ArmRpoApi`.

I'll leave it to you whether you see value in that.

chrisbellingham-hmcts (approved):
Approved on the basis this is lift-and-shift.

src/main/java/uk/gov/hmcts/darts/arm/rpo/ArmRpoApi.java
karen-hedges: done

PR #2625, opened 2025-02-20
Dmp 3375 implement admin endpoint get case details by

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-3375

### Change description ###

Added admin endpoint to get case details

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/cases/mapper/CasesMapper.java
```diff
@@ -162,6 +165,45 @@ public SingleCase mapToSingleCase(CourtCaseEntity caseEntity) {
         return singleCase;
     }
 
+    @Transactional
```
Ben-Edwards-cgi, line 168: Can this be marked as a read only transactional?

Ben-Edwards-cgi, line 168: Does this method need to be transactional?

If so can it be made read only?

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
```diff
@@ -258,4 +259,11 @@ public List<AdminCasesSearchResponseItem> adminCaseSearch(AdminCasesSearchReques
     public CourtCaseEntity saveCase(CourtCaseEntity courtCase) {
         return caseRepository.saveAndFlush(courtCase);
     }
+
+    @Override
+    @Transactional
+    public AdminSingleCaseResponseItem adminGetCaseById(Integer caseId) {
+        CourtCaseEntity caseEntity = getCourtCaseById(caseId);
```
Ben-Edwards-cgi, line 266: Assuming getCourtCaseById will chuck a not found if the case does not exist?

src/main/java/uk/gov/hmcts/darts/cases/mapper/CasesMapper.java
```diff
@@ -162,6 +165,45 @@ public SingleCase mapToSingleCase(CourtCaseEntity caseEntity) {
...
+    public AdminSingleCaseResponseItem mapToAdminSingleCaseResponseItem(CourtCaseEntity courtCase) {
```
Ben-Edwards-cgi, line 169: This method would also be better as being static as we do not use any global variables

src/test/java/uk/gov/hmcts/darts/cases/mapper/CasesMapperTest.java
```diff
@@ -216,5 +233,45 @@ void testMapToSingleCaseWithReportingRestriction() throws Exception {
         JSONAssert.assertEquals(expectedResponse, actualResponse, JSONCompareMode.NON_EXTENSIBLE);
     }
 
+    @Test
+    void mapToAdminSingleCaseResponseItem_ReturnsMappedData() throws IOException {
```
Ben-Edwards-cgi, line 237: Needs more tests to cover each conditional.
For example, case open, case close.
Null getReportingRestrictions
Has caseRetentionOptional does not have caseRetentionOptional

src/main/java/uk/gov/hmcts/darts/cases/mapper/CasesMapper.java
karen-hedges: Yes because it uses lazily loaded objects

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
```diff
@@ -258,4 +259,11 @@ public List<AdminCasesSearchResponseItem> adminCaseSearch(AdminCasesSearchReques
...
+        return casesMapper.mapToAdminSingleCaseResponseItem(caseEntity);
```
Ben-Edwards-cgi, line 267: This method is missing unit tests

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerAdminGetCaseByIdIntTest.java
```diff
@@ -0,0 +1,247 @@
...
+@AutoConfigureMockMvc
+class CaseControllerAdminGetCaseByIdIntTest extends IntegrationBase {
+
+    private static final String ENDPOINT_URL = "/admin/cases/{id}";
+
+    private static final OffsetDateTime SOME_DATE_TIME = OffsetDateTime.parse("2023-01-01T12:00Z");
+    private static final String SOME_COURTHOUSE = "SOME-COURTHOUSE";
+    private static final String SOME_COURTROOM = "some-courtroom";
+    private static final String SOME_CASE_NUMBER = "1";
+    public static final OffsetDateTime DATE_TIME = OffsetDateTime.parse("2023-06-26T13:00:00Z");
```
Ben-Edwards-cgi, line 55: Why is this public and the rest private?

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerAdminGetCaseByIdIntTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+    @Autowired
+    private transient MockMvc mockMvc;
```
Ben-Edwards-cgi, line 58: I don't believe transient is required here

src/main/java/uk/gov/hmcts/darts/cases/mapper/CasesMapper.java
karen-hedges: It uses various methods in this class. Changing to static would then mean changing all the other methods it uses which it uses, that are used by other methods in this class.

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
karen-hedges: yes its using the same method that is used by get case by id for a normal user

Ben-Edwards-cgi (changes requested):
Test instability should be fixed

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
```diff
@@ -258,4 +259,11 @@ public List<AdminCasesSearchResponseItem> adminCaseSearch(AdminCasesSearchReques
     public CourtCaseEntity saveCase(CourtCaseEntity courtCase) {
         return caseRepository.saveAndFlush(courtCase);
     }
+
+    @Override
+    @Transactional
```
Ben-Edwards-cgi, line 264: Could this be made read only?

src/main/java/uk/gov/hmcts/darts/cases/mapper/CasesMapper.java
Ben-Edwards-cgi: As the CourtEntity is already fetched I don't believe transactional is able to lazy load from this entity.
The reason this is working to date is because the calling method is also wrapped in a transactional.

src/main/java/uk/gov/hmcts/darts/cases/mapper/CasesMapper.java
Ben-Edwards-cgi: I think if it is a small refactor I think it make sense to convert to static but if this would require significant work. Then we can tidy up as a tech debt ticket down stream

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
karen-hedges: I did most the test a integration tests not unit tests

Ben-Edwards-cgi: We should be aiming to shift left our testing, so they should ideally be in unit tests as they run much quicker then integration.

karen-hedges: Unit tests are not as accurate as integration tests

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerAdminGetCaseByIdIntTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+        // when
+        MvcResult mvcResult = mockMvc.perform(requestBuilder).andExpect(status().isOk()).andReturn();
+
+        // then
+        String actualResponse = mvcResult.getResponse().getContentAsString();
+        String expectedResponse = getContentsFromFile(
+            "tests/cases/CaseControllerAdminGetCaseByIdTest/testCaseOpen/expectedResponse.json");
+        expectedResponse = expectedResponse.replace("<CREATED_AT>", courtCase.getCreatedDateTime().toString());
+        expectedResponse = expectedResponse.replace("<LAST_MODIFIED_AT>", courtCase.getLastModifiedDateTime().toString());
```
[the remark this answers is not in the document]
karen-hedges: done

src/main/java/uk/gov/hmcts/darts/cases/mapper/CasesMapper.java
chrisbellingham-hmcts: Isn't there a testability impact of making methods like this static, as they're harder to mock in unit tests?

Generally we don't use static methods in Spring managed components as it goes against the DI approach.

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
Ben-Edwards-cgi: Unit tests are designed to test individual components of an application.
Integration tests are designed to test how the components talk to eachother.
So we need the unit tests to ensure the contracts between the components are working as we expect.

Simply moving tests to integration because it tests more things inherently adds risk as you are not focusing on individual component actions which you can reuse but rather the entire flow. Increasing the likelihood of missing things.

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerAdminGetCaseByIdIntTest.java
karen-hedges: done

src/main/java/uk/gov/hmcts/darts/cases/mapper/CasesMapper.java
karen-hedges: done

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
karen-hedges: done

src/main/java/uk/gov/hmcts/darts/cases/mapper/CasesMapper.java
Ben-Edwards-cgi: Looks good  :)

src/main/java/uk/gov/hmcts/darts/cases/mapper/CasesMapper.java
Ben-Edwards-cgi: Agreed :)

src/test/java/uk/gov/hmcts/darts/cases/mapper/CasesMapperTest.java
Ben-Edwards-cgi: New tests looks good but we could still benefit from testing 
Null getReportingRestrictions and the conditions in method: populateRetentionDetails (You may want to make this method package-private so you can test it in isolation)

src/main/java/uk/gov/hmcts/darts/cases/service/impl/CaseServiceImpl.java
Ben-Edwards-cgi: Looks good :)

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerAdminGetCaseByIdIntTest.java
Ben-Edwards-cgi: Looks good :)

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerAdminGetCaseByIdIntTest.java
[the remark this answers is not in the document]
Ben-Edwards-cgi: looks good :)

src/test/java/uk/gov/hmcts/darts/cases/mapper/CasesMapperTest.java
karen-hedges: Added tests to show difference between null restrictions and default restrictions

chrisbellingham-hmcts (approved):

src/test/java/uk/gov/hmcts/darts/cases/mapper/CasesMapperTest.java
Ben-Edwards-cgi: I don't see these tests? But I could be looking in the wrong place

jackmaloney (approved):

src/test/java/uk/gov/hmcts/darts/cases/mapper/CasesMapperTest.java
karen-hedges: There are tests:
mapToAdminSingleCaseResponseItem_WithCaseOpenNullReportingRestrictions
mapToAdminSingleCaseResponseItem_WithCaseOpenDefaultReporting
mapToAdminSingleCaseResponseItem_WithCaseClosed

Ben-Edwards-cgi (approved):

PR #2633, opened 2025-02-21
DMP-4783: Completed Transcript tab search on hearing date does not always return expected results 

Ben-Edwards-cgi:
### Change description ###
In the admin portal, some transcription documents are not being returned in the 'Completed Transcripts' search results when hearing date is used as the filter. This looks to be because the transcripts in question do not have a hearing associated with them. However, there is a document that exists for them, and there is a hearing date set in the Transcripts record. 
h4. Prerequisites:
 # Find a transcription document assigned to a transcription request that has no link to a hearing in the 'hearing_transcription_ae' table. 
 # Ensure the transcription request has a hearing date set in the 'transcription' table

h4. Steps to reproduce:
 # Go to the transcription in the Transcripts > 'Completed Transcripts' tab on admin portal 
 # Search for the transcription set up in the Prerequisites by entering the 'hearing date' set in step 2 above. 

h4. Expected results:
 # The transcription document is shown in the results

h4. Actual results:
 # It is not shown in the results

h4. Impact:
 # Admin users cannot find the required transcription document

h4. Workarounds:
 # Admin would need to perform a database query to find what they are looking for 

 

*Technical* 

For reference, look at the searchTranscriptions method in TranscriptionSearchQueryImpl. This solves this issue for the 'Transcript requests' tab, which is similar to the 'Completed transcripts' tab, by performing 2 separate queries, one specifically for modernised and one specifically for legacy (as the links for hearings are not always there for migrated data from legacy), and then merges the results. For this 'Completed transcripts' tab, the query for legacy data is currently missing and needs to use the hearing date from the transcription record and use the 'transcription_linked_case' table to find other required data e.g. courthouse. 

Query 1 should fetch modernised data with link to transcription through the hearing_transcription_ae table. If a link does not exist then do not return. 

Query 2 should fetch legacy data with link to transcription through the transcription_linked_case table. If a link does not exist, then just return the details that are available from the transcription table. 

Note, it looks like quite a few fields are returned to the frontend in the current query (TranscriptionDocumentRepository.findTranscriptionMedia) which are potentially not used. We may want to tidy this up as part of this ticket to make the current query and new query cleaner but will require a conversation with the frontend to make sure no changes are required there. 

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/darts/common/repository/TranscriptionDocumentRepository.java
```diff
@@ -50,14 +68,54 @@ public interface TranscriptionDocumentRepository extends JpaRepository<Transcrip
...
-    List<TranscriptionDocumentResult> findTranscriptionMedia(String caseNumber,
-                                                             String courtHouseDisplayName,
-                                                             LocalDate hearingDate,
-                                                             String requestedBy,
-                                                             OffsetDateTime requestedAtFrom,
-                                                             OffsetDateTime requestedAtTo,
-                                                             Boolean isManualTranscription,
-                                                             String owner);
+    @SuppressWarnings("java:S107")//We need more then 7 parameters for this select statement
+    List<TranscriptionDocumentResult> findTranscriptionMediaModenised(String caseNumber,
```
jackmaloney, line 72: typo in 'Modernised'

src/main/java/uk/gov/hmcts/darts/transcriptions/model/TranscriptionDocumentResult.java
```diff
@@ -1,53 +1,17 @@
...
     String hearingCourthouseDisplayName,
-    Integer hearingId,
     LocalDate hearingDate,
     boolean isManualTranscription,
     boolean isHidden
 ) {
-
-    @SuppressWarnings({"PMD.SimplifyBooleanReturns"})
-    @Override
-    public boolean equals(Object other) {
```
jackmaloney, line 25: If making the previous comment a set, then reinstate this.

src/main/java/uk/gov/hmcts/darts/common/repository/TranscriptionDocumentRepository.java
Ben-Edwards-cgi: Updated

src/main/java/uk/gov/hmcts/darts/common/repository/TranscriptionDocumentRepository.java
```diff
@@ -9,23 +9,41 @@
...
+    @SuppressWarnings("java:S107")//We need more then 7 parameters for this select statement
+    default List<TranscriptionDocumentResult> findTranscriptionMedia(String caseNumber,
+                                                                     String courtHouseDisplayName,
+                                                                     LocalDate hearingDate,
+                                                                     String requestedBy,
+                                                                     OffsetDateTime requestedAtFrom,
+                                                                     OffsetDateTime requestedAtTo,
+                                                                     Boolean isManualTranscription,
+                                                                     String owner) {
+        List<TranscriptionDocumentResult> data = new ArrayList<>();
```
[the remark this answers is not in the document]
Ben-Edwards-cgi: Updated

src/main/java/uk/gov/hmcts/darts/transcriptions/model/TranscriptionDocumentResult.java
Ben-Edwards-cgi: This should not be needed as if the object is the same it will have the same hash

src/integrationTest/java/uk/gov/hmcts/darts/transcriptions/controller/TranscriptionControllerAdminGetTranscriptionIntTest.java
```diff
@@ -542,6 +542,41 @@ void transcriptionDocumentSearch_onHearingDate_shouldBringBackLegacyEntries() th
...
+                                                  .header("Content-Type", "application/json")
+                                                  .content(objectMapper.writeValueAsString(request)))
+            .andExpect(status().is2xxSuccessful())
+            .andReturn();
+
+        assertEquals(200, mvcResult.getResponse().getStatus());
+
+        SearchTranscriptionDocumentResponse[] transformedMediaResponses
+            = objectMapper.readValue(mvcResult.getResponse().getContentAsByteArray(), SearchTranscriptionDocumentResponse[].class);
+        assertEquals(transcriptionDocumentResults.size(), transformedMediaResponses.length);
```
jackmaloney, line 572: Is this assert right? Don't we want to check that only 1 result is returned? And so shouldn't the test explicitly check that the expected result is 1?

Ben-Edwards-cgi: transcriptionDocumentResults only contains one item so the size is 1

jackmaloney (approved):

hemantasharma1129 (approved):

PR #2647, opened 2025-02-26
DMP-3392: Include pre-amble/post-amble in job that links cases to audio through events

Ben-Edwards-cgi:
### Change description ###
# Summary of Git Diff

This Git Diff introduces changes to the `AudioConfigurationProperties` and related classes to update the data types of certain properties from `Integer` to `Duration` for `preAmbleDuration` and `postAmbleDuration`. This change is reflected throughout the service implementation and the application configuration files to ensure consistent handling of audio timing.

## Highlights

- **Property Type Changes**:
  - Changed `preAmbleDuration` and `postAmbleDuration` from `Integer` to `Duration` in `AudioConfigurationProperties.java`.
  
- **Service Logic Updates**:
  - Updated the methods in `AudioAsyncServiceImpl.java` and `AudioLinkingAutomatedTask.java` to use `Duration` for calculating the start and end times instead of minutes offset.

- **Configuration Changes**:
  - In `application.yaml`, the duration values for `pre-amble-duration` and `post-amble-duration` are modified to include the unit (`30m` instead of `30`).

- **Test Adjustments**:
  - Updated the test cases in `AudioLinkingAutomatedTaskTest.java` to reflect the new `Duration` handling for pre- and post-amble durations.

These changes aim to enhance the clarity and functionality of the audio processing logic by leveraging `Duration` types for better time management.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/task/runner/impl/AudioLinkingAutomatedTask.java
+++ b/src/main/java/uk/gov/hmcts/darts/task/runner/impl/AudioLinkingAutomatedTask.java
@@ -69,9 +69,13 @@ public class AudioLinkingAutomatedTask
         private final EventService eventService;
         private final MediaLinkedCaseHelper mediaLinkedCaseHelper;
 
+
         @Getter
-        @Value("${darts.automated-tasks.audio-linking.audio-buffer:0s}")
-        private final Duration audioBuffer;
+        @Value("${darts.automated-tasks.pre-amble-duration:0s}")
+        private final Duration preAmbleDuration;
+        @Getter
+        @Value("${darts.automated-tasks.post-amble-duration:0s}")
+        private final Duration postAmbleDuration;
         private final UserIdentity userIdentity;
 
 
@@ -84,8 +88,8 @@ public class AudioLinkingAutomatedTask
                 EventEntity event = eventService.getEventByEveId(eveId);
                 List<MediaEntity> mediaEntities = mediaRepository.findAllByMediaTimeContains(
                     event.getCourtroom().getId(),
-                    event.getTimestamp().plus(getAudioBuffer()),
-                    event.getTimestamp().minus(getAudioBuffer()));
+                    event.getTimestamp().minus(getPreAmbleDuration()),
+                    event.getTimestamp().plus(getPostAmbleDuration()));
                 mediaEntities.forEach(mediaEntity -> {
                     mediaLinkedCaseHelper.linkMediaByEvent(event, mediaEntity, MediaLinkedCaseSourceType.AUDIO_LINKING_TASK, userAccount);
                 });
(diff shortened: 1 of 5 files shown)


jackmaloney (approved):

src/main/resources/application.yaml
```diff
@@ -128,12 +128,12 @@ darts:
     merge-workspace: ${user.home}/audiotransform/merge
     outbounddeleter:
       last-accessed-deletion-day: 2
-    pre-amble-duration: 30
+    pre-amble-duration: 30m
```
davet1985, line 131: should we allow these to be overridden by flux config?

davet1985 (approved):

src/main/java/uk/gov/hmcts/darts/task/runner/impl/AudioLinkingAutomatedTask.java
```diff
@@ -84,8 +88,8 @@ public void processEvent(Integer eveId) {
                 EventEntity event = eventService.getEventByEveId(eveId);
                 List<MediaEntity> mediaEntities = mediaRepository.findAllByMediaTimeContains(
                     event.getCourtroom().getId(),
-                    event.getTimestamp().plus(getAudioBuffer()),
-                    event.getTimestamp().minus(getAudioBuffer()));
+                    event.getTimestamp().minus(getPreAmbleDuration()),
+                    event.getTimestamp().plus(getPostAmbleDuration()));
```
chrisbellingham-hmcts, line 92: I'm not 100% sure of the expected logic but I see the args have been reversed from the original - is that intended?

So if we have an event @ 12:00, and a media in the DB that starts at 10:00 and ends at 11:40 then the media would not be returned here. But I think it should be returned since the event time is within 30 mins of the media end time?

I couldn't quite wrap my head around it, so this is the temp test I used to verify that (test fails):

```java
    @Test
    void test() {
        CourtroomEntity courtroomEntity = PersistableFactory.getCourtroomTestData()
            .someMinimal();
        dartsPersistence.save(courtroomEntity);

        MediaEntity media = PersistableFactory.getMediaTestData().someMinimalBuilder()
            .start(OffsetDateTime.parse("2025-01-01T10:00:00Z"))
            .end(OffsetDateTime.parse("2025-01-01T11:40:00Z"))
            .courtroom(courtroomEntity)
            .build()
            .getEntity();
        dartsPersistence.save(media);

        OffsetDateTime eventTime = OffsetDateTime.parse("2025-01-01T12:00:00Z");
        List<MediaEntity> medias = mediaRepository.findAllByMediaTimeContains(courtroomEntity.getId(),
                                                                                              eventTime.minusMinutes(30),
                                                                                              eventTime.plusMinutes(30));
        assertEquals(1, medias.size());
    }
```

What do you think?

Ben-Edwards-cgi: Very good spot @chrisbellingham-hmcts  :)

I have reverted this change and added additional tests to ensure this is working how we want it to

chrisbellingham-hmcts (approved):

davet1985 (approved):

PR #2669, opened 2025-03-05
DMP-4750: DEV ONLY: Flaky test? processDetsToArm_emptyList

Ben-Edwards-cgi:
### Change description ###
# Summary of Git Diff

The diff introduces a refactoring of logging assertions in several test files. The updates replace direct assertions of captured output with a new utility method called `assertOutputHasMessage`, which combines waiting for a message and asserting its presence into a single method call.

## Highlights

- **File Changes**:
  - **DetsToArmBatchPushProcessorImplTest.java**: 
    - Replaced `LogUtil.waitUntilMessage` followed by `assertThat` with `LogUtil.assertOutputHasMessage` to streamline the code.
  
  - **AudioUploadServiceImplTest.java**: 
    - Similar updates to replace multiple lines of code for logging assertions with the new utility method `assertOutputHasMessage`.

  - **AbstractLockableAutomatedTaskTest.java**: 
    - Updated multiple instances of logging assertions to use `assertOutputHasMessage` instead of the previous wait and assert pattern.

  - **LogUtil.java**: 
    - Introduced the new method `assertOutputHasMessage` that encapsulates the logic for both waiting for a message and asserting its presence, improving code readability and maintainability.

- **Code Quality Improvement**: 
  - The changes contribute to cleaner and more concise test code, reducing redundancy and enhancing the clarity of log assertion processes.

```
[ ] Yes
[X] No
```


src/test/java/uk/gov/hmcts/darts/util/LogUtil.java
```diff
@@ -24,4 +25,9 @@ public static void waitUntilMessage(CapturedOutput capturedOutput, String messag
             Thread.sleep(100);
         }
     }
+
+    public static void assertOutputHasMessage(CapturedOutput output, String message, int timeout) {
```
chrisbellingham-hmcts, line 29: To avoid ambiguity around the required time unit, consider `Duration` rather than `int` for the `timeout` type.

Ben-Edwards-cgi: I thought about this but this should never be higher then a few seconds as it will slow down tests, but I do agree this should be more clear that it is in seconds I will update the name to be more clear

davet1985 (approved):

chrisbellingham-hmcts (approved):

PR #2675, opened 2025-03-06
WIP: DMP-4639 Reduce mock usage for ARM int tests

chrisbellingham-hmcts:
WIP

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFilesIntTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFilesIntTest.java
@@ -8,7 +8,6 @@ import org.jetbrains.annotations.NotNull;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
-import org.junit.jupiter.api.io.TempDir;
 import org.mockito.Mock;
 import org.mockito.MockedStatic;
 import org.mockito.Mockito;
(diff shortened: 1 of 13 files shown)


src/integrationTest/java/uk/gov/hmcts/darts/arm/service/impl/AbstractArmBatchProcessResponseFilesIntTest.java
```diff
@@ -812,9 +773,12 @@ void batchProcessResponseFiles_With3InvalidLineFilesCausingFailure() throws IOEx
         verifyNoMoreInteractions(armDataManagementApi);
     }
 
-
     @Test
     void batchProcessResponseFiles_WithMediaUsingSmallContinuationTokenReturnsSuccess() throws IOException {
+        // We want to update the existing config value, but we need to set it back to the original afterward so that the change doesn't affect other tests.
+        // Yes, this is ugly, there must be a better way.
+        final Integer originalMaxContinuationBatchSize = armDataManagementConfiguration.getMaxContinuationBatchSize();
+        armDataManagementConfiguration.setMaxContinuationBatchSize(1);
```
Ben-Edwards-cgi, line 781: We can do this with @TestPropertySource I believe :)

chrisbellingham-hmcts: Sadly I think this only works at class level, and so the applied config would apply to all test methods. I did attempt to shift this test into a dedicated inner (`@Nested`) class using `@TestPropertySource`, but I encountered hibernate/transaction issues and the root cause wasn't clear. So admittedly I gave up and went with this slightly ugly approach.

Ben-Edwards-cgi (approved):

PR #2685, opened 2025-03-10
Dmp 4873 performance testing investigate azcopy command line parameter options

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-4873

### Change description ###

Modified the azcopy command to reduce the size of the log files generated and reduce the size of the console output.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/util/AzureCopyUtil.java
+++ b/src/main/java/uk/gov/hmcts/darts/util/AzureCopyUtil.java
@@ -21,9 +21,9 @@ public class AzureCopyUtil {
 
     private static final int SUCCESS_EXIT_CODE = 0;
     private static final String WHITESPACE = " ";
-    public static final String COPY_COMMAND = "copy";
-    public static final int INDEX_OF_SOURCE = 2;
-    public static final int INDEX_OF_DESTINATION = 3;
+    private static final String COPY_COMMAND = "copy";
+    private static final int INDEX_OF_SOURCE = 2;
+    private static final int INDEX_OF_DESTINATION = 3;
     private final DataManagementConfiguration config;
 
     public void copy(String source, String destination) {
--- a/src/main/java/uk/gov/hmcts/darts/util/AzureCopyUtil.java
+++ b/src/main/java/uk/gov/hmcts/darts/util/AzureCopyUtil.java
@@ -33,6 +33,9 @@ public class AzureCopyUtil {
             if (StringUtils.isNotEmpty(config.getAzCopyLogLevel())) {
                 command.add(config.getAzCopyLogLevel());
             }
+            if (StringUtils.isNotEmpty(config.getAzCopyCheckLength())) {
+                command.add(config.getAzCopyCheckLength());
+            }
             builder.command(command);
 
             var startTime = Instant.now();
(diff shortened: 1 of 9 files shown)


src/test/java/uk/gov/hmcts/darts/util/AzureCopyUtilTest.java
```diff
@@ -48,4 +49,84 @@ void testInterruptedExceptionMessageDoesNotIncludeSourceOrDestinationInfoToAvoid
             .hasMessageNotContaining(destinationSasUrl);
     }
 
+    @Test
+    void copy_withPreserveAccessTier() {
```
chrisbellingham-hmcts, line 53: I guess these 5 tests are following a similar pattern to other tests in this class, but they read as if we expect an exception to be thrown when `--s2s-preserve-access-tier=false`, or when `--log-level=ERROR` etc, which doesn't really make sense.

Are we really just trying to verify that the command string is being built correctly, and the exception message is being used to verify that in a roundabout way? As a more self-explanatory alternative, can we use a captor to verify the `command` string that's passed into `builder.command(command)`?

Or update the test names to better explain the intent.

src/main/java/uk/gov/hmcts/darts/util/AzureCopyUtil.java
```diff
@@ -1,26 +1,38 @@
...
 
     public void copy(String source, String destination) {
+        StringJoiner runCommand = new StringJoiner(WHITESPACE);
         try {
+            List<String> command = buildCommand(source, destination);
             ProcessBuilder builder = new ProcessBuilder();
-            builder.command(config.getAzCopyExecutable(), "copy", source, destination, config.getAzCopyPreserveAccessTier());
-
+            builder.command(command);
+            buildCensoredRunCommand(command, runCommand);
```
hemantasharma1129, line 35: This method is of no use. Please remove it.

src/main/java/uk/gov/hmcts/darts/util/AzureCopyUtil.java
```diff
@@ -29,22 +41,57 @@ public void copy(String source, String destination) {
...
-                    result);
-                log.error(errorMessage);
-                throw new DartsException(errorMessage);
+            if (exitValue != SUCCESS_EXIT_CODE) {
+                handleError(source, destination, process, builder);
             }
         } catch (InterruptedException ie) {
             Thread.currentThread().interrupt();
-            throw new DartsException("Failed to execute azure copy - interrupted", ie);
+            throw new DartsException("Failed to execute azure copy - interrupted " + runCommand, ie);
```
hemantasharma1129, line 49: Same as above, please remove "runCommand" reference.

src/test/java/uk/gov/hmcts/darts/util/AzureCopyUtilTest.java
karen-hedges: We don't want azcopy command to actually run so was using the exception to gather what the command was but had to refactor due to cognitive complexity so testing the command line builder method instead.

src/main/java/uk/gov/hmcts/darts/util/AzureCopyUtil.java
karen-hedges: removed

src/main/java/uk/gov/hmcts/darts/util/AzureCopyUtil.java
karen-hedges: removed

PR #2754, opened 2025-03-25
DMP-4796: ApplyRetentionCaseAssociatedObjectsSingleCaseProcessorImpl failing as No retentions found.

Ben-Edwards-cgi:
### Change description ###
# Summary of Git Diff

This Git Diff introduces new test cases to the `CaseRepositoryIntTest` class and modifies the `CaseRepository` class to ensure proper functionality when retrieving case IDs based on retention status and retries. The changes enhance the repository's ability to handle cases with no associated retention entities and to prevent duplicate ID returns for cases with multiple retention objects.

## Highlights

- **New Test Cases Added**:
  - `findIdsByIsRetentionUpdatedTrueAndRetentionRetriesLessThan_caseHasNoAssocaitedCaseRetentionEntity_shouldNotReturn`: Confirms that if a case has no associated retention entities, it does not return any IDs.
  - `findIdsByIsRetentionUpdatedTrueAndRetentionRetriesLessThan_oneCaseHasMultipleRetentionObjects_shouldOnlyReturnIdOnce`: Ensures that a case with multiple retention entities only returns its ID once.

- **Repository Query Changes**:
  - The SQL query in `findIdsByIsRetentionUpdatedTrueAndRetentionRetriesLessThan` was modified to select distinct IDs from `CourtCaseEntity` and ensure that cases with null retention entities are excluded.

- **Improved Logic**:
  - Enhanced logic within the query to improve efficiency and accuracy in returning case IDs based on specified conditions.

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/darts/common/repository/CaseRepository.java
```diff
@@ -52,16 +52,18 @@ AND NOT EXISTS (select 1 from CaseRetentionEntity cre
     List<Integer> findOpenCasesToClose(OffsetDateTime cutoffDate, Limit limit);
 
     @Query("""
-        SELECT cc.id
+        SELECT distinct cc.id
         FROM CourtCaseEntity cc
-        WHERE cc.isRetentionUpdated = true
+        JOIN cc.caseRetentionEntities cr                
+        WHERE cr is not null
```
jackmaloney, line 58: If you have a join on `cr` above, do you need the `WHERE cr is not null`?

Ben-Edwards-cgi: I don't think it is strictly required, but I included it as i think it makes the code more readable / understandable.

As we are not returning or checking any fields on the join someone down stream may remove it thinking it is not required.

This was to try and make it more clear

jackmaloney (approved):

PR #2758, opened 2025-03-27
DMP-4791: Update any Business validation rules to use a 422 status code (instead of 400) and prevent exception logging

Ben-Edwards-cgi:
### Change description ###
Throughout the darts application we throw a number of exceptions for business validation rules such as too many results, too broad of a search criteria, and other such validation that can only be performed once we check the database structure. These verifications validate business rule's rather then schema rules.

As such we should update these business rule validations to return a 422 status code instead of a 400.

This is so in production if we see any APIs that return a 400 status code we know we need to investigate as the FE will be sending bad data to the BE (which should not happen).

This will require both FE and BE effort.

Any endpoints that get updated should be attached to this ticket to ensure the FE updates all validation on there side correctly.

Some exceptions that fall under this category can be seen below.   
|uk.gov.hmcts.darts.common.exception.DartsApiException at uk.gov.hmcts.darts.cases.service.impl.CaseServiceImpl.adminCaseSearch(CaseServiceImpl.java:250)  (Count: 8)|
|uk.gov.hmcts.darts.common.exception.DartsApiException at uk.gov.hmcts.darts.cases.service.impl.CaseServiceImpl.advancedSearch(CaseServiceImpl.java:177)  (Count: 6)|
|uk.gov.hmcts.darts.common.exception.DartsApiException at uk.gov.hmcts.darts.audio.controller.AudioRequestsController.addAudioRequest(AudioRequestsController.java:97)  (Count: 2)|
|uk.gov.hmcts.darts.common.exception.DartsApiException at uk.gov.hmcts.darts.cases.util.RequestValidator.checkComplexity(RequestValidator.java:64)  (Count: 226)|
|uk.gov.hmcts.darts.common.exception.DartsApiException at uk.gov.hmcts.darts.retention.service.impl.RetentionPostServiceImpl.validation(RetentionPostServiceImpl.java:150)  (Count: 10)|
|uk.gov.hmcts.darts.common.exception.DartsApiException at uk.gov.hmcts.darts.retention.service.impl.RetentionPostServiceImpl.validation(RetentionPostServiceImpl.java:134)  (Count: 10)|
|uk.gov.hmcts.darts.common.exception.DartsApiException at uk.gov.hmcts.darts.hearings.service.impl.AdminHearingsServiceImpl.adminHearingSearch(AdminHearingsServiceImpl.java:41)  (Count: 4)|

The second part of this ticket is to ensure that when an exception is thrown that is related to a business rule violation we consume this and log this out as an warn instead of an error. Such that it does not show in the Azure exception logs. (As this would lead to false positives when we go into production)

*Any endpoints / errors that were updated should be logged in the comments of this ticket so that FE can check for any potential impacts before we merge this ticket*

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/darts/retention/exception/RetentionApiError.java
```diff
@@ -21,19 +21,19 @@ public enum RetentionApiError implements DartsApiError {
         RetentionTitleErrors.RETENTION_DATE_TOO_EARLY.toString()
     ), INVALID_REQUEST(
         RetentionErrorCode.INVALID_REQUEST.getValue(),
-        HttpStatus.BAD_REQUEST,
+        HttpStatus.UNPROCESSABLE_ENTITY,
         RetentionTitleErrors.INVALID_REQUEST.toString()
     ), CASE_NOT_FOUND(
         RetentionErrorCode.CASE_NOT_FOUND.getValue(),
-        HttpStatus.BAD_REQUEST,
+        HttpStatus.UNPROCESSABLE_ENTITY,
```
jackmaloney, line 28: If a case isn't found, don't we want to know about it in the logs?

src/main/java/uk/gov/hmcts/darts/retention/exception/RetentionApiError.java
```diff
@@ -57,31 +57,31 @@ public enum RetentionApiError implements DartsApiError {
...
         RetentionTitleErrors.POLICY_START_DATE_MUST_BE_PAST.toString()
     ), NON_UNIQUE_FIXED_POLICY_KEY(
         RetentionErrorCode.NON_UNIQUE_FIXED_POLICY_KEY.getValue(),
         HttpStatus.CONFLICT,
         RetentionTitleErrors.NON_UNIQUE_FIXED_POLICY_KEY.toString()
     ), FIXED_POLICY_KEY_NOT_FOUND(
         RetentionErrorCode.FIXED_POLICY_KEY_NOT_FOUND.getValue(),
-        HttpStatus.BAD_REQUEST,
+        HttpStatus.UNPROCESSABLE_ENTITY,
```
jackmaloney, line 76: If a fixed policy key isn't found, we probably want to know about it.

src/main/java/uk/gov/hmcts/darts/transcriptions/exception/TranscriptionApiError.java
```diff
@@ -55,12 +55,12 @@ public enum TranscriptionApiError implements DartsApiError {
     ),
     FAILED_TO_ATTACH_TRANSCRIPT(
         TranscriptionsErrorCode.FAILED_TO_ATTACH_TRANSCRIPT.getValue(),
-        HttpStatus.BAD_REQUEST,
+        HttpStatus.UNPROCESSABLE_ENTITY,
```
jackmaloney, line 58: Do we want to change this? Does this not mean something actually went wrong?

src/main/java/uk/gov/hmcts/darts/transcriptions/exception/TranscriptionApiError.java
```diff
@@ -55,12 +55,12 @@ public enum TranscriptionApiError implements DartsApiError {
...
         TranscriptionsTitleErrors.FAILED_TO_ATTACH_TRANSCRIPT.toString()
     ),
     FAILED_TO_DOWNLOAD_TRANSCRIPT(
         TranscriptionsErrorCode.FAILED_TO_DOWNLOAD_TRANSCRIPT.getValue(),
-        HttpStatus.BAD_REQUEST,
+        HttpStatus.UNPROCESSABLE_ENTITY,
```
jackmaloney, line 63: Do we want to change this? Does this not mean something actually went wrong?

src/main/java/uk/gov/hmcts/darts/transcriptions/exception/TranscriptionApiError.java
```diff
@@ -70,7 +70,7 @@ public enum TranscriptionApiError implements DartsApiError {
     ),
     FAILED_TO_UPDATE_TRANSCRIPTIONS(
         UpdateTranscriptions400ErrorCode.UPDATE_TRANSCRIPTIONS_PARTIAL_PROBLEM.getValue(),
-        HttpStatus.BAD_REQUEST,
+        HttpStatus.UNPROCESSABLE_ENTITY,
```
jackmaloney, line 73: Do we want to change this? Does this not mean something actually went wrong?

src/main/java/uk/gov/hmcts/darts/retention/exception/RetentionApiError.java
Ben-Edwards-cgi: Agreed, I have updated to BAD_REQUEST

src/main/java/uk/gov/hmcts/darts/transcriptions/exception/TranscriptionApiError.java
Ben-Edwards-cgi: This happens when the user gives us a transcript that has an invalid extension, content type or file size.
As such the request passes schema validation, but need to be alerted of the invalid size. 
In this case I don't think we need to log the error, as this is not an application failure condition.

src/main/java/uk/gov/hmcts/darts/transcriptions/exception/TranscriptionApiError.java
Ben-Edwards-cgi: This error is never used in application code should this be removed or should the code be updated to ensure this gets thrown?

src/main/java/uk/gov/hmcts/darts/transcriptions/exception/TranscriptionApiError.java
Ben-Edwards-cgi: Agreed but I think this should be a 500 instead of 400 as it is an app error not a schema error

src/main/java/uk/gov/hmcts/darts/retention/exception/RetentionApiError.java
davet1985: This will be fine on the FE

src/main/java/uk/gov/hmcts/darts/noderegistration/exception/NodeRegistrationApiError.java
```diff
@@ -1,31 +0,0 @@
...
-import org.springframework.http.HttpStatus;
-import uk.gov.hmcts.darts.common.exception.DartsApiError;
-import uk.gov.hmcts.darts.noderegistration.model.NoDeregistrationErrorCode;
-import uk.gov.hmcts.darts.noderegistration.model.NoDeregistrationTitleErrors;
-
-@Getter
-@RequiredArgsConstructor
-public enum NodeRegistrationApiError implements DartsApiError {
-
-    INVALID_COURTROOM(
```
davet1985, line 14: Is this not used?

src/main/java/uk/gov/hmcts/darts/transcriptions/exception/TranscriptionApiError.java
davet1985: This is also thrown here, where it probably should be a 500?

```
        try {
            BinaryData binaryData = BinaryData.fromStream(transcript.getInputStream());
            checksum = fileContentChecksum.calculate(binaryData.toBytes());
            inboundBlobCLient = dataManagementApi.saveBlobDataToContainer(binaryData, DatastoreContainerType.INBOUND, metadata);
            unstructuredBlobClient = dataManagementApi.saveBlobDataToContainer(binaryData, DatastoreContainerType.UNSTRUCTURED, metadata);
        } catch (IOException e) {
            throw new DartsApiException(FAILED_TO_ATTACH_TRANSCRIPT, e);
        }
```

Ben-Edwards-cgi: This feels like a distinctly unique error code, the previous is bad user data this is us failing to upload so having a new code here could make more sense any suggestions? FAILED_TO_UPLOAD_TRANSCRIPT?

src/main/java/uk/gov/hmcts/darts/transcriptions/exception/TranscriptionApiError.java
```diff
@@ -70,7 +70,7 @@ public enum TranscriptionApiError implements DartsApiError {
     ),
     FAILED_TO_UPDATE_TRANSCRIPTIONS(
         UpdateTranscriptions400ErrorCode.UPDATE_TRANSCRIPTIONS_PARTIAL_PROBLEM.getValue(),
-        HttpStatus.BAD_REQUEST,
+        HttpStatus.INTERNAL_SERVER_ERROR,
```
davet1985, line 73: The frontend handles 400 or 422 for this endpoint, suggest we make it 422.

src/main/java/uk/gov/hmcts/darts/transcriptions/exception/TranscriptionApiError.java
davet1985: Can we remove it if it's not used?

src/main/java/uk/gov/hmcts/darts/transcriptions/exception/TranscriptionApiError.java
davet1985: See my comment here https://github.com/hmcts/darts-api/pull/2758#discussion_r2026454037

src/main/java/uk/gov/hmcts/darts/transcriptions/exception/TranscriptionApiError.java
Ben-Edwards-cgi: Cool updated :)

src/main/java/uk/gov/hmcts/darts/transcriptions/exception/TranscriptionApiError.java
Ben-Edwards-cgi: I have removed in this PR, but we may want a bug raised if we are meant to chuck this error

src/main/java/uk/gov/hmcts/darts/transcriptions/exception/TranscriptionApiError.java
Ben-Edwards-cgi: Looking at the code a 422 seems to make sense here as it gets chucked when validation fails so will update, instead of any internal process

src/main/java/uk/gov/hmcts/darts/transcriptions/exception/TranscriptionApiError.java
Ben-Edwards-cgi: See previous comment

davet1985 (approved):

PR #2761, opened 2025-03-31
DMP-4944: Work out correct index sequences for all environments

Ben-Edwards-cgi:
### Change description ###
# Summary of Git Diff

A new SQL migration file has been added to set the current values of various sequences in the database based on the maximum ID present in their respective tables.

## Highlights

- **File Added**: `V1_469__sequences.sql`
- **Total SQL Statements**: 58
- **Functionality**: Each statement sets the current value of a sequence to the maximum ID found in a related table, ensuring that sequences are in sync with existing data.
- **Examples**:
  - Sets the value of `aat_seq` to the maximum `aat_id` from the `arm_automated_task` table.
  - Sets the value of `aud_seq` to the maximum `aud_id` from the `audit` table.
- **Overall Purpose**: To maintain the integrity and consistency of database sequences across multiple tables.

```
[ ] Yes
[X] No
```


jackmaloney (approved):

jackmaloney:
Please wait for go ahead from Jonathan before merging this.

jackmaloney:
Potentially we don't want to run this on Prod so we may need to prepend the flyway script with a word (forget what word we used when we did this in the past but we can dig it out if needed) so that it doesn't get applied automatically. And then manually apply this to each of the lower environments.

Ben-Edwards-cgi:
> Potentially we don't want to run this on Prod so we may need to prepend the flyway script with a word (forget what word we used when we did this in the past but we can dig it out if needed) so that it doesn't get applied automatically. And then manually apply this to each of the lower environments.

Is there any risk to applying this to prod if we are just updating the seq to an unused value?

PR #2762, opened 2025-03-31
DMP-4953 Your transcript screen can show the same transcript multiple times

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-4953

### Change description ###

Modified the your transcipts request to return the first requested date so if the admin puts a transcript back to requested, it only shows 1 result and the first requested date time

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/transcriptions/controller/TranscriptionControllerGetYourTranscriptsIntTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/transcriptions/controller/TranscriptionControllerGetYourTranscriptsIntTest.java
@@ -80,6 +80,7 @@ class TranscriptionControllerGetYourTranscriptsIntTest extends IntegrationBase {
 
         systemUser = authorisationStub.getSystemUser();
         testUser = authorisationStub.getTestUser();
+
     }
 
     @Test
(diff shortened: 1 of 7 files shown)


src/integrationTest/java/uk/gov/hmcts/darts/transcriptions/controller/TranscriptionControllerGetYourTranscriptsIntTest.java
```diff
@@ -366,6 +362,61 @@ void getYourTranscriptsWithApprovedStatusShouldReturnApproveOnTimeStamp() throws
 
     }
 
+    @Test
+    void getYourTranscripts_ShouldReturnSingleWorkflow_WhenWorkflowHasBeenReverted() throws Exception {
```
jackmaloney, line 366: Can you write a failing test with the old code and then make sure it passes once the new code is applied? This one seems to pass with the old and new code so can't be sure it addresses the issue.

karen-hedges: done

jackmaloney: The test only appears to cover the 'modernised' query scenario. Can we make sure the legacy scenario is also covered? The majority of migrated data will be coming down that route as most cannot be linked to hearings.

karen-hedges: Done. Had to create a new test class to handle legacy due to the setup data

Ben-Edwards-cgi (approved):

PR #2784, opened 2025-04-08
DMP-4986: Unable to search a Current version of the audio

Ben-Edwards-cgi:
### Change description ###
# Summary of Git Diff

This diff includes multiple changes across different Java files in the HMCTS DARTS application, primarily focusing on enhancements to media handling in audio-related services. It introduces new functionalities, modifies existing methods, and updates tests to ensure the correctness of the newly implemented features.

## Highlights

- **New Imports**: Added `CourtCaseEntity` and `HearingEntity` imports in `AudioControllerPatchAdminMediasByIdIntTest.java` to support new functionalities related to court cases and hearings.
  
- **Media Entity Enhancements**: 
  - Added methods in `MediaEntity` class to manage the association between media and hearings more effectively.
  - Introduced a method `isCurrent()` to check the current status of media entities.

- **Transactional Changes**: 
  - Updated the `createAndSaveMediaEntity` method to create and link a `CourtCaseEntity` with a `MediaEntity`.
  - Enhanced transaction handling when checking or updating media statuses.

- **Audio Upload Service**: 
  - New method `deleteMediaLinkingAndSetCurrentFalse` added in `AudioUploadService` to unlink hearings from media entities.

- **Admin Media Service Modifications**: 
  - Updated `patchMediasById` method to handle media linking properly while ensuring the correct media is set to current.
  - Removed the old method that set associated media to current false, reflecting a better approach using the new audio upload service method.

- **Database Stub Updates**: 
  - Marked the old method `createMediaLinkedCase` as deprecated in `DartsDatabaseStub.java`, promoting the use of the new method that uses `CourtCaseEntity`.

- **Tests Updates**: 
  - Modified unit tests in `AdminMediaServiceImplTest.java` to validate the new logic around media updates and linking.
  - Added new test cases to ensure the correct behavior of the `linkAudioToHearings` method in scenarios where hearings may or may not exist.

- **Code Cleanup**: 
  - Removed redundant comments and outdated code segments to improve overall code quality and maintainability.

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/darts/audio/service/impl/AdminMediaServiceImpl.java
```diff
@@ -336,15 +340,28 @@ public void patchMediasById(Integer id, PatchAdminMediasByIdRequest patchAdminMe
...
+
+        mediaEntityToUpdate.setIsCurrent(true);
+        mediaRepository.save(mediaEntityToUpdate);
+
+        mediaEntityToUpdate.getMediaLinkedCaseList()
+            .forEach(mediaLinkedCaseEntity -> hearingCommonService.linkAudioToHearings(
+                mediaLinkedCaseEntity.getCourtCase(),
+                mediaEntityToUpdate
+            ));
     }
```
davet1985, line 362: Is there any reason why we are attempting to construct the media to hearing links again, rather than just using the links already in place for the media that has `is_current` set to true? The new current audio should be linked to exactly the same hearings as the previous current audio.

There are other processes that can apply links between media and hearings, such as the audio linking job, and these links could be lost.

Ben-Edwards-cgi: Yes, when we check for duplicates we do not take into account case_numbers as such it is possible that the current is_current = true has a different set of hearings to the old one so the safest way to reconstruct is to use media linked case.

davet1985: Ok, that's fine.

src/main/java/uk/gov/hmcts/darts/common/service/impl/HearingCommonServiceImpl.java
```diff
@@ -69,4 +71,47 @@ private HearingEntity setHearingLastDateModifiedBy(final HearingEntity hearingEn
...
+        if (courtCaseEntity == null) {
+            log.info("Can not link hearing to media {} as CourtCaseEntity is null", mediaEntity.getId());
+            return false;
+        }
+
+        Optional<HearingEntity> hearingEntityOptional = hearingRepository
+            .findHearing(
+                courtCaseEntity,
+                mediaEntity.getCourtroom(),
+                mediaEntity.getStart().toLocalDate()
```
jackmaloney, line 95: Will this link the audio to _any_ hearing that occurred on that day? Not sure that is what we want?

jackmaloney: Maybe this is fine because you're searching by court case.

jackmaloney (approved):

hemantasharma1129 (approved):

PR #2797, opened 2025-04-10
DMP-4911 Include batch size into automated tasks logs

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-4911

### Change description ###

Added a generic message to log the batch size when the task is started but also added a log for each task when it gets data so there is a log for how many were being processed against for the batch size

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/task/runner/impl/AbstractLockableAutomatedTask.java
+++ b/src/main/java/uk/gov/hmcts/darts/task/runner/impl/AbstractLockableAutomatedTask.java
@@ -113,8 +113,7 @@ public abstract class AbstractLockableAutomatedTask<T extends AbstractAutomatedT
                         if (!TRUE.equals(automatedTask.getTaskEnabled())) {
                             log.info("Task: {} is inactive but has been run manually", getTaskName());
                         }
-                        logApi.taskStarted(executionId.get(), this.getTaskName());
-                        log.info("Task: {} running with batch size: {}", getTaskName(), getAutomatedTaskBatchSize());
+                        logApi.taskStarted(executionId.get(), this.getTaskName(), getAutomatedTaskBatchSize());
                         lockService.getLockingTaskExecutor().executeWithLock(createLockableTask(), getLockConfiguration());
                     } else {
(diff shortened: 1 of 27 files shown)


jackmaloney (approved):

src/main/java/uk/gov/hmcts/darts/task/runner/impl/AbstractLockableAutomatedTask.java
```diff
@@ -114,6 +114,7 @@ public void run(boolean isManualRun) {
                             log.info("Task: {} is inactive but has been run manually", getTaskName());
                         }
                         logApi.taskStarted(executionId.get(), this.getTaskName());
+                        log.info("Task: {} running with batch size: {}", getTaskName(), getAutomatedTaskBatchSize());
```
davet1985, line 117: I think this should go into the logApi message above, like this

```
"Task started: run_id={}, task_name={} batch_size={}"
```

davet1985 (changes requested):
Add the batch size to the log in the logApi package instead of adding a new log entry

src/main/java/uk/gov/hmcts/darts/task/runner/impl/AbstractLockableAutomatedTask.java
karen-hedges: The log api is dynatrace, so we don't normally change these logs or the tiles have to change?

davet1985: It's ok to add to the log messages, just not change the format etc. I think we should do this and remove the extra log message.

Ben-Edwards-cgi (approved):

src/main/java/uk/gov/hmcts/darts/task/runner/impl/AbstractLockableAutomatedTask.java
karen-hedges: done

davet1985 (approved):

PR #2800, opened 2025-04-10
DMP-4832: Validate "isHidden: null" in hide media request

Ben-Edwards-cgi:
### Change description ###
# Summary of Git Diff

This diff introduces new functionality to the `MediaControllerAdminPostMediaIntTest` Java class, specifically adding a test to validate the handling of a null value for the `isHidden` field in `MediaHideRequest`. Additionally, it updates the OpenAPI specification to reflect changes related to the required field for `MediaHideRequest`.

## Highlights

- **New Test Case Added**: 
  - A test method `postAdminHideMediaId_usingNullIsHidden_shouldBeRejected` is introduced to ensure that a null value for `isHidden` in `MediaHideRequest` is properly rejected, returning a 400 error with a specific violation message.
  
- **Modification in MediaHideRequest**:
  - The `is_hidden` property in the `MediaHideRequest` schema is now marked as a required field.

- **OpenAPI Specification Updates**:
  - Minor adjustments were made to the formatting of the OpenAPI YAML file, ensuring consistent reference formatting for the `problem.yaml` schema.
  - Error descriptions for 400 responses were included to clarify the nature of the constraints.

This update improves the robustness of the media hiding feature in the application by enforcing stricter validation on input data.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/MediaControllerAdminPostMediaIntTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/audio/controller/MediaControllerAdminPostMediaIntTest.java
@@ -231,6 +231,7 @@ class MediaControllerAdminPostMediaIntTest extends IntegrationBase {
         superAdminUserStub.givenUserIsAuthorised(userIdentity, SecurityRoleEnum.SUPER_USER);
 
         MediaHideRequest mediaHideRequest = new MediaHideRequest();
+        mediaHideRequest.setIsHidden(true);
 
         mockMvc.perform(post(ENDPOINT_URL.replace(
                 MEDIA_ID_SUBSTITUTION_KEY, Integer.valueOf(-12).toString()))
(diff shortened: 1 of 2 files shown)


jackmaloney (approved):

jackmaloney:
This doesn't affect frontend does it? The previous implementation had this defaulting the value to 'true' if it was missing. Are we sure the frontend always passes this value in?

Ben-Edwards-cgi:
> This doesn't affect frontend does it? The previous implementation had this defaulting the value to 'true' if it was missing. Are we sure the frontend always passes this value in?

it should not as this ticket was about making this field not null so a required field, as such I am assuming FE should always provided it @davet1985 please correct me if I am incorrect here.

davet1985:
> > This doesn't affect frontend does it? The previous implementation had this defaulting the value to 'true' if it was missing. Are we sure the frontend always passes this value in?
> 
> it should not as this ticket was about making this field not null so a required field, as such I am assuming FE should always provided it @davet1985 please correct me if I am incorrect here.

The portal will always explicitly set `true` or `false` so this is fine.

jackmaloney:
My interpretation of that ticket was slightly different. I thought it was saying you can explicitly pass 'null' in as the value for isHidden, in which case a NPE would be thrown. But if you didn't pass _any_ value in then that openapi config will default it to true. So feel like this is a potential change in behaviour if there are places where the frontend does not pass this value in.

jackmaloney:
Just seen @davet1985 reply, so looks like we are all good.

PR #2819, opened 2025-04-16
Updated file size request limits

Ben-Edwards-cgi:
Updated file size request limits to support larger then 1MB

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -19,8 +19,8 @@ springdoc:
 spring:
   servlet:
     multipart:
-      max-file-size: 15MB
-      max-request-size: 20MB
+      max-file-size: 25MB
+      max-request-size: 30MB
   lifecycle:
     timeout-per-shutdown-phase: ${GRACEFUL_SHUTDOWN_TIMEOUT:30s}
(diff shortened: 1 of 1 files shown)


src/main/resources/application.yaml
```diff
@@ -17,6 +17,10 @@ springdoc:
   packagesToScan: uk.gov.hmcts.darts
 
 spring:
+  servlet:
+    multipart:
+      max-file-size: 15MB
```
karen-hedges, line 22: Isn't the max size for transcriptions 10MB and 20MB for annotations or the other way around @jackmaloney

jackmaloney (approved):

src/main/resources/application.yaml
jackmaloney: Yes you are correct @karen-hedges , good spot.

karen-hedges: Will this impact audio uploads?

Ben-Edwards-cgi: Audio never gets set to API it is accepted and uploaded on Gateway :)

karen-hedges (approved):

PR #2827, opened 2025-04-22
DMP-4633 Review and resolve sonar issues with the master branch

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-4633

### Change description ###

Fixed major sonar compliants - Rename field 'x' to prevent any misunderstanding/clash with field 'X' and add at least one assertion to this test case. Also high sonar complaint Remove this use of 'createNativeQuery'; it is deprecated.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/audio/helper/TransformedMediaHelper.java
+++ b/src/main/java/uk/gov/hmcts/darts/audio/helper/TransformedMediaHelper.java
@@ -82,7 +82,6 @@ public class TransformedMediaHelper {
         return blobName;
     }
 
-    @Transactional
     public TransformedMediaEntity createTransformedMediaEntity(MediaRequestEntity mediaRequest, String filename,
                                                                OffsetDateTime startTime, OffsetDateTime endTime,
                                                                Long fileSize) {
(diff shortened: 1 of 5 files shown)


hemantasharma1129 (approved):

src/main/java/uk/gov/hmcts/darts/audio/helper/TransformedMediaHelper.java
```diff
@@ -82,7 +82,6 @@ public String saveToStorage(MediaRequestEntity mediaRequest, InputStream inputSt
         return blobName;
     }
 
-    @Transactional
```
davet1985, line 85: This being remove seems out of place in the PR. Is it intentional?

karen-hedges: It was a sonar complaint because @Transactional doesn't work when its called by the same class. Its called by the method saveToStorage which is already wrapped in an @Transactional so I think its just a legacy issue when these generic classes were created.

karen-hedges: The sonar complaint is https://sonarcloud.io/project/issues?impactSeverities=HIGH&issueStatuses=OPEN%2CCONFIRMED&id=uk.gov.hmcts.reform%3Adarts-api&open=AZGPVSipsRHppe0VGu9L

davet1985: Thanks for explaining 👍

davet1985 (approved):

PR #2858, opened 2025-04-29
DMP-5046: Make logs go into AppInsights for Dev environment

Ben-Edwards-cgi:
### Change description ###
# Summary of Git Diff

This Git diff reflects changes made to several Jenkinsfiles and YAML configuration files, primarily focusing on updating the alias for the `app-insights-connection-string` secret to `APPLICATIONINSIGHTS_CONNECTION_STRING`.

## Highlights
- **Jenkinsfiles Updated**:
  - **Jenkinsfile_CNP**
    - Changed the alias from `app-insights-connection-string` to `APPLICATIONINSIGHTS_CONNECTION_STRING`.
  - **Jenkinsfile_nightly**
    - Similar update as in Jenkinsfile_CNP.
  - **Jenkinsfile_parameterized**
    - Alias change for the `app-insights-connection-string` to `APPLICATIONINSIGHTS_CONNECTION_STRING`.

- **YAML Configuration Files Updated**:
  - **values.dev.template.yaml**
    - Commented out line for `app-insights-connection-string` updated to reflect the new alias.
  - **values.yaml**
    - Updated in two places:
      - Under the `java` section and the `function` section, the alias for `app-insights-connection-string` changed to `APPLICATIONINSIGHTS_CONNECTION_STRING`.

```
[ ] Yes
[X] No
```

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -12,7 +12,7 @@ def component = "api"
 def branchesToSync = ['demo', 'perftest', 'ithc']
 
 def secrets = [
-    'darts-${env}': [
+    'darts-stg': [
         secret('GovukNotifyTestApiKey', 'GOVUK_NOTIFY_API_KEY'),
         secret('app-insights-connection-string', 'APPLICATIONINSIGHTS_CONNECTION_STRING'),
         secret('AzureAdB2CTenantId', 'AAD_B2C_TENANT_ID'),
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -222,8 +222,10 @@ withPipeline(type, product, component) {
   }
 
   afterAlways('functionalTest:dev') {
-    stage('SKIP_TESTS OR SKIP_SONAR IS TRUE') {
-      error("Build can not fully complete when SKIP_TESTS or SKIP_SONAR is set to true")
+    if(env.SKIP_TESTS || env.SKIP_SONAR) {
+      stage('SKIP_TESTS OR SKIP_SONAR IS TRUE') {
+        error("Build can not fully complete when SKIP_TESTS or SKIP_SONAR is set to true")
+      }
     }
   }
 
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -100,8 +100,8 @@ GradleBuilder builder = new GradleBuilder(this, product)
 
 withPipeline(type, product, component) {
     onPR {
-        Object obj = secrets.remove('darts-${env}');
-        map.put('darts-stg', obj);
+        Object obj = secrets.remove('darts-${env}')
+        map.put('darts-stg', obj)
     }
     enableDbMigration(product)
     loadVaultSecrets(secrets)
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -12,7 +12,7 @@ def component = "api"
 def branchesToSync = ['demo', 'perftest', 'ithc']
 
 def secrets = [
-    'darts-stg': [
+    'darts-${env}': [
         secret('GovukNotifyTestApiKey', 'GOVUK_NOTIFY_API_KEY'),
         secret('app-insights-connection-string', 'APPLICATIONINSIGHTS_CONNECTION_STRING'),
         secret('AzureAdB2CTenantId', 'AAD_B2C_TENANT_ID'),
@@ -99,6 +99,10 @@ def determineDevEnvironmentDeployment() {
 GradleBuilder builder = new GradleBuilder(this, product)
 
 withPipeline(type, product, component) {
+    onPR {
+        Object obj = secrets.remove('darts-${env}');
+        map.put('darts-stg', obj);
+    }
     enableDbMigration(product)
     loadVaultSecrets(secrets)
     enableSlackNotifications('#darts-builds')
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -3,7 +3,7 @@ properties([
     parameters([
         booleanParam(name: 'SKIP_TESTS', defaultValue: false, description: 'Skip all tests'),
         booleanParam(name: 'SKIP_SONAR', defaultValue: false, description: 'Skip sonar'),
-        booleanParam(name: 'FAIL_FAST', defaultValue: true, description: 'Should tests fail fast'),
+        booleanParam(name: 'FAIL_FAST', defaultValue: false, description: 'Should tests fail fast'),
     ])
 ])
 @Library("Infrastructure")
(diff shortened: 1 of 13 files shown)


charts/darts-api/values.dev.template.yaml
```diff
@@ -115,6 +115,7 @@ java:
     IS_MOCK_ARM_RPO_DOWNLOAD_CSV: false
     AZCOPY_LOG_LEVEL: "--log-level=ERROR"
     AZCOPY_OUTPUT_LEVEL: "--output-level=essential"
+    ROLE_NAME: "DARTS API ${CHANGE_ID}"
```
davet1985, line 118:
```suggestion
    ROLE_NAME: "DARTS API PR-${CHANGE_ID}"
```

Ben-Edwards-cgi: Agreed updated

davet1985 (approved):

Jenkinsfile_CNP
```diff
@@ -1,5 +1,11 @@
 #!groovy
-
+properties([
+    parameters([
+        booleanParam(name: 'SKIP_TESTS', defaultValue: false, description: 'Skip all tests'),
+        booleanParam(name: 'SKIP_SONAR', defaultValue: false, description: 'Skip sonar'),
+        booleanParam(name: 'FAIL_FAST', defaultValue: true, description: 'Should tests fail fast'),
+    ])
+])
```
davet1985, line 8: I don't think we should add this, it could easily be mis-used.

davet1985 (changes requested):
I don't think we should have the options for skipping tests and sonar, it's there for a reason and those parameters will be mis-used.

Ben-Edwards-cgi:
> I don't think we should have the options for skipping tests and sonar, it's there for a reason and those parameters will be mis-used.

The reason why this was added is sometimes we can only test issues once deployed, so we need to update and deploy consistently, in order to verify the change. (The final merge should always have these params set to false)

For Sonar sometimes we get code coverage errors on things we do not need to resolve. For example log statements, IO exceptions.


When I will do is update the build to fail it on GitHub at the very end if these values are set to ensure we can deploy into dev without them but we can not merge without them

build.gradle
```diff
@@ -1,46 +1,51 @@
 buildscript {
-    dependencies {
-        classpath("org.flywaydb:flyway-database-postgresql:11.8.0")
-    }
+  dependencies {
```
jackmaloney, line 2: Are all these formatting changes necessary? They make it difficult to review and you lose the history of the last person to change a line making it abit more tricky to trace changes back.

src/integrationTest/java/uk/gov/hmcts/JenkinsTest.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.hmcts;
+
+import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.Test;
+
+// This is a placeholder test to ensure the Jenkins pipeline runs successfully.
+//This calss can not be moved into the darts package else it will be excluded
```
jackmaloney, line 7: typo in 'class'. Same in below files.

src/functionalTest/java/uk/gov/hmcts/JenkinsTest.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.hmcts;
+
+import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.Test;
+
+// This is a placeholder test to ensure the Jenkins pipeline runs successfully.
+//This calss can not be moved into the darts package else it will be excluded
+@Slf4j
+class JenkinsTest {
```
jackmaloney, line 9: What's the purpose of having one of these in each package?

Jenkinsfile_CNP
jackmaloney: Why do we want to skip sonar? And what does the fail fast one do?

Ben-Edwards-cgi: FAST_FAIL will stop running fail the test run as soon as the first test fails instead of after running them all.
Essentially meaning we get easier feedback.

Sonar just saves us a few mins

davet1985 (approved):
Approved, but @jackmaloney has outstanding comments

src/functionalTest/java/uk/gov/hmcts/JenkinsTest.java
Ben-Edwards-cgi: This ensures the test report is generated allowing the pipeline to continue even when SKIP_TESTS is true

Jenkinsfile_CNP
jackmaloney: I can sort of see the value in the skip tests one if you wanted to quickly try out a change on the Dev environment. But can't see the value in the other 2. Surely if you've got failing tests you would want to know all of the ones that are failing instead of just one. And Sonar issues should be quick to resolve anyway. @davet1985 @hemantasharma1129 any thoughts? I'd suggest removing `SKIP_SONAR` and `FAIL_FAST` from this config.

davet1985: Sonar can be a pain like the tests if you want to get something out to the dev env quickly. I had a chat with @Ben-Edwards-cgi and we agreed that you wouldn't want to allow the build to pass if you are skipping sonar or tests.

Maybe we should set the `FAIL_FAST` default to false, to continue running as we are now, but allow the option?

Ben-Edwards-cgi: I can see both ways,

I like FAIL_FAST being true just so the build fails as soon as it finds an error so I know I need to fix the code instead of having to wait the full 40 mins to know

On the other hand knowing the full set of test failures up front is handy to know so we can fix them all in one go.

I personally prefer fail fast so flakey tests are found earlier so a re-build can happen quicker, and if one integration test fails I usually run full integration locally to view the logs easier anyway.

But I am easy either way so happy to update the default to either

build.gradle
Ben-Edwards-cgi: This is just InteliJ default formatting updating it.
I can revert if we know what the HMCTS style rules are for this

src/integrationTest/java/uk/gov/hmcts/JenkinsTest.java
Ben-Edwards-cgi: Updated

Jenkinsfile_CNP
davet1985: I think we keep it the same right now ie. default to false. @jackmaloney?

jackmaloney: yes please

Ben-Edwards-cgi: Cool updated :)

jackmaloney (approved):

PR #2867, opened 2025-05-02
Order by update

Ben-Edwards-cgi:
Updated order by so hearings get ordered by createdDateTime, then hearingDate then id

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/entity/MediaEntity.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/entity/MediaEntity.java
@@ -207,6 +207,6 @@ public class MediaEntity extends CreatedModifiedBaseEntity
      */
     @Deprecated
     public HearingEntity getHearing() {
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/darts/common/entity/MediaEntity.java
```diff
@@ -207,6 +207,6 @@ public boolean isCurrent() {
      */
     @Deprecated
```
karen-hedges, line 208: Isn't this method deprecated so shouldn't you be using the correct method not amending a deprecated method

Ben-Edwards-cgi: This is around ManyToMany mappings being treated as ManyToOne in code.
We have a backlog ticket to resolve these

jackmaloney (approved):

davet1985 (approved):

PR #2886, opened 2025-05-09
Fixing transcription documents search when tra.requested_by is null

davet1985:

N/A

### Change description ###

To ensure we can see transcription documents even if the associated transcription requested by is null

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/common/repository/TranscriptionDocumentTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/common/repository/TranscriptionDocumentTest.java
@@ -1,6 +1,7 @@
 package uk.gov.hmcts.darts.common.repository;
 
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.EnumSource;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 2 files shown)


src/integrationTest/java/uk/gov/hmcts/darts/common/repository/TranscriptionDocumentTest.java
```diff
@@ -815,6 +817,19 @@ void testFindTranscriptionDocumentWithAllQueryParameters(TestType testType) {
...
+    void findTranscriptionMedia_shouldFindResult_whenTranscriptionRequestedByIsNull() {
+        dataSetup(TestType.LEGACY);
+        generatedDocumentEntities.forEach(generatedDocumentEntity -> {
+            generatedDocumentEntity.getTranscription().setRequestedBy(null);
+            dartsDatabase.save(generatedDocumentEntity.getTranscription());
+        });
+        List<TranscriptionDocumentResult> transcriptionDocumentResults
+            = transcriptionDocumentRepository
+            .findTranscriptionMedia(null, null, null, null, null, null, null, null, true);
+        Assertions.assertEquals(4, transcriptionDocumentResults.size());
```
Ben-Edwards-cgi, line 830: This does not verify t.requestedBy is called or the value that is returned?

davet1985: The requested by value is neither returned or used. Without adding it explicitly in the query, JPA would JOIN from transcription to user_account and exclude any rows where it is null. Before the change, this assert would fail because the query would return 0 rows.

karen-hedges (approved):

Ben-Edwards-cgi (approved):

jackmaloney (approved):

PR #2892, opened 2025-05-12
DMP-5063: Transcription marked for deletion ts genrated with is marked for deletion flag false

Ben-Edwards-cgi:
### Change description ###
# Summary of Git Diff

This Git diff reflects changes made to the `AdminTranscriptionServiceImpl.java` and `AdminTranscriptionServiceTest.java` files in the transcription service of the HMCTS DARTS project. The modifications focus on the handling of transcription document hiding logic, particularly around the management of deletion markers and user accounts.

### Changes in `AdminTranscriptionServiceImpl.java`
- **User Account Handling**: 
  - Replaced direct calls to `userIdentity.getUserAccount()` with a stored variable `currentUser` for clarity and performance.
  
- **Object Admin Action Entity Modifications**:
  - The method now checks if the `objectHiddenReason` is marked for deletion before setting the `markedForManualDeletion` flag.
  - If marked for deletion, the fields for manual deletion are set using `currentUser`.

### Changes in `AdminTranscriptionServiceTest.java`
- **Test Method Enhancements**:
  - Introduced new test cases to verify the behavior of the `hideOrShowTranscriptionDocumentById` method regarding the deletion flag.
  - Modified the existing test cases to use an assertion method `assertTestTranscriptionDocumentHide`, which checks both the hidden state and the deletion status effectively.
  
- **Assertions on Object Admin Action**:
  - Improved assertions to check for null values and correctness of the manual deletion fields based on the `isMarkedForDeletion` parameter.

These changes enhance the robustness of the transcription document management by ensuring that the deletion logic is accurately reflected in both the service implementation and its associated tests.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/AdminTranscriptionServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/AdminTranscriptionServiceImpl.java
@@ -207,18 +207,21 @@ public class AdminTranscriptionServiceImpl implements AdminTranscriptionService 
 
                 auditApi.record(HIDE_TRANSCRIPTION);
 
+                ObjectHiddenReasonEntity objectHiddenReason = objectHiddenReasonEntity.get();
+                UserAccountEntity currentUser = userIdentity.getUserAccount();
                 // on hiding add the relevant hide record
                 ObjectAdminActionEntity objectAdminActionEntity = new ObjectAdminActionEntity();
-                objectAdminActionEntity.setObjectHiddenReason(objectHiddenReasonEntity.get());
+                objectAdminActionEntity.setObjectHiddenReason(objectHiddenReason);
                 objectAdminActionEntity.setTicketReference(transcriptionDocumentHideRequest.getAdminAction().getTicketReference());
                 objectAdminActionEntity.setComments(transcriptionDocumentHideRequest.getAdminAction().getComments());
                 objectAdminActionEntity.setTranscriptionDocument(documentEntity);
-                objectAdminActionEntity.setHiddenBy(userIdentity.getUserAccount());
+                objectAdminActionEntity.setHiddenBy(currentUser);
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/AdminTranscriptionServiceImpl.java
```diff
@@ -207,18 +207,21 @@ public TranscriptionDocumentHideResponse hideOrShowTranscriptionDocumentById(Lon
...
                 objectAdminActionEntity.setComments(transcriptionDocumentHideRequest.getAdminAction().getComments());
                 objectAdminActionEntity.setTranscriptionDocument(documentEntity);
-                objectAdminActionEntity.setHiddenBy(userIdentity.getUserAccount());
+                objectAdminActionEntity.setHiddenBy(currentUser);
                 objectAdminActionEntity.setHiddenDateTime(OffsetDateTime.now());
-                objectAdminActionEntity.setMarkedForManualDeletion(false);
-                objectAdminActionEntity.setMarkedForManualDelBy(userIdentity.getUserAccount());
-                objectAdminActionEntity.setMarkedForManualDelDateTime(OffsetDateTime.now());
-
+                if (objectHiddenReason.isMarkedForDeletion()) {
```
jackmaloney, line 220: Does this implicitly set the 'MarkedForManualDeletion' field to false if it comes down the isMarkedForDeletion() = false route? If not, consider adding an else statement to this if block so it behaves the same as before and sets this field to `false`.

Ben-Edwards-cgi: The field in the entity is a boolean field so is always set to false unless overridden.
Opposed to Boolean which is defaulted to null

jackmaloney (approved):

hemantasharma1129 (approved):

PR #2897, opened 2025-05-13
DMP-5100 ARM Replay needs to reset input_upload_processed_ts

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-5100

### Change description ###

Changed the location where the input_upload_processed_ts is reset to when the replay job is called. Moved the private methods below the tests. I also had to fix the issue where the ids are now longs not ints on the EOD records. 

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmRpoServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmRpoServiceImpl.java
@@ -142,7 +142,6 @@ public class ArmRpoServiceImpl implements ArmRpoService {
                     externalObjectDirectoryEntity.setStatus(EodHelper.storedStatus());
                 } else {
                     externalObjectDirectoryEntity.setStatus(EodHelper.armReplayStatus());
-                    externalObjectDirectoryEntity.setInputUploadProcessedTs(null);
                 }
             }
         );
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/hmcts/darts/arm/service/impl/ArmRpoServiceImpl.java
```diff
@@ -142,7 +142,6 @@ public void reconcileArmRpoCsvData(ArmRpoExecutionDetailEntity armRpoExecutionDe
                     externalObjectDirectoryEntity.setStatus(EodHelper.storedStatus());
                 } else {
                     externalObjectDirectoryEntity.setStatus(EodHelper.armReplayStatus());
-                    externalObjectDirectoryEntity.setInputUploadProcessedTs(null);
```
hemantasharma1129, line 145: Where are we setting the null value in InputuploadTs? I thought we will do it part of ARM_Replay job but cant find that change

karen-hedges: I have removed it setting it to null here and moved it to the replay job when it does the update

hemantasharma1129 (approved):

jackmaloney (approved):

PR #2901, opened 2025-05-13
DMP-3411: Implement endpoint to make event current version

Ben-Edwards-cgi:
### Change description ###

This is needed to add an additional layer of protection such that event_id 0 that is_current = false (which should not happen) does cause catastrophic data issues by linking all event_ids = 0 and setting all other ones to false 

# Summary of Git Diff

The Git Diff reflects a series of changes made to the event management service in the codebase. Key changes include updating error handling for API requests, modifying test cases, and adding new error codes for specific scenarios.

## Highlights

- **Error Handling Update**: The error response for when `isCurrent` is set to `false` has been changed from HTTP status `422 Unprocessable Entity` to `400 Bad Request` in both the integration test and the event service implementation.
  
- **New Error Code**: Introduced a new error code `CAN_NOT_UPDATE_EVENT_ID_0` in the `EventError` enum to handle the specific case when an event ID is `0`.

- **Integration Test Updates**:
  - Updated the test case to expect a `400 Bad Request` status instead of `422` when `isCurrent` is set to `false`.
  - Added a new test case to check for the `CAN_NOT_UPDATE_EVENT_ID_0` error when trying to patch an event with an ID of `0`.

- **OpenAPI Update**: The OpenAPI specification has been updated to include the new error code `CAN_NOT_UPDATE_EVENT_ID_0`.

- **Consistent Error Responses**: The changes ensure that error responses are consistent across the API and the service layer, enhancing the robustness of error handling in the application.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerPatchAdminEventByIdIntTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/darts/event/controller/EventsControllerPatchAdminEventByIdIntTest.java
@@ -86,13 +86,15 @@ class EventsControllerPatchAdminEventByIdIntTest extends IntegrationBase {
     }
 
     @Test
-    void shouldReturn422_whenIsCurrentIsSetToFalse() throws Exception {
+    void shouldReturn400_whenIsCurrentIsSetToFalse() throws Exception {
         given.anAuthenticatedUserWithGlobalAccessAndRole(SecurityRoleEnum.SUPER_ADMIN);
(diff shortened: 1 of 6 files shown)


jackmaloney:
Please can you add a description of why we need this change? 
On a side note, I find the AI generated PR descriptions pretty pointless. Sometimes it would be good if these can be replaced with an actual written explanation of why a change is needed.

hemantasharma1129 (approved):

PR #2922, opened 2025-05-16
Cleaned up error logging

Ben-Edwards-cgi:
Cleaned up error logging so that we only log errors we need to action


src/main/java/uk/gov/hmcts/darts/common/exception/DartsApiTrait.java
```diff
@@ -78,6 +78,6 @@ static Problem getContentForException(DartsApiException exception) {
 
     private static boolean shouldLogException(DartsApiException exception) {
         DartsApiError error = exception.getError();
-        return error != null && error.getHttpStatus() != HttpStatus.UNPROCESSABLE_ENTITY;
+        return error == null || (error.shouldLogException() && error.getHttpStatus() != HttpStatus.UNPROCESSABLE_ENTITY);
```
jackmaloney, line 81: Could error being null cause NPE's? Wonder if a couple of places in this class should be tweaked to handle this scenario?

Ben-Edwards-cgi: If we do throw a DartsApiException but do not set the Error code for some reason we still want this to log to we can see why it is missing the code hence the error == null change.

For NPE explicitly this wont chuck a DartsApiException (which is what we handle here) so will be processed via the standard exception route.

I am unsure if this route has built in error logging, but I have seen NPE in the Azure logs before

jackmaloney (approved):

PR #2926, opened 2025-05-16
DMP-5113 ARM Batch Processor is not using the correct value for getting IU files

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-5113

### Change description ###

Added fix so that the correct no of IU files is retrieved between the batch size and the continuation token max size

```
[ ] Yes
[ x] No
```


karen-hedges:
not required

PR #2944, opened 2025-05-22
DMP-3393 Get node registrations endpoint

jackmaloney:
https://tools.hmcts.net/jira/browse/DMP-3393

### Change description ###
Create a new endpoint for super admins to get details about node registrations

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/noderegistration/service/impl/NodeRegistrationServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/noderegistration/service/impl/NodeRegistrationServiceImpl.java
@@ -9,7 +9,12 @@ import uk.gov.hmcts.darts.common.entity.NodeRegisterEntity;
 import uk.gov.hmcts.darts.common.entity.UserAccountEntity;
 import uk.gov.hmcts.darts.common.repository.NodeRegisterRepository;
 import uk.gov.hmcts.darts.common.service.RetrieveCoreObjectService;
+import uk.gov.hmcts.darts.noderegistration.model.GetNodeRegisterManagementResponse;
+import uk.gov.hmcts.darts.noderegistration.model.GetNodeRegisterManagementResponseCourthouse;
+import uk.gov.hmcts.darts.noderegistration.model.GetNodeRegisterManagementResponseCourtroom;
 import uk.gov.hmcts.darts.noderegistration.service.NodeRegistrationService;
+
+import java.util.List;
 
 @Service
 @RequiredArgsConstructor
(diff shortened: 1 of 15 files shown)


src/main/resources/openapi/noderegistration.yaml
```diff
@@ -97,6 +117,45 @@ components:
           type: integer
           example: 42
 
+    GetNodeRegisterManagementResponse:
+      type: object
+      properties:
```
Ben-Edwards-cgi, line 122: Is it worth adding in the required: section just so we know what values you always get and what ones you might not?

src/main/java/uk/gov/hmcts/darts/noderegistration/service/impl/NodeRegistrationServiceImpl.java
```diff
@@ -34,4 +39,40 @@ public Integer registerDevices(String nodeType, String courthouse, String courtR
         nodeRegisterEntity.setCreatedBy(userAccount);
         return nodeRegisterRepository.saveAndFlush(nodeRegisterEntity).getNodeId();
     }
+
+    @Override
+    public List<GetNodeRegisterManagementResponse> getNodeRegisterDevices() {
+        return nodeRegisterRepository.findAll().stream()
```
Ben-Edwards-cgi, line 45: Could benefit from an order by to keep consistent order

src/test/java/uk/gov/hmcts/darts/noderegistration/service/impl/NodeRegistrationServiceImplTest.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+@SuppressWarnings({"PMD.AvoidUsingHardCodedIP"})
+class NodeRegistrationServiceImplTest {
```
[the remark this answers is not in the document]
jackmaloney: Actually, these methods should probably be marked as private. I can make that change if you're in agreement. There are already dedicated unit tests further down that test the null courthouse and null courtroom scenario.

src/main/resources/openapi/noderegistration.yaml
Ben-Edwards-cgi: From a response viewpoint the client needs to know what they have to send us, but they also need to know what fields are always returned, optionally returned and what structure/validation rules apply to the data.

Less important for what we are doing here as we control both FE and BE but if we ever exposed APIs downstream these validation rules on responses become more important :)

src/test/java/uk/gov/hmcts/darts/noderegistration/service/impl/NodeRegistrationServiceImplTest.java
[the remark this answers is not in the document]
Ben-Edwards-cgi: I'm happy for them to be private as long as the method is fully covered in the other unit tests :)

src/main/resources/openapi/noderegistration.yaml
jackmaloney: We don't seem to have done this on any other endpoint and this returns all fields anyway so don't think there's any value in doinf this. There are no plans to expose these API's downstream.

hemantasharma1129 (approved):

src/main/resources/openapi/noderegistration.yaml
davet1985: I agree we could do this but it differs from what we currently have so I think it's best to keep it inline right now. In the setup that we have it doesn't offer much value.

davet1985 (approved):

Ben-Edwards-cgi (approved):

PR #2953, opened 2025-05-27
DMP-5126: Exclude event_id = 0 from duplicate checks

Ben-Edwards-cgi:
### Change description ###
When a event with a event_id = 0 is sent to gateway, this gets accepted and does some async processing.

This processing perform a duplicate check, to prevent processing on events that match previously submitted events.

It has been noticed in production that when this check is performed on an event_id = 0 it is taking a significant amount of time and spikes database CPU.

This is because we have 108 million records in production with event_id = 0.

This ticket is to remove the duplicate check on event_id =0 as this event id should always be unique as it comes from a bug in xhibit.

To verify this change, we will need to submit two events with event_id = 0 that met the duplicate logic (Same message_id, event_id and event_text). 

When the second event comes in (that previously would be marked as a duplicate and deleted). We should process the event as if it was not a duplicate.

```
[ ] Yes
[X] No
```

### DIFF
--- a/infrastructure/import.tf
+++ b/infrastructure/import.tf
@@ -1,0 +1,4 @@
+import {
+  to = module.postgresql_flexible.azurerm_postgresql_flexible_server.pgsql_server
+  id = "/subscriptions/3eec5bde-7feb-4566-bfb6-805df6e10b90/resourceGroups/darts-test-rg/providers/Microsoft.DBforPostgreSQL/flexibleServers/darts-api-test"
+}


src/main/java/uk/gov/hmcts/darts/event/service/impl/RemoveDuplicateEventsProcessorImpl.java
```diff
@@ -24,6 +24,10 @@ public class RemoveDuplicateEventsProcessorImpl implements RemoveDuplicateEvents
     @Override
     @Transactional
     public boolean findAndRemoveDuplicateEvent(Integer eventId) {
+        if (eventId == 0) {
+            //No need to continue if event_id = 0 as this can not have duplicates
```
davet1985, line 28: `cannot` typo

Ben-Edwards-cgi: Updated :)

src/main/java/uk/gov/hmcts/darts/event/service/impl/RemoveDuplicateEventsProcessorImpl.java
```diff
@@ -43,7 +49,7 @@ public boolean findAndRemoveDuplicateEvent(Integer eventId) {
         eventRepository.deleteAllById(eventEntitiesIdsToDelete);
         eventRepository.flush();
 
-        log.info("Duplicate events found. Removing the following events {}", eventEntitiesIdsToDelete);
+        log.info("Duplicate events found for eveId {}. Removing the following events {}.", eveIdToKeep, eventEntitiesIdsToDelete);
```
jackmaloney: Is it possible to log out the message ID of the one affected? Would have made my life so much easier today.

jackmaloney (approved):

PR #2962, opened 2025-05-29
DMP-5129: Improve duplicate case check query for eventId = 0

Ben-Edwards-cgi:
### Change description ###
An edge case has been observed on production where an event with eventId = 0 has been incorrectly deleted.

Here is a summary of the issue:

At 14:37:24 an event with event id zero was received that took more than 30 seconds to be processed. The request timed out and an error was returned back to XHIBIT at 14:37:54, causing them to retry that event, which was received at 14:38:01. However, the original request was actually processed and saved to the database after 30 seconds at 14:37:58. This meant that this retried event was a duplicate. However, there was an issue with the duplicate check query in that if an unrelated request with event id zero was received before the duplicate had been cleaned up, what actually happened was this unrelated request grouped all duplicate event id zero's and removed them. Therefore, what we have seen is that for the request highlighted above, both the original and duplicate were removed. Therefore, as things stand there is a risk of a loss of data with this query for event id's zero.

 

DARTS should not be deleting events that have the same event id if the message id and event text are different. 
h4. Prerequisites:
 # Multiple events exist with event id = 0, with message id and event text set to the same values for each of these events (For testing, inject these via SQL into the database, not through Postman otherwise the duplicate logic will kick in)  e.g.

 
||No.||Event ID||Message ID||Event Text||
|1|0|ABC123|Test 1|
|2|0|ABC123|Test 1|
|3|0|ZYX987|Test 2|

Ensure the 3 above have slightly different created dates, in ascending order from 1 - 3. 

 
h4. Steps to reproduce:
 # Send an event in with the following detail:

||Event ID||Message ID||Event Text||
|0|ZYX987|Test 2|
h4. Expected results:
 # The newly inserted event with messageId ZYX987 is deleted
 # A log message with the event id, message id and event text is printed with the details of the event that has been deleted

||No.||Event ID||Message ID||Event Text||
|1|0|ABC123|Test 1|
|2|0|ABC123|Test 1|
|3|0|ZYX987|Test 2|
h4. Actual results:
 # The 2 events with message id ZYX987 and 1 of the ABC123 events are incorrectly deleted

||No.||Event ID||Message ID||Event Text||
|1|0|ABC123|Test 1|
h4. Impact:
 # Loss of data

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/repository/EventRepository.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/repository/EventRepository.java
@@ -128,18 +128,12 @@ public interface EventRepository extends JpaRepository<EventEntity, Long> {
     List<Long> findAllByEventStatusAndNotCourtrooms(Integer statusNumber, List<String> courtroomNumbers, Limit limit);
 
     @Query("""
-                        SELECT e3.id from EventEntity e3
-                        JOIN (                        
-                            SELECT e.eventId as eventId, e.messageId as messageId, e.eventText as eventText
-                            FROM EventEntity e
-                            WHERE e.eventId IS NOT NULL and e.messageId IS NOT NULL and e.eventId = :eventId
-                            GROUP BY e.eventId, e.messageId, e.eventText
-                            HAVING COUNT(e) > 1) e2
-                         ON e2.eventId = e3.eventId and e2.messageId = e3.messageId and e2.eventText = e3.eventText
-                         WHERE e3.eventId >= :eventId
-                         ORDER BY e3.createdDateTime ASC  
+        SELECT e3.id from EventEntity e3
+        WHERE e3.eventId = :eventId and e3.messageId = :messageId and e3.eventText = :eventText
+        ORDER BY e3.createdDateTime ASC  
         """)
-    List<Long> findDuplicateEventIds(Integer eventId);
+    List<Long> findDuplicateEventIds(Integer eventId, String messageId, String eventText);
+
 
     @Query("""
            SELECT ee FROM EventEntity ee
--- a/src/main/resources/db/migration/common/V1_476__new_index.sql
+++ b/src/main/resources/db/migration/common/V1_476__new_index.sql
@@ -1,0 +1,1 @@
+CREATE INDEX event_event_id_idx ON darts."event" (event_id,message_id,event_text);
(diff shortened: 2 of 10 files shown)


src/main/java/uk/gov/hmcts/darts/event/helper/DarNotifyAsyncHelper.java
```diff
@@ -53,9 +53,11 @@ public void notifyDarPcAsync(DarNotifyApplicationEvent event, List<String> openC
 
         try {
             dartsGatewayClient.darNotify(darNotifyEvent);
-            log.info("DarNotify request sent successfully: event_id={}, notification_url={}", event.getDartsEvent().getEventId(), notificationUrl);
+            log.info("DarNotify request sent successfully: event_id={}, notification_url={}, event_type={}",
```
jackmaloney, line 56: I wasn't 100% sure if this was being used by Dynatrace or not. I've got a separate PR to improve the logs here anyway, so maybe remove from this PR? My PR for reference: https://github.com/hmcts/darts-api/pull/2963

src/main/resources/db/migration/common/V1_476__new_index.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX event_event_id_idx ON darts."event" (event_id,message_id,event_text);
```
jackmaloney, line 1: Can you just run this one past Barry please and make sure he's happy for it to go in?

src/main/java/uk/gov/hmcts/darts/common/repository/EventRepository.java
```diff
@@ -128,18 +128,14 @@ AND e.courtroom.name not in (:courtroomNumbers)
...
-                            WHERE e.eventId IS NOT NULL and e.messageId IS NOT NULL and e.eventId = :eventId
-                            GROUP BY e.eventId, e.messageId, e.eventText
-                            HAVING COUNT(e) > 1) e2
-                         ON e2.eventId = e3.eventId and e2.messageId = e3.messageId and e2.eventText = e3.eventText
-                         WHERE e3.eventId = :eventId
-                         ORDER BY e3.createdDateTime ASC  
+        SELECT e3.id from EventEntity e3
+        WHERE e3.eventId = :eventId 
+          and e3.messageId = :messageId 
+          and ((:eventText is null and e3.eventText is null) or (:eventText is not null and e3.eventText = :eventText))
```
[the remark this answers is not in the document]
Ben-Edwards-cgi: Yes but I don't know what performance impact this will have.

As if :eventText is null but e3.eventText is not null it will still go to the second or query and start scanning.
The :eventText is not null is designed to stop it from scanning the other rows

jackmaloney (approved):

PR #2982, opened 2025-06-13
DMP-5162 Better differentiate between 403 and 401 errors

davet1985:

https://tools.hmcts.net/jira/browse/DMP-5162

### Change description ###

- amending `USER_DETAILS_INVALID` error to return 401
- amending request filter to only set 403 error when the user is inactive
- otherwise 401 used as the user could not be found and should authenticate again

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/authorisation/component/impl/UserIdentityImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/authorisation/component/impl/UserIdentityImpl.java
@@ -22,6 +22,7 @@ import java.util.stream.Collectors;
 
 import static java.util.Objects.nonNull;
 import static uk.gov.hmcts.darts.authorisation.exception.AuthorisationError.USER_DETAILS_INVALID;
+import static uk.gov.hmcts.darts.authorisation.exception.AuthorisationError.USER_NOT_ACTIVE;
 
 @Component
 @AllArgsConstructor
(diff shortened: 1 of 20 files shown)


src/main/java/uk/gov/hmcts/darts/authorisation/component/impl/UserIdentityImpl.java
```diff
@@ -66,9 +73,15 @@ public Optional<UserAccountEntity> getUserAccountOptional(Jwt jwt) {
...
-            .findFirst();
+        List<UserAccountEntity> userAccounts = userAccountRepository.findByEmailAddressIgnoreCase(emailAddressFromToken);
+
+        var activeUserAccount = userAccounts.stream().filter(UserAccountEntity::isActive).findFirst();
+        if (activeUserAccount.isPresent()) {
+            return activeUserAccount;
+        }
+
+        // return inactive user account if no active user account found, else empty optional
+        return userAccounts.stream().filter(u -> !u.isActive()).findFirst();
```
Ben-Edwards-cgi, line 84: This filter is not needed.

We know only inactive accounts can be  returned here due to the previous is active check

src/main/java/uk/gov/hmcts/darts/common/repository/UserAccountRepository.java
```diff
@@ -39,16 +39,6 @@ public interface UserAccountRepository extends
...
     List<UserAccountEntity> findByRoleAndCourthouse(int securityRole, CourthouseEntity courthouse, Set<UserAccountEntity> excludingUsers);
 
-    @Query("""
-        SELECT userAccount
-        FROM UserAccountEntity userAccount
-        WHERE userAccount.active = true
-        AND userAccount.isSystemUser = true
-        AND userAccount.accountGuid = :uuid
-        """)
-    UserAccountEntity findSystemUser(String uuid);
```
davet1985, line 49: This was not used.

PR #2987, opened 2025-06-17
Dmp 5090 fix deactivate own account error

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-5090

### Change description ###

Added restriction to stop an admin user deactivating themselves

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/darts/usermanagement/validator/NotSameUserValidator.java
```diff
@@ -0,0 +1,28 @@
...
+
+@Component
+@RequiredArgsConstructor
+public class NotSameUserValidator implements Validator<IdRequest<UserPatch, Integer>> {
+
+    private final AuthorisationApi authorisationApi;
+
+    @Override
+    public void validate(IdRequest<UserPatch, Integer> userPatch) {
+        if (!(userPatch.getPayload().getActive() != null && !userPatch.getPayload().getActive())) {
```
jackmaloney, line 20: All these not's is making my head hurt! Is there anyway of tidying up this logic to make it abit easier to read as double negatives can be really confusing. 

I think you can do this by: 

```
if (userPatch.getPayload().getActive() != null && !userPatch.getPayload().getActive())) {
        var currentUser = authorisationApi.getCurrentUser();
        if (currentUser.getId().equals(userPatch.getId())) {
            throw new DartsApiException(AuthorisationError.UNABLE_TO_DEACTIVATE_USER);
        }
    }
```

PR #3251, opened 2026-04-23
DMP-5531 Update retention confidence categories

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-5531

### Change description ###

Updated ret_conf_score to 0 for categories, 4, 9 and 10. Deleted 2 digit categories greater than 10.

```
[ ] Yes
[x] No
```


src/main/resources/db/migration/common/V1_508__remove_legacy_retention_categories.sql
```diff
@@ -0,0 +1,7 @@
+DELETE
+FROM retention_confidence_category_mapper
+where confidence_category between 11 and 100;
```
stephenokeefe: Would this need to be between 11 and 99 as it's inclusive?

karen-hedges: Although there is no confidence category of 100 I have updated to 99 instead.

stephenokeefe (approved):

PR #3294, opened 2026-06-10
DMP-5266 API Validation /events POST Input Hardening

cgi-jhjsimpson:
### JIRA link ###

[DMP-5266](https://tools.hmcts.net/jira/browse/DMP-5266)

### Change description ###

Input validation has been added to fields in the /events POST schema.

- [x] AI-assisted

Initial suggestions were generated for the input validation levels, along with the tests. When reviewing these changes against the requirements, manual updates/changes were made.

- [ ] Yes
- [x] No


src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerAdminGetCaseByIdIntTest.java
```diff
@@ -134,8 +134,8 @@ void adminGetCaseById_WithCaseOpenAndReportingRestrictions() throws Exception {
         String actualResponse = mvcResult.getResponse().getContentAsString();
         String expectedResponse = getContentsFromFile(
             "tests/cases/CaseControllerAdminGetCaseByIdTest/testCaseOpen/expectedResponse.json");
-        expectedResponse = expectedResponse.replace("<CREATED_AT>", courtCase.getCreatedDateTime().format(DateTimeFormatter.ISO_DATE_TIME));
-        expectedResponse = expectedResponse.replace("<LAST_MODIFIED_AT>", courtCase.getLastModifiedDateTime().format(DateTimeFormatter.ISO_DATE_TIME));
+        expectedResponse = expectedResponse.replace("<CREATED_AT>", formatAsUtc(courtCase.getCreatedDateTime()));
```
cgi-jhjsimpson: These were failing locally, as BST was adding +01:00 to the actual response. Is this an acceptable fix or am I missing something with the timezones here?

src/main/resources/openapi/event.yaml
```diff
@@ -747,35 +747,46 @@ components:
...
+          maxLength: 512
         type:
           type: string
+          maxLength: 512
         sub_type:
           type: string
+          maxLength: 512
         event_id:
           type: string
+          maxLength: 10
```
cgi-jhjsimpson: The event_id string is converted to Integer, so anything larger than a signed integer looks like it causes a 500 server error when a `NumberFormatException` is thrown from `EventPersistenceService.java`. If we limit this to the length of an integer (although this doesn't cover numbers larger than a signed integer within this range) then perhaps it would give the called more clarity on why it has failed?

src/main/resources/openapi/event.yaml
```diff
@@ -747,35 +747,46 @@ components:
...
       properties:
         message_id:
           type: string
+          maxLength: 512
         type:
           type: string
+          maxLength: 512
         sub_type:
           type: string
+          maxLength: 512
```
cgi-jhjsimpson, line 760: I've added some liberal validation levels here. varchar in the DB.

src/main/resources/openapi/event.yaml
```diff
@@ -747,35 +747,46 @@ components:
...
           type: string
+          maxLength: 10
+          pattern: '^[0-9]+$'
         courthouse:
           $ref: '#/components/schemas/Courthouse'
         courtroom:
           $ref: '#/components/schemas/Courtroom'
         case_numbers:
           type: array
+          maxItems: 100
```
cgi-jhjsimpson: I'm not entirely sure about the number of cases that may be associated with an event. I'm not sure if Steve or Lee want to sign off on this?

src/main/resources/openapi/event.yaml
```diff
@@ -747,35 +747,46 @@ components:
...
           items:
             $ref: '#/components/schemas/CaseNumber'
         event_text:
           type: string
+          maxLength: 2048
```
cgi-jhjsimpson, line 775: varchar in the DB, so a reasonable value has been provided.

src/main/resources/openapi/event.yaml
```diff
@@ -747,35 +747,46 @@ components:
...
         retention_policy:
           type: object
+          additionalProperties: false
           properties:
             case_retention_fixed_policy:
               type: string
+              maxLength: 512
             case_total_sentence:
               type: string
+              maxLength: 512
```
cgi-jhjsimpson, line 788: I've added some liberal validation levels here.

`case_retention_fixed_policy` is a string used against a lookup table.
`case_total_sentence` is stored in `total_sentence`, a varchar in the DB.

src/main/resources/openapi/event.yaml
karen-hedges: Will this prevent us from logging the data if we restrict the value to a smaller item rather than getting number format exception which will display what was invalid?

cgi-jhjsimpson: Logging would fail, as the initial logging event is [here](https://github.com/hmcts/darts-api/blob/master/src/main/java/uk/gov/hmcts/darts/event/service/impl/EventDispatcherImpl.java#L38). Arguably, adding any limit will prevent us from seeing this logged, though I can loosen this up to account for what might be typically sent. I suppose we're dealing with CP and a legacy system too, and as discussed, we don't want too much restriction.

It doesn't appear that a `NumberFormatException` is caught and handled properly, instead it bubbles up and is handled by the framework. Is this acceptable?

stephenokeefe:
Temporarily closing due to lack of callback URLs

stephenokeefe:
Temporarily closing as this is not in a current sprint

PR #3308, opened 2026-06-19
DMP-5268 API Validation transcriptions/transcriptions POST Input Hardening

cgi-jhjsimpson:
### JIRA link ###

[DMP-5268](https://tools.hmcts.net/jira/browse/DMP-5268)

### Change description ###

Input validation has been added to fields in the /transcriptions POST schema where there were none. Unit tests were added to verify these new limits.

- [ ] AI-assisted

- [ ] Yes
- [ ] No


src/main/resources/openapi/transcriptions.yaml
```diff
@@ -1194,15 +1194,27 @@ components:
...
           type: integer
+          format: int32
+          minimum: 1
+          maximum: 2147483647
           example: 2
         transcription_type_id:
           type: integer
+          format: int32
+          minimum: 1
+          maximum: 2147483647
```
cgi-jhjsimpson, line 1217: I believe a minimum should be set to 1 here. However, though extremely unlikely, this isn't specifically demanded anywhere. So it could be a large minus number? Any input would be appreciated.

PR #3334, opened 2026-06-29
DMP-5586 Add is_data_anonymised To 3 APIs

cgi-jhjsimpson:
### JIRA link ###

[DMP-5586](https://tools.hmcts.net/jira/browse/DMP-5586)

### Change description ###

Opening expired cases in darts-portal was causing 404 errors. The is_data_anonymised field has been added to the following endpoint responses to prevent this behaviour. 

- GET /admin/medias/<id>
- GET /admin/events/<id>
- POST /admin/hearings/search

See ticket DMP-5437.

- [x] AI-assisted

Supported test writing and understanding around data flow.

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/resources/openapi/audio.yaml
+++ b/src/main/resources/openapi/audio.yaml
@@ -848,6 +848,8 @@ components:
       properties:
         id:
           $ref: '#/components/schemas/HearingId'
+        is_data_anonymised:
+          type: boolean
         case_id:
           $ref: '#/components/schemas/CaseId'
         case_number:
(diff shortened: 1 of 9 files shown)


src/main/resources/openapi/audio.yaml
```diff
@@ -848,6 +848,8 @@ components:
       properties:
         id:
           $ref: '#/components/schemas/HearingId'
+        is_data_anonymised:
+          type: boolean
```
stephenokeefe, line 852: For consistency could this use the same format at the other yaml files i.e. $ref: '#/components/schemas/is_data_anonymised' which is defined as a boolean

cgi-jhjsimpson: I think in those files (like hearings.yaml) there are two instances, so it perhaps makes sense to reference the value. There is only one in this file. Going off other examples, there wasn't a clear pattern. I'm happy to make this change if that's how we're doing things?

stephenokeefe: OK let's keep it as is for now if there's only one instance

stephenokeefe (approved):

stephenokeefe:
Merged as part of https://github.com/hmcts/darts-api/pull/3361

PR #3350, opened 2026-07-01
DMP-5272 API Validation /transcriptions/{transcription_id}/document POST Input Hardening

cgi-jhjsimpson:
### JIRA link ###

[DMP-5272](https://tools.hmcts.net/jira/browse/DMP-5272)

### Change description ###

Input validation has been added to fields in the /transcriptions/{transcription_id}/document POST schema.

- [x] AI-assisted

Understanding validation flow, and test creation.

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/resources/openapi/transcriptions.yaml
+++ b/src/main/resources/openapi/transcriptions.yaml
@@ -456,17 +456,21 @@ paths:
           schema:
             type: integer
             format: int64
+            description: Transcription ID. Validated by the application for min and max range.
       requestBody:
+        required: true
         content:
           multipart/form-data:
             schema:
               required:
                 - transcript
               type: object
+              additionalProperties: false
               properties:
                 transcript:
                   type: string
                   format: binary
+                  description: Transcript file. Validated by the application for file extension, content type, and maximum size.
       responses:
         '200':
           description: OK
--- a/src/main/resources/openapi/transcriptions.yaml
+++ b/src/main/resources/openapi/transcriptions.yaml
@@ -470,7 +470,7 @@ paths:
                 transcript:
                   type: string
                   format: binary
-                  description: Transcript file. Validated by the application for file extension, content type, and maximum size.
+                  description: Transcript file. Validated by the application for file extension, content type, and maximum size (10MB).
       responses:
         '200':
           description: OK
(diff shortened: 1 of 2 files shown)


src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerAdminGetCaseByIdIntTest.java
```diff
@@ -328,4 +328,8 @@ private List<EventEntity> createEventsWithDifferentTimestamps(List<OffsetDateTim
                 return event;
             }).toList();
     }
+
+    private static String formatAsUtc(OffsetDateTime dateTime) {
```
karen-hedges: You have this change in multiple PR's. This is not desirable as when you try rebase whichever gets merged into master first, you will get lots of conflicts.

src/main/resources/openapi/transcriptions.yaml
```diff
@@ -456,17 +456,21 @@ paths:
...
         content:
           multipart/form-data:
             schema:
               required:
                 - transcript
               type: object
+              additionalProperties: false
               properties:
                 transcript:
                   type: string
```
stephenokeefe, line 471: Should we add some maxLength validation here?

src/integrationTest/java/uk/gov/hmcts/darts/cases/controller/CaseControllerAdminGetCaseByIdIntTest.java
cgi-jhjsimpson: This was resolved with the fix that went in.

src/main/resources/openapi/transcriptions.yaml
cgi-jhjsimpson: This is a file upload, not a normal text field. OpenAPI represents uploaded files as type: string with format: binary, but the value is really the file contents. I don't believe maxLength is the right option here, as that's for text length. I will detail the 10MB max file size in the description, as it is checked in the app logic in TranscriptFileValidator.java, and this grabs config values from the application.yaml

`max-file-size: 10485760`

stephenokeefe (approved):

PR #3360, opened 2026-07-16
DMP-3394 Modify Automated Task Cron Expression

cgi-jhjsimpson:
### JIRA link ###

[DMP-3394](https://tools.hmcts.net/jira/browse/DMP-3394)

### Change description ###

- Added an additional admin endpoint to allow automated task run schedules to be updated
- POST returns a preview list of future runs, with the returned number defined and modifiable in config
- PATCH updates the task Cron Expression in the database

- [x] AI-assisted

Initial changes completed manually, with AI used to question concepts and implementation and support testing. Some refactoring was completed with AI to speed up delivery, though this required some intervention due to unnecessary complexity.

- [ ] Yes
- [x] No


src/main/java/uk/gov/hmcts/darts/task/controller/AutomatedTasksController.java
```diff
@@ -55,5 +58,22 @@ public ResponseEntity<DetailedAutomatedTask> patchAutomatedTask(Integer taskId,
         var automatedTask = adminAutomatedTaskService.updateAutomatedTask(taskId, automatedTaskPatch);
         return new ResponseEntity<>(automatedTask, HttpStatus.OK);
     }
-}
 
+    @SecurityRequirement(name = SECURITY_SCHEMES_BEARER_AUTH)
+    @Authorisation(contextId = ANY_ENTITY_ID, securityRoles = {SUPER_ADMIN}, globalAccessSecurityRoles = {SUPER_ADMIN})
+    @Override
+    public ResponseEntity<Void> patchAutomatedTaskCronExpression(Integer taskId, AutomatedTaskCronExpressionPatch automatedTaskCronExpressionPatch) {
```
karen-hedges: Can you name the requests that come in as request so AutomatedTaskCronExpressionPatch would be AutomatedTaskCronExpressionPatchRequest

src/main/java/uk/gov/hmcts/darts/task/controller/AutomatedTasksController.java
```diff
@@ -55,5 +58,22 @@ public ResponseEntity<DetailedAutomatedTask> patchAutomatedTask(Integer taskId,
...
+        adminAutomatedTaskService.updateAutomatedTaskCronExpressionSchedule(taskId, automatedTaskCronExpressionPatch);
+        return new ResponseEntity<>(HttpStatus.OK);
+    }
+
+    @SecurityRequirement(name = SECURITY_SCHEMES_BEARER_AUTH)
+    @Authorisation(contextId = ANY_ENTITY_ID, securityRoles = {SUPER_ADMIN}, globalAccessSecurityRoles = {SUPER_ADMIN})
+    @Override
+    public ResponseEntity<List<AutomatedTaskCronExpressionSchedule>> postAutomatedTaskCronExpression(
```
karen-hedges: Can you name the requests and responses so that they end with either Request or Response so AutomatedTaskCronExpressionSchedule would be AutomatedTaskCronExpressionScheduleResponse and AutomatedTaskCronExpressionPost would be AutomatedTaskCronExpressionPostRequest.

src/main/java/uk/gov/hmcts/darts/task/exception/AutomatedTaskApiError.java
```diff
@@ -31,6 +31,11 @@ public enum AutomatedTaskApiError implements DartsApiError {
         AutomatedTaskErrorCode.INCORRECT_AUTOMATED_TASK_TYPE.getValue(),
         HttpStatus.UNPROCESSABLE_ENTITY,
         AutomatedTaskTitleErrors.INCORRECT_AUTOMATED_TASK_TYPE.toString()
+    ),
+    AUTOMATED_TASK_BAD_REQUEST(
```
karen-hedges: Bad request is not descriptive enough. All the ones that are used are specific to what the issue they are related to so you should have invalid cron expression as one, if you need others they should be named appropriately

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceHelper.java
```diff
@@ -0,0 +1,54 @@
...
+        scheduledRun.setExecutionNumber(String.valueOf(executionNumber));
+        scheduledRun.setScheduledAt(scheduledAt);
+        return scheduledRun;
+    }
+
+    public CronExpression validateAndParseCronExpression(String cronExpression) {
+        try {
+            return CronExpression.parse(cronExpression);
+        } catch (IllegalArgumentException e) {
+            throw new DartsApiException(AUTOMATED_TASK_BAD_REQUEST, e);
```
karen-hedges: This should thrown invalid cron expression. We don't use generic messages but specific one so its clear where the error came from. The only exception is the darts-gateway as the errors have to passed to 3rd party systems.

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceHelper.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.List;
+
+import static uk.gov.hmcts.darts.common.util.DateConverterUtil.EUROPE_LONDON_ZONE;
+import static uk.gov.hmcts.darts.task.exception.AutomatedTaskApiError.AUTOMATED_TASK_BAD_REQUEST;
+
+@Service
+@RequiredArgsConstructor
+public class AdminAutomatedTasksServiceHelper {
+
+    private static final int CRON_SCHEDULE_PREVIEW_COUNT = 10;
```
karen-hedges: make this a configurable value in application.yaml. Use CLOSE_OPEN_CASES_OLDER_THAN_PERIOD as an example and set a default value of 10.

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceHelper.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private final Clock clock;
+
+    public List<AutomatedTaskCronExpressionSchedule> getCronExpressionSchedulePreview(CronExpression cron) {
+        List<AutomatedTaskCronExpressionSchedule> scheduledRunTimes = new ArrayList<>();
+
+        ZonedDateTime next = ZonedDateTime.ofInstant(clock.instant(), EUROPE_LONDON_ZONE);
```
karen-hedges: There is a helper class for date time. I think it is DateTimeHelper. See how it is done in other classes

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceHelper.java
```diff
@@ -0,0 +1,54 @@
...
+
+        for (int executionNumber = 1; executionNumber <= CRON_SCHEDULE_PREVIEW_COUNT; executionNumber++) {
+            next = cron.next(next);
+            assert next != null;
```
karen-hedges: don't add asserts in prod code. We rarely use them as they break the code.

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceHelper.java
```diff
@@ -0,0 +1,54 @@
...
+            scheduledRunTimes.add(createScheduledRun(executionNumber, next.toOffsetDateTime()));
+        }
+
+        return scheduledRunTimes;
+    }
+
+    private static AutomatedTaskCronExpressionSchedule createScheduledRun(int executionNumber, OffsetDateTime scheduledAt) {
```
karen-hedges: I'm not sure why you are doing this. You don't need to set the next scheduled runs as that is already done by the AbstractLockableAutomatedTask

src/main/java/uk/gov/hmcts/darts/task/exception/AutomatedTaskApiError.java
cgi-jhjsimpson: I did have this more appropriately named, though as it is used to address several different responses that require a 400, I thought I'd package it into one and keep it generic. I'll break it out into two.

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceHelper.java
cgi-jhjsimpson: I did have this is a value in the application yaml, though reversed it, as I thought a constant could handle it. I also doubt that we're wanting to set an env variable higher up that it could inherit? I do like it in the config though, I'll add it back.

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceHelper.java
cgi-jhjsimpson: Frustratingly it will never be null, because of the null checks, yet I'm getting a complaint. I'll remove it.

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceHelper.java
cgi-jhjsimpson: This returns the scheduled run object, nothing more.

src/main/java/uk/gov/hmcts/darts/task/exception/AutomatedTaskApiError.java
karen-hedges: If you search throughout the code even ones that return the same status have specifically named error codes so its easy to identify where the error came from

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceHelper.java
karen-hedges: You are no incrementing the date so they will all come back with the same dates?

src/integrationTest/java/uk/gov/hmcts/darts/task/service/AdminEditAutomatedTaskCronExpressionIntTest.java
karen-hedges, line 1: Integration tests should be named IntTest not just Test to differentiate for unit tests which only end Test

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceHelper.java
cgi-jhjsimpson: I'm re-reading what you've asked for. When you say you want a an implementation like `close-open-cases-older-than-period: ${CLOSE_OPEN_CASES_OLDER_THAN_PERIOD:P6Y}` with a default value, do you then want us to be able to pass down a value if it's present, though otherwise default to 10? I.e. rather than hardcoding 10, similar to `overridable-fixed-policy-keys: 3`

src/main/java/uk/gov/hmcts/darts/task/controller/AutomatedTasksController.java
cgi-jhjsimpson: Done.

src/main/java/uk/gov/hmcts/darts/task/controller/AutomatedTasksController.java
cgi-jhjsimpson: Done.

src/main/java/uk/gov/hmcts/darts/task/exception/AutomatedTaskApiError.java
cgi-jhjsimpson: Done.

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceHelper.java
cgi-jhjsimpson: Done.

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceHelper.java
cgi-jhjsimpson: Done.

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceHelper.java
cgi-jhjsimpson: Done.

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceHelper.java
cgi-jhjsimpson: Done.

src/main/java/uk/gov/hmcts/darts/task/service/impl/AdminAutomatedTasksServiceHelper.java
cgi-jhjsimpson: This generates the objects (depending on the config value count) that contain the execution_number and scheduled_at values for each subsequent run. I've updated the naming though, as maybe that's a little confusing.

src/integrationTest/java/uk/gov/hmcts/darts/task/service/AdminEditAutomatedTaskCronExpressionIntTest.java
cgi-jhjsimpson: Done.

src/main/java/uk/gov/hmcts/darts/util/DateTimeHelper.java
```diff
@@ -18,4 +22,8 @@ public static String getDateTimeIsoFormatted(OffsetDateTime dateTime) {
     public static OffsetDateTime floorToMinutes(OffsetDateTime dateTime) {
         return Objects.nonNull(dateTime) ? dateTime.truncatedTo(ChronoUnit.MILLIS).withSecond(0).withNano(0) : null;
     }
+
+    public static ZonedDateTime getCurrentTimeInEuropeLondon(Clock clock) {
```
karen-hedges: Shouldn't you use toZonedDateTime instead of creating this new one

cgi-jhjsimpson: Done.

stephenokeefe (approved):

karen-hedges:
Merged into release 25 https://github.com/hmcts/darts-api/pull/3407

PR #3374, opened 2026-07-31
DMP-5190 AudioLinkingAutomatedTask Rethrow InterruptedException

cgi-jhjsimpson:
### JIRA link  ###

[DMP-5190](https://tools.hmcts.net/jira/browse/DMP-5190)

### Change description ###

InterruptedException is now handled in the case of an automated task reaching its max lock time. An exception is thrown and handled higher up, instead of being swallowed.

- [x] AI-assisted

Provided suggestions around thread handling and tests.

**Does this PR introduce a breaking change?**

- [ ] Yes
- [x] No


src/main/java/uk/gov/hmcts/darts/task/runner/impl/AudioLinkingAutomatedTask.java
```diff
@@ -107,7 +107,22 @@ public void processEvent(Long eveId) {
...
             } catch (Exception e) {
                 log.error("Error attempting to link media for event with eveId {}", eveId, e);
+                if (causedByInterruptedException(e)) {
+                    Thread.currentThread().interrupt();
+                    throw new IllegalStateException("Audio linking task interrupted", e);
+                }
             }
         }
+
+        private static boolean causedByInterruptedException(Throwable exception) {
```
karen-hedges: It might make more sense to put this method into a helper class as there are various tickets to handle InterruptedException for automated tasks

cgi-jhjsimpson: That's a fair comment, I'll do that.

src/main/java/uk/gov/hmcts/darts/task/helper/AutomatedTaskExceptionHelper.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.hmcts.darts.task.helper;
+
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class AutomatedTaskExceptionHelper {
+    public boolean causedByInterruptedException(Throwable exception) {
+        Throwable cause = exception;
+        while (cause != null) {
```
karen-hedges: Why are using while here, surely it should just be an if statement?

src/main/java/uk/gov/hmcts/darts/task/helper/AutomatedTaskExceptionHelper.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.hmcts.darts.task.helper;
+
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class AutomatedTaskExceptionHelper {
+    public boolean causedByInterruptedException(Throwable exception) {
+        Throwable cause = exception;
```
karen-hedges: Why are creating a new object here, is it really necessary?

cgi-jhjsimpson: I think this is perhaps over engineering for readability, I'll update it.

src/main/java/uk/gov/hmcts/darts/task/helper/AutomatedTaskExceptionHelper.java
cgi-jhjsimpson: Codex questioned my original implementation, as it was concerned that there could be a case where an InterruptedException could be caught and wrapped inside another exception type. In that case, it wouldn't be caught by the if statement. I think it has taken a more bolts and braces approach, rather than perhaps something that is necessary. However, if we're using this method across multiple automated tasks, it might be worth keeping?

src/main/java/uk/gov/hmcts/darts/task/helper/AutomatedTaskExceptionHelper.java
cgi-jhjsimpson: Ah, PMD doesn't like us reassigning the parameter. I could just keep this in rather than suppressing, what do you think?

src/main/java/uk/gov/hmcts/darts/task/helper/AutomatedTaskExceptionHelper.java
karen-hedges: While condition implies an iterator but you are not sending an iterator so you don't need to handle this scenario. If there are places further up in the code for the automated task that catches exception then you need to make sure it rethrows that exception not returns true. I think you will find the base class for interrupted exception handles it and also it sets the task to failed.

karen-hedges:
Merged into release 25 https://github.com/hmcts/darts-api/pull/3407

PR #3377, opened 2026-08-06
DMP-3712 Migrate towards the use of Spring 3 Problem Details

karen-hedges:

https://tools.hmcts.net/jira/browse/DMP-3712

### Change description ###

Migrate to use spring 3 problem details

[x] AI-assisted

AI used to transform from using RFC-7808 zalando to using spring boot 3 problem details

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/darts/common/exception/DartsApiTrait.java
```diff
@@ -2,46 +2,36 @@
...
-import org.zalando.problem.spring.web.advice.AdviceTrait;
 import uk.gov.hmcts.darts.authorisation.exception.AuthorisationError;
 
 import java.io.IOException;
 import java.net.URI;
+import java.util.HashMap;
 import java.util.Map.Entry;
 
-public interface DartsApiTrait extends AdviceTrait {
+public interface DartsApiTrait {
```
stephenokeefe: DartsApiTrait being an interface with non abstract methods seems confusing, as DartsApiTraitImpl implements it but does not need to implement the same methods. Could these be renamed or combined in to one Controller Advice class?

src/main/resources/openapi/usermanagement.yaml
```diff
@@ -119,12 +119,13 @@ paths:
               schema:
                 $ref: './problem.yaml'
               example:
-                violations:
-                  - field: emailAddress
-                    message: size must be between 1 and 256
-                type: https://zalando.github.io/problem/constraint-violation
-                status: 400
+                type: about:blank
```
stephenokeefe: The yaml suggests that the type for a bad request error is about:blank, but the searchShouldReturnBadRequestError test expects the type to be COMMON_104

src/main/java/uk/gov/hmcts/darts/common/exception/DartsApiTrait.java
karen-hedges: Replaced with DartsApiExceptionHandler, DartsApiProblemDetailFactory and DartsApiErrorResponseWriter

src/main/resources/openapi/usermanagement.yaml
karen-hedges: Corrected this

PR #3381, opened 2026-08-14
DMP-5312 Disable User Accounts Inactive For Six Months Or More

cgi-jhjsimpson:
### JIRA link ###

[DMP-5312](https://tools.hmcts.net/jira/browse/DMP-5312)

### Change description ###

A new automated task, DisableInactiveUserAccounts, has been added. This disables user accounts (except for specific/higher level accounts) that have not been used for six months or more. It also removes the account from any security groups it is associated with. This does not delete accounts.

- [x] AI-assisted

Use of Codex to generate the automated task implementation, based on the Codex SKILL guide.

**Does this PR introduce a breaking change?**

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/java/uk/gov/hmcts/darts/common/repository/UserAccountRepository.java
+++ b/src/main/java/uk/gov/hmcts/darts/common/repository/UserAccountRepository.java
@@ -1,11 +1,13 @@
 package uk.gov.hmcts.darts.common.repository;
 
+import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
 import org.springframework.data.jpa.repository.Modifying;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.history.RevisionRepository;
+import org.springframework.data.repository.query.Param;
 import org.springframework.stereotype.Repository;
 import uk.gov.hmcts.darts.common.entity.CourthouseEntity;
 import uk.gov.hmcts.darts.common.entity.UserAccountEntity;
@@ -16,6 +18,7 @@ import java.util.Optional;
 import java.util.Set;
 
 @Repository
+@SuppressWarnings("PMD.TooManyMethods")//Repository class so low complexity in this case
 public interface UserAccountRepository extends
     RevisionRepository<UserAccountEntity, Integer, Long>,
     JpaRepository<UserAccountEntity, Integer>,
@@ -72,6 +75,28 @@ public interface UserAccountRepository extends
         WHERE id = :userId
         """)
     void updateLastLoginTime(Integer userId, OffsetDateTime now);
+
+    @Query("""
+        SELECT DISTINCT userAccount
+        FROM UserAccountEntity userAccount
+        WHERE userAccount.active = true
+        AND userAccount.isSystemUser = false
+        AND (
+            userAccount.lastLoginTime <= :cutoffDateTime
+            OR (userAccount.lastLoginTime IS NULL AND userAccount.createdDateTime <= :cutoffDateTime)
+        )
+        AND NOT EXISTS (
+            SELECT securityGroup
+            FROM UserAccountEntity excludedUser
+            JOIN excludedUser.securityGroupEntities securityGroup
+            WHERE excludedUser = userAccount
+            AND securityGroup.securityRoleEntity.id IN :excludedRoleIds
+        )
+        ORDER BY userAccount.id ASC
+        """)
+    List<UserAccountEntity> findInactiveUsersExcludingRoles(@Param("cutoffDateTime") OffsetDateTime cutoffDateTime,
+                                                            @Param("excludedRoleIds") Set<Integer> excludedRoleIds,
+                                                            Pageable pageable);
 
     List<UserAccountEntity> findByIdInAndActive(List<Integer> userIds, Boolean active);
 
--- a/src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/DisableInactiveUserAccountsServiceImpl.java
+++ b/src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/DisableInactiveUserAccountsServiceImpl.java
@@ -1,0 +1,59 @@
+package uk.gov.hmcts.darts.usermanagement.service.impl;
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.data.domain.PageRequest;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+import uk.gov.hmcts.darts.common.entity.UserAccountEntity;
+import uk.gov.hmcts.darts.common.helper.CurrentTimeHelper;
+import uk.gov.hmcts.darts.common.repository.UserAccountRepository;
+import uk.gov.hmcts.darts.usermanagement.service.DisableInactiveUserAccountsService;
+
+import java.time.OffsetDateTime;
+import java.time.Period;
+import java.util.List;
+import java.util.Set;
+
+import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.SUPER_ADMIN;
+import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.SUPER_USER;
+
+@Service
+@Slf4j
+@RequiredArgsConstructor
+public class DisableInactiveUserAccountsServiceImpl implements DisableInactiveUserAccountsService {
+
+    private static final Period INACTIVITY_PERIOD = Period.ofMonths(6);
+    private static final Set<Integer> EXCLUDED_ROLE_IDS = Set.of(SUPER_USER.getId(), SUPER_ADMIN.getId());
+    private static final int MINIMUM_BATCH_SIZE = 1;
+
+    private final UserAccountRepository userAccountRepository;
+    private final CurrentTimeHelper currentTimeHelper;
+
+    @Override
+    @Transactional
+    public void process(int batchSize) {
+        int safeBatchSize = Math.max(batchSize, MINIMUM_BATCH_SIZE);
+        OffsetDateTime cutoffDateTime = currentTimeHelper.currentOffsetDateTime().minus(INACTIVITY_PERIOD);
+        List<UserAccountEntity> inactiveUsers = userAccountRepository.findInactiveUsersExcludingRoles(
+            cutoffDateTime,
+            EXCLUDED_ROLE_IDS,
+            PageRequest.of(0, safeBatchSize)
+        );
+
+        if (inactiveUsers.isEmpty()) {
+            log.info("No inactive user accounts found to disable");
+            return;
+        }
+
+        inactiveUsers.forEach(this::disableAndRemoveFromSecurityGroups);
+        userAccountRepository.saveAll(inactiveUsers);
+
+        log.info("Disabled {} inactive user accounts", inactiveUsers.size());
+    }
+
+    private void disableAndRemoveFromSecurityGroups(UserAccountEntity userAccount) {
+        userAccount.getSecurityGroupEntities().clear();
+        userAccount.setActive(false);
+    }
+}
--- a/skills/README.md
+++ b/skills/README.md
@@ -1,0 +1,23 @@
+# DARTS Skills
+
+This folder holds reusable Codex skill guides for `darts-api`. Add future skills as separate subfolders so each guide can
+keep its instructions, examples, and README together.
+
+## Available Skills
+
+- [Automated Task Generator](automated-task-generator/README.md): creates the standard DARTS automated-task wiring,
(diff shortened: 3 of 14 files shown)


src/main/java/uk/gov/hmcts/darts/common/repository/UserAccountRepository.java
```diff
@@ -73,6 +76,28 @@ AND securityRole.id IN (:roleIds)
...
+            FROM UserAccountEntity excludedUser
+            JOIN excludedUser.securityGroupEntities securityGroup
+            WHERE excludedUser = userAccount
+            AND securityGroup.securityRoleEntity.id IN :excludedRoleIds
+        )
+        ORDER BY userAccount.id ASC
+        """)
+    List<UserAccountEntity> findInactiveUsersExcludingRoles(@Param("cutoffDateTime") OffsetDateTime cutoffDateTime,
+                                                            @Param("excludedRoleIds") Set<Integer> excludedRoleIds,
+                                                            Pageable pageable);
```
stephenokeefe: Could we use LIMIT instead of pageable if we're not actually using the pagination. That is what we tend to do in the repo queries

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/DisableInactiveUserAccountsServiceImpl.java
```diff
@@ -0,0 +1,75 @@
...
+
+        inactiveUsers.forEach(this::disableAndRemoveFromSecurityGroups);
+        userAccountRepository.saveAll(inactiveUsers);
+
+        log.info("Disabled {} inactive user accounts", inactiveUsers.size());
+    }
+
+    private void disableAndRemoveFromSecurityGroups(UserAccountEntity userAccount) {
+        rollbackAssignedTranscriptionsIfTranscriber(userAccount);
+        userAccount.getSecurityGroupEntities().clear();
```
stephenokeefe: Could call UserManagementService.unassignUserFromGroupsTheyArePartOf here

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/DisableInactiveUserAccountsServiceImpl.java
```diff
@@ -0,0 +1,75 @@
...
+    @Override
+    @Transactional
+    public void process(int batchSize) {
+        int safeBatchSize = Math.max(batchSize, MINIMUM_BATCH_SIZE);
+        OffsetDateTime cutoffDateTime = currentTimeHelper.currentOffsetDateTime().minus(INACTIVITY_PERIOD);
+
+        // System users are excluded in the repository query with isSystemUser = false.
+        List<UserAccountEntity> inactiveUsers = userAccountRepository.findInactiveUsersExcludingRoles(
+            cutoffDateTime,
+            PRIVILEGED_USER_ROLE_IDS,
```
stephenokeefe: As well as the currently excluded users, users where user_email_address ilike '%localhost%' should also be excluded

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/DisableInactiveUserAccountsServiceImpl.java
```diff
@@ -0,0 +1,75 @@
...
+import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.SUPER_ADMIN;
+import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.SUPER_USER;
+import static uk.gov.hmcts.darts.common.enums.SecurityRoleEnum.TRANSCRIBER;
+
+@Service
+@Slf4j
+@RequiredArgsConstructor
+public class DisableInactiveUserAccountsServiceImpl implements DisableInactiveUserAccountsService {
+
+    private static final Period INACTIVITY_PERIOD = Period.ofMonths(6);
```
stephenokeefe: Could this be a configurable value in disable-inactive-user-accounts in application.yaml

src/main/java/uk/gov/hmcts/darts/common/repository/UserAccountRepository.java
cgi-jhjsimpson: Done, users with localhost as their email address have also been removed.

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/DisableInactiveUserAccountsServiceImpl.java
cgi-jhjsimpson: Done.

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/DisableInactiveUserAccountsServiceImpl.java
cgi-jhjsimpson: Done in the query.

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/DisableInactiveUserAccountsServiceImpl.java
cgi-jhjsimpson: This has been extracted to a separate class, and is then called by both services, rather than making it a public method.

stephenokeefe (approved):

src/integrationTest/java/uk/gov/hmcts/darts/common/repository/UserAccountRepositoryIntTest.java
```diff
@@ -29,16 +55,41 @@ class UserAccountRepositoryIntTest extends PostgresIntegrationBase {
...
     @BeforeEach
     void setUp() {
-        integrationTestUser = dartsDatabase.getUserAccountStub().getIntegrationTestUserAccountEntity();
-        Mockito.when(authorisationApi.getCurrentUser()).thenReturn(integrationTestUser);
+        UserAccountEntity user1 = PersistableFactory.getUserAccountTestData().someMinimalBuilder()
+            .emailAddress("some.user.email@example.net")
+            .build()
+            .getEntity();
+        dartsPersistence.save(user1);
+        userAccountEntity1 = user1;
```
karen-hedges: why are you creating 1 user and assigning another to it. Just use userAccountEntity1. Just so you know, when you save an object to the DB, the save method returns the object with the saved elements such as id

cgi-jhjsimpson: Done.

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/DisableInactiveUserAccountsServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+@Slf4j
+@RequiredArgsConstructor
+public class DisableInactiveUserAccountsServiceImpl implements DisableInactiveUserAccountsService {
+
+    @Value("${darts.automated.task.disable-inactive-user-accounts.inactivity-period-limit:P6M}")
+    private Period inactivityPeriodLimit;
+
+    private static final String OWNER_WAS_DISABLED_DUE_TO_INACTIVITY = "Owner was disabled due to inactivity";
+    private static final Set<Integer> PRIVILEGED_USER_ROLE_IDS = Set.of(SUPER_USER.getId(), SUPER_ADMIN.getId());
+    private static final int MINIMUM_BATCH_SIZE = 1;
```
stephenokeefe: I think it would be simpler to just use the batch size that is set by the user even if it is 0. This likely wouldn't happen but as it's not clear on the frontend that setting a batch size of 0 would mean that the batch would actually be 1, if the batch size is set as 0 that should be accepted. If having a batch size of 0 causes an unavoidable failure, then it should at least be logged if it's changed to 1 in darts-api. Preferably it would just be kept at 0

src/integrationTest/java/uk/gov/hmcts/darts/common/repository/UserAccountRepositoryIntTest.java
```diff
@@ -47,4 +95,260 @@ void findByEmailAddressIgnoreCase_shouldReturnExpectedUserAccount_whenItExistsIg
         assertEquals(1, foundUsers.size());
         assertEquals(integrationTestUser.getEmailAddress(), foundUsers.get(0).getEmailAddress());
     }
+
+    @Test
```
stephenokeefe: A lot of the findusers tests in UserAccountRepositoryIntTest are already covered in UserAccountRepositoryIntTest so can be removed to improve build times

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/DisableInactiveUserAccountsServiceImpl.java
karen-hedges: Also batchsize is retrievable from the automated task. See ArmRpoBacklogCatchupAutomatedTask for an example of how batch size for automated task is passed to service. It will use what is configured in the DB.

src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/TranscriptionServiceImpl.java
```diff
@@ -417,6 +417,17 @@ public void closeTranscription(Long transcriptionId, String transcriptionComment
         }
     }
 
+    @Override
+    @Transactional
```
karen-hedges: You have @Transactional in DisableInactiveUserAccountsServiceImpl. You shouldn't really have more than 1. Is there a reason you have another one? @Transactional is a rollback point.

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/DisableInactiveUserAccountsServiceImpl.java
cgi-jhjsimpson: I've updated this to follow the existing patterns.

src/integrationTest/java/uk/gov/hmcts/darts/common/repository/UserAccountRepositoryIntTest.java
cgi-jhjsimpson: Good spot, I've removed those.

src/main/java/uk/gov/hmcts/darts/transcriptions/service/impl/TranscriptionServiceImpl.java
cgi-jhjsimpson: I can remove it if you think that's best. From what I'm reading, it will gather it up into the calling @Transactional. However, it is public, so if called it would protect itself if there was a failure. Having said that, maybe someone would be calling it with a method that is @Transcational and so this wouldn't matter so much. Any thoughts?

cgi-jhjsimpson: I've removed this `@Transactional` and extracted the shared code into a separate method. Both these methods will be wrapped in the calling methods `@Transactional`. Initially this was done to resolve SonarQube flagging the internal call from this method to a `@Transactional` method within the class (`closeUserTranscriptions()` -> `@Transactional closeTranscription()`). It was warning that it might seem to those calling `closeUserTranscriptions()` directly that this would be a `@Transactional` operation. However, as it's making the call internally, the proxy wouldn't pick it up. Extracting the logic resolves the complaint. It also prevents any options added to the @Transcational not being picked up, as those for the internal call would be ignored.

cgi-jhjsimpson: I've just seen that this problem occurs again, as the new method I've extracted calls another public method that is @Transactional. What a pain!

karen-hedges: I think what you can do is if the @Transactional method is a shared method, create a new wrapper method with transactional that calls this method and move it to that new method

src/main/java/uk/gov/hmcts/darts/usermanagement/service/impl/DisableInactiveUserAccountsServiceImpl.java
```diff
@@ -68,6 +68,6 @@ private void processInactiveUser(UserAccountEntity userAccount) {
             userAccount.setActive(false);
         }
         // Any user that has been inactive for the specified period will be made INACTIVE, removed from any groups and their transcriptions will be closed.
-        log.info("User account {} has been disabled or cleaned up due to inactivity", userAccount.getId());
+        log.info("User account {} has been disabled or cleaned up due to inactivity, with email {}", userAccount.getId(), userAccount.getEmailAddress());
```
karen-hedges: You misunderstood my previous comment. Don't log with info any PII data. Only use usr_id but for debugging you can have a new log.debug that has the extra information. This does mean 2 log statements but these are targetted.

